PR #300, opened 2014-08-29
Cubes!

yarokifor:
I added cubes to worldedit. As well as a helpful command, platfrom. This feature have been both tested farelly well in only bukkit. I add a new configuration feild for cubes because going off MaxRadius didn't seem to make sense. Also adding the raise flag for pyrmids. Note this commits might not look right because I kinda suck at using git but I am getting better.


kenzierocks:
Closing this PR, there are too many conflicts for easy updating + merging. I do like the feature, so if anyone is willing to re-write this on top of `master`, feel free.

PR #312, opened 2014-12-13
Some fixes and tweaks for PR #309

luacs1998:
My sincere apologies for the inconvenience caused.

This PR fixes some commands not being registered with the Forge command manager, as well as switches the declaration for permissions handling to NORMAL.


luacs1998:
This fixes a crash that can be encountered should a command have no permissions.

sk89q:
Pulling but I may end up changing the interfaces if that's possible...

luacs1998:
Change it all you want. I will get notified and will make updates as needed anyway...

PR #314, opened 2015-01-14
Directly access entity list

kenzierocks:
This replaces the very slow chunk dance that was being used, and will fix [WORLDEDIT-3241](http://youtrack.sk89q.com/issue/WORLDEDIT-3241) (~~testing in progress~~ all good!).


kenzierocks:
Ready for pull.

PR #315, opened 2015-01-14
Add world flag

kenzierocks:
As per PseudoKnight's suggestion on [WORLDEDIT-3241](http://youtrack.sk89q.com/issue/WORLDEDIT-3241), this adds a flag for globally butchering, which is normally done by changing the default to -1 in the config.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -178,7 +178,7 @@ public class EditSession implements Extent {
         if (world != null) {
             Extent extent;
 
-            // This extents are ALWAYS used
+            // These extents are ALWAYS used
             extent = fastModeExtent = new FastModeExtent(world, false);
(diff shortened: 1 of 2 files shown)


TheMolkaPL:
:+1:

wizjany:
I feel as if this is a bit redundant. We're not really even saving typing time vs. doing -1. The suggestion on the linked issue was simply because someone used //butcher 5000, where //butcher -1 would have worked fine (I believe PseudoKnight just overlooked the presence of -1).

kenzierocks:
You actually can't use -1, that chooses the max radius.

wizjany:
Which defaults to the whole (loaded) world. If someone changes the max in their config, the -w flag would bypass that. This is more about loading chunks.

kenzierocks:
I see. Permissions maybe?

kenzierocks:
Should I close this or redo the commits?

kenzierocks:
Closing this, merge is a mess.

wizjany:
As I said before, I don't see any need for the feature.

PR #317, opened 2015-02-12
Update .gitignore with more files

kenzierocks:
Currently, when testing WorldEdit I get upwards of 1000 files that I need to make sure I don't commit. This gitignore patch will ignore those files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.metadata
+.recommenders
 .classpath
 .project
 .settings
--- a/worldedit-forge/.gitignore
+++ b/worldedit-forge/.gitignore
@@ -1,0 +1,5 @@
+/config
+/logs
+/saves
+/options.txt
+/usernamecache.json


wizjany:
._. how exactly do you test worldedit?

kenzierocks:
1) With the Run button in Eclipse, it's easiest and I know of no other way.
2) Really? I'll update that exclude then.

kenzierocks:
Okay, should be good now.

wizjany:
Hm, I can't remember eclipse making random stuff like that. Do you have your run configuration set to startup a server or something? Maybe you need to change the working directory?

kenzierocks:
It runs a full Forge client. I could change the directory, but that breaks things if I remember.

wizjany:
I feel like these are some rather generic folders that would be better suited for your personal excludes file.

kenzierocks:
You can do nested personal excludes?

kenzierocks:
Well, that closed that. I'll do it locally.

PR #320, opened 2015-02-21
Fix imprecise rotations bug.

Dumbo52:
This fixes a bug which causes inaccuracies when rotating large selections by multiples of 90 degrees. For example, in the image below, the rectangle on the left was rotated 180 degrees to produce the rectangle on the right:

The bug is also documented [here](http://youtrack.sk89q.com/issue/WORLDEDIT-3220).


sk89q:
Thanks for the PR.

PR #321, opened 2015-02-28
Add stone to usesData HashSet.

Dumbo52:
When WorldGuard's 'protection.item-durability' option is set to false, breaking a block when holding a stone variant in your hand (granite, diorite, andesite, and their polished versions) will cause the item being held to [turn back to stone](https://github.com/sk89q/WorldGuard/blob/a1a6ee7107bd8c3f86bf213214dd83a696f93dec/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java#L109).


wizjany:
Oh man I totally forgot about this feature...someone actually mentioned this bug a bit ago and I didn't think worldguard did this at all.

PR #326, opened 2015-05-23
Add spigot_v1_8_R3 adapter.

twizmwazin:
This pull request adds a spigot_v1_8_R3 adapter.


sk89q:
@ other WE devs: we should compile this ourselves for security reasons

luacs1998:
Though, why would someone PR a class file into the upstream project, as opposed to just providing unofficial builds?

DarkArc:
@luacs1998 You mean maliciously? It's a matter of distribution, if you can "infect" the official build, then there's a potential for the vast majority of users to become victims -- not just those whom were too impatient to wait for an official release.

PikachuEXE:
I wonder if there is a proper process for this kind of PR (which involves class file not source code)

sk89q:
Well, someone submitting a PR to the adapters repo probably means that we should compile it and put it in WE.

TheMolkaPL:
@wizjany pushed it in a3a40396469a143fa56bf121a05f1498cd927ccf.

Can be closed. :apple:

PR #328, opened 2015-06-15
Command Tool

spaceribs:
Relatively simple stuff, this should allow people to hook up tools directly to their craftscripts. Parameters are set when it's attached to a tool, other button does nothing yet.

```
/tool command something.js param1 [param2…]
```


TomyLobo:
why restrict it to javascript?
In fact, I already [wrote] (https://github.com/FoxelBox/FoxBukkit/blob/06df44a911055ae931c55ac8ddedb7019dcf557c/src/main/java/com/foxelbox/foxbukkit/main/commands/BindCommand.java#L42) such a command for another plugin and called it ``/bind``.

spaceribs:
what would be the other we commands that you would want to bind to a tool beyond what's already there? Many of them require a selection box anyway, which would kind of restrict the usability beyond craftscripts made to be used as tools.

sk89q:
Well, all other functions could be adapted to brushes easily if they only require a region because a region can be generated at the point of brush click... however, that still requires a restructure of the current command code to make that possible and that's something I've been working on.

TomyLobo:
Don't let your lack of imagination hinder the usefulness of a feature.
We've used it for plenty of stuff.
Although I should mention that the ``/bind`` command I wrote supported running multiple commands in sequence. For instance, you could use a selection-based command and then move the selection with the same bind.

sk89q:
That's pretty hacky in the case of adapting WE commands though. It'd be better if all functions (set, replace, so on) had factories that made them easy to adapt to brushes, selection commands, and area effect commands.

spaceribs:
haha, it's not an imagination issue, it's just that we could run those worldedit commands in javascript anyway and with a lot more control.

@sk89q I'd love to see how that would work, it sounds great. My use case is tools for a new city generator, each tool doing a specific job like flattening, placing grids, placing roads. It's less oriented towards brushes and more towards gridding out chunks, but having a vine/decay brush would be really cool.

sk89q:
Technically you could get a vine brush going with:

```/br sphere vine <radius>
/mask <vine 0
```

spaceribs:
I was thinking something even more complex, where theres probability involved between wood rotting away, stone being replaced with a cracked version, dirt/grass growing in random places, all dependent on the radius of the brush and the amount of "decay time" you could set.

sk89q:
Ah okay. I suppose that's doable with masks and the probability pattern on a brush, but you'd need a brush for every transformation (one for wood, one for stone, one for dirt/grass, etc.).

spaceribs:
yeah, theres also a weird amount of logic too, like dirt can't just be placed anywhere, you would want it contained by other blocks, and only sometimes have grass

spaceribs:
Added a `getClicked` and `setClicked` to the player class so that craftscripts can access the element that was rightclicked

spaceribs:
Hey, I noticed this PR has been open for a few years, should I close it?

me4502:
Is this limited to only 10 spaces in the command name?

Does the WorldEdit command system allow ""s to read a string as a single argument?

spaceribs:
@me4502 I'm not entirely sure, the @Command decorator is definitely higher up than my function, and a space should denote a different argument. Would you like me to make this match:

https://github.com/spaceribs/WorldEdit/blob/12b4a60fee58ea49c6c692506f4cb69495ec47b8/worldedit-core/src/main/java/com/sk89q/worldedit/command/ScriptingCommands.java#L59

me4502:
I don't know how it works, I was more asking if that limitation was in your implementation.

@wizjany may know more about the WorldEdit command system, I don't believe I've used it.

boy0001:
You should just ommit the max and it'll default to `-1` and won't impose any limitation on the number of arguments. 

> Does the WorldEdit command system allow ""s to read a string as a single argument?

Yes

kenzierocks:
Thank you for your contribution. I've updated this to WE 7 standards as #519.

PR #329, opened 2015-06-15
[WIP] Use more of guava

kenzierocks:
Re-write some sections to take advantage of `guava` now that it is available everywhere. Not everything will be converted (it'll take forever to do that solo) but many of the immutable collections will be used.

Things to do:
- [ ] Add a bunch of `null` checks (`checkNotNull`) where there are none and the parameters cannot be `null`
 - [ ] Core
 - [ ] Bukkit
 - [ ] Forge

This PR is currently for tracking changes only and is in no way ready to merge, but please comment if you see something terribly wrong.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/blocks/BlockType.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/blocks/BlockType.java
@@ -25,7 +25,6 @@ import com.sk89q.util.StringUtil;
 import com.sk89q.worldedit.PlayerDirection;
 
 import javax.annotation.Nullable;
-
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Random;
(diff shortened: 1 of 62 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/blocks/BlockType.java
```diff
@@ -625,62 +641,64 @@ public boolean canPassThrough() {
     /**
      * HashSet for centralTopLimit.
      */
-    private static final Map<Integer, Double> centralTopLimit = new HashMap<Integer, Double>();
+    private static final Map<Integer, Double> centralTopLimit;
```
kenzierocks: @TomyLobo This `Map` has conflicts, should it be having those or does this code need to be rewritten?

kenzierocks:
If we really need this I'll think it over again later.

PR #330, opened 2015-06-26
Fixes issue 3315 with VirtualPlayers

Europia79:
http://youtrack.sk89q.com/issue/WORLDEDIT-3315

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/session/SessionManager.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/session/SessionManager.java
@@ -154,7 +154,8 @@ public class SessionManager {
             session.setConfiguration(config);
             session.setBlockChangeLimit(config.defaultChangeLimit);
 
-            // Remember the session if the session is still active
+            // Remember the session regardless of if it's currently active or not.
+            // And have the SessionTracker FLUSH inactive sessions.
             sessions.put(getKey(owner), new SessionHolder(sessionKey, session));
         }
 
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/session/SessionManager.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/session/SessionManager.java
@@ -155,9 +155,7 @@ public class SessionManager {
             session.setBlockChangeLimit(config.defaultChangeLimit);
 
             // Remember the session if the session is still active
-            if (sessionKey.isActive()) {
-                sessions.put(getKey(owner), new SessionHolder(sessionKey, session));
-            }
+            sessions.put(getKey(owner), new SessionHolder(sessionKey, session));
         }
 
         // Set the limit on the number of blocks that an operation can


LaughNgamez:
I highly encourage that this is accepted as it does no harm and unlocks potential.

Europia79:
http://bukkit.org/threads/virtualplayer-cannot-execute-worldguard-worldedit-cmds.352155/

I had a lot of people report this issue to me... At first, I thought it was a non-issue... But then we had the recent sign exploit... And having these VirtualPlayers execute WorldEdit commands to fix signs all over the world was incredibly helpful for fixing my server !!! Who knew ?

worldedit-core/src/main/java/com/sk89q/worldedit/session/SessionManager.java
```diff
@@ -155,9 +155,7 @@ public synchronized LocalSession get(SessionOwner owner) {
             session.setBlockChangeLimit(config.defaultChangeLimit);
 
             // Remember the session if the session is still active
```
kenzierocks: You should probably change this comment, it no longer explains the code correctly.

TomyLobo:
The proper place to solve this would be in VirtualPlayers, not WorldEdit.
Otherwise you'll always have players that aren't really there issuing commands and some plugins might not like it.

In order to make session.isActive() true on Bukkit, you have to fulfill this condition:
````
return Bukkit.getServer().getPlayerExact(name) != null;
````
https://github.com/sk89q/WorldEdit/blob/master/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java#L238

Unfortunately, getPlayerExact is implemented differently on CraftBukkit and Spigot so you cannot just reflect yourself into it and hope for the best :)
[CraftBukkit] (https://github.com/starlis/EMC-CraftBukkit/blob/5ac44d361574ea0585b166d4e68c5a6ef9e4d0de/src/main/java/org/bukkit/craftbukkit/CraftServer.java)
````
    @Override
    @Deprecated
    public Player getPlayerExact(String name) {
        Validate.notNull(name, "Name cannot be null");

        String lname = name.toLowerCase();

        for (Player player : getOnlinePlayers()) {
            if (player.getName().equalsIgnoreCase(lname)) {
                return player;
            }
        }

        return null;
    }
````
[Spigot] (https://github.com/starlis/EMC-CraftBukkit/blob/master/src/main/java/org/bukkit/craftbukkit/CraftServer.java)
````
    @Override
    @Deprecated
    public Player getPlayerExact(String name) {
        // Spigot start - replace whole method
        EntityPlayer player = playerList.playerMap.get(name);
        return player != null ? player.getBukkitEntity() : null;
        // Spigot end
    }
````

Have you looked for API to add players?

wizjany:
We've talked about this all at length in IRC, but he wants to change worldedit instead.

sk89q:
Adding a player to the server isn't really a good option IMO and admittedly this isn't either; however, this one has no unintended effects and just reverts the behavior of the session code to that of 5.x and earlier.

Europia79:
> The proper place to solve this would be in VirtualPlayers, not WorldEdit. Otherwise you'll always have players that aren't really there issuing commands and some plugins might not like it.

"**issuing commands**": This is the real heart of the issue.

Why is an isActive() check being run for a player that has just run a command ? They just ran the command, therefore, yes, they are active.

First, in order to run a command, you must have an instanceof ```CommandSender```... Anything beyond this requirement is a **command restriction.** For example, a lot of plugins further restrict command usage to instanceof ```org.bukkit.entity.Player```. In this situation, WorldEdit places a further restriction that the ```BukkitPlayer``` also be contained within ```CraftServer.playerList``` field.

This is the first of many reports that I received on Mar 8th, 2015:
> Hello, I've used your virtual players plugin for a very very long time and it has honestly been perfect. I'm sure like many people who use it one of virtual players main features was to be able to use world edit/guard through console.

> However in the latest version of world guard/world edit there is an issue occurring. Before this error a Virtual Player could easily go and select a region and then a second later world edit within the region it selected. However now there is the issue of the Virtual Player not being remembered to have selected a region. As before I can go select a region with a virtual player and then go to world edit within the region but I get the simple message "select a region first" so it appears that world edit is not allowing the Virtual Player to remember the data. It worked fine in the past versions of world guard/edit.

> I tried contacting the support of World Guard/ Edit and I was simply told that it could have something to do with this plugin. I heavily depend on this plugin and the version of World Guard which I'm using. I was wondering if you'd do me the great favour of fixing this if possible. I would also happily donate to the plugin to have this done.

> Thanks in advance and thanks for such a great plugin.

And I actually agree with @wizjany that this is weird and should be fixed by VirtualPlayers. But as you can see, I've been working on this problem for a very long time without success... And I don't get paid to do this. I just want to make the players happy (even if they're trying to do something that seems weird to me). I actually told this user to donate to WorldEdit instead.

So hopefully, we can get this fixed. I would be very happy. And I know a lot of VirtualPlayers + WorldEdit users would be happy too.

And also, I do realize that (to any given problem), there are many possible solutions. Like I said: I did try to apply a fix to VirtualPlayers first, by injecting them into ```CraftServer.playerList``` field. Which was really complicated and broke the entire VP plugin logic because they were no longer consider virtual, but they were considered real because they were inside the server's player list. And constructing a PlayerConnection was insanely difficult too.

This commit (removing the ```isActive()``` check) is probably the simplest solution. But I'm sure there are other solutions too. Like change the implementation details for isActive(). Like, have WorldEdit softdepend on VirtualPlayers. And use v2.0+ ```VirtualPlayer.getOnlinePlayers()``` (because i'm fixing the implementation details in v2.0 to be backwards compatible along with changing the return value). ```VirtualPlayers.getOnlinePlayers()``` would include all the real players on the server, and the VirtualPlayers too.

I'm sure there are other ways too. I'm open to suggestions. Thanks for helping !

TomyLobo:
Have you tried replacing the Server instance that {{Bukkit.getServer()}} returns by a wrapper?
That way you could just override {{getPlayerExact}}.

sk89q:
That can't be a good idea in the grand scheme of things.

Europia79:
> Why is an isActive() check being run for a player that has just run a command ? They just ran the command, therefore, yes, they are active.

This question isn't meant to be hostile or rhetorical. This something I honestly don't understand. You guys really do need the isActive() check there ? If so, why ?

To me, this commit removes duplicated code... Because isActive() already resides in the SessionTracker. Altho, I'm admittedly biased, ha !

Also, the only reason why I submitted this pull request was because I was under the impression that sk89q already approved the change in IRC last week.

Anyways, I really appreciate your help on this issue ! Thanks!

LaughNgamez:
Years later but yay!

PR #331, opened 2015-07-02
Add support for delegating permissions checks to Sponge when installed

zml2008:
This allows Sponge permissions plugins to be used for permissions checks without having to write a whole Sponge implementation. It should continue to work as previously if Sponge is not installed.

@luacs1998 since you did this system, any thoughts?

### DIFF
--- a/config/checkstyle/import-control.xml
+++ b/config/checkstyle/import-control.xml
@@ -54,6 +54,7 @@
       <allow pkg="org.apache.logging.log4j"/>
       <allow pkg="org.lwjgl"/>
       <allow pkg="io.netty.buffer"/>
+      <allow pkg="org.spongepowered.api" />
     </subpackage>
   </subpackage>
 </import-control>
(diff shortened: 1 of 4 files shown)


luacs1998:
You'd be better off speaking to SK about a proper implementation of WEPIF for Forge/Sponge - the current system is a crude one for FE interoperability. I'll need to sit down and look at this later if you want a more detailed response.

zml2008:
This isn't really intended to be a full system, just a quick hack until sponge support is properly in WE. @kenzierocks suggested I ping you -- I think he was mostly concerned about whether this would break anything FE does.

luacs1998:
I am planning to add code for FE 1.8 to not use WorldEdit-Forge (and by extension, this API) if Sponge and WorldEdit-Sponge is installed, and to use WorldEdit-Sponge instead.

Go ahead if you want, then.

PR #332, opened 2015-07-02
Fixed wrong torch metadata.

Pieter12345:
When using //rotate, normal torches, powered redstone torches and unpowered redstone torches would rotate unexpectedly. This was caused by swapped metadata 3 and 4 (North and South). This commit fixes that bug.


ewized:
Whats changed?

DarkArc:
@ewized "data" changed

wizjany:
Don't we have unit tests for this? either the old behavior or this new one should fail the test.
or the test is broken.

Pieter12345:
@wizjany I'm not sure, but on CraftBukkit 1.8.7, torches have the same metadata for noth, west, east and south as levers and buttons, which //rotate fine. All I know is that this commit makes the players on my redstone server happy again.

wizjany:
Fixed in 1e6e54b85a050309f053d898200f29fcb5569826

PR #333, opened 2015-07-08
Update forge world mappings

clienthax:


worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorld.java
```diff
@@ -238,7 +238,7 @@ public boolean regenerate(Region region, EditSession editSession) {
                 try {
                     u = ChunkProviderServer.class.getDeclaredField("field_73248_b"); // chunksToUnload
```
Lunatrius: Missed the comments. Anyway, wouldn't it be easier to just use:
```java
Field u = ReflectionHelper.findField("field_73248_b", "droppedChunksSet");
```

clienthax: Possibly, however this works and is inline with the original code, i remember having issues with reflectionhelper screwing in the past.

wizjany:
When were these renamed?

clienthax:
Mcpbot says 2 years ago which seems a bit out of range..

wizjany:
yea kinda doubt that since these were at least working on 1.7. i don't know if this changed at the same time as 1.8 or later though, because if it's working on one version but not another we should update deps and require the correct version

clienthax:
This is only a issue when running in dev envs btw, normally it uses the obfuscated mappings

TomyLobo:
2 years... that would be the [1.6 release] (http://minecraft.gamepedia.com/Versions)

kenzierocks:
I guess I never used regenerate in development...

kenzierocks:
If you rename the comments then I think this is good.

zml2008:
Pretty sure we'd have to become a coremod to have an AT, which is really something we should avoid  if possible.

kenzierocks:
That's true, we can just use reflection then.

TomyLobo:
excuse the uneducated question: what is AT?

clienthax:
@zml normal mods can use AT's but not class transformers

kenzierocks:
@TomyLobo access transformers, can modify modifiers.

TomyLobo:
modifying modifiers at runtime? well, we've been removing "private" and "final" modifiers from fields for a long time, without the need for a core mod

kenzierocks:
This does it without reflection, when the bytecode is loaded I believe.

TomyLobo:
Which doesn't help, since you still have to use reflection, no?

kenzierocks:
Nope! ForgeGradle applies them in development so you can compile against the field.

TomyLobo:
So if all plugins have to be aware of the fact that any other plugin may make any field public, why don't they just make everything public when installing forge, with something like that plugin I linked?
Why have people make lists of fields to make public?

sk89q:
Various libraries do special things with public fields so it's potentially dangerous.

TomyLobo:
good point.
anyway, sorry for putting you off topic so much :)

Lunatrius:
You don't need a core mod, you can use FMLAT instead.

kenzierocks:
No longer needed, 5b45c7cd9e789a3a65b786bae7b2554968bdba53.

PR #334, opened 2015-07-15
Fix unfortunate typo in AffineTransform

jedediah:
It's upstream in javaGeom as well.


TomyLobo:
word of advice: Don't PR from your master branch, because if you move that branch forward, the same is going to happen to a PR from that branch.

PR #335, opened 2015-07-16
Option to register permissions with Bukkit

jedediah:
Config option to have WE register all of its command permissions with Bukkit. This allows `no-op-permissions` to work properly when other plugins test for WE perms directly through the Bukkit API.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitConfiguration.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitConfiguration.java
@@ -30,6 +30,7 @@ import java.io.File;
 public class BukkitConfiguration extends YAMLConfiguration {
 
     public boolean noOpPermissions = false;
+    public boolean registerPermissions = false;
     private final WorldEditPlugin plugin;
 
     public BukkitConfiguration(YAMLProcessor config, WorldEditPlugin plugin) {
@@ -41,6 +42,7 @@ public class BukkitConfiguration extends YAMLConfiguration {
     public void load() {
         super.load();
         noOpPermissions = config.getBoolean("no-op-permissions", false);
+        registerPermissions = config.getBoolean("register-permissions", false);
         migrateLegacyFolders();
     }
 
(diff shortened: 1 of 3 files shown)


wizjany:
er, what exactly is wrong with no-op-permissions?
and what do you mean by "other plugins test for WE perms"?

jedediah:
Set `no-op-permissions: true` and use the Bukkit API i.e. `Permissible.hasPermission` to test for a WE permission on an operator who does not have it. It will claim that they have the permission, even though they shouldn't. The logic that makes `no-op-permissions` work is internal to WE, so it doesn't affect the Bukkit permission system. However, permissions can be registered with Bukkit so that ops do not have them by default. This patch adds an option for WE to do that automatically, so that testing WE perms through the Bukkit API will give a correct result.

wizjany:
I believe that is intentional...the option, as I understand, means worldedit won't automatically give permissions to ops... i.e. won't bypass checking the backend permissions system (whatever it's set to). I think the distinction is important because in the absence of a unified permissions system like bukkit, this would be equivalent to PEX/GM/zPerms/etc deciding what to do with ops - it's not really worldedit's job to mess with the permissions backend imo.

I also don't think other plugins should necessarily be checking for worldedit permissions, but that may be a separate issue.

jedediah:
WE always bypasses the backend, regardless of the `no-op-permissions` setting. If it didn't bypass the backend, then that setting would have no effect, because Bukkit would grant all perms to ops in place of WE.

As I see it, the issue is that WE doesn't interpret permissions the same way as the backend its using, at least not when the backend is Bukkit, e.g. https://github.com/sk89q/WorldEdit/blob/master/worldedit-bukkit/src/main/java/com/sk89q/wepif/DinnerPermsResolver.java#L148

If you pass that method an op player and an unregistered permission, it will report that the permission is "denied by default", but Bukkit allows by default for unregistered perms on operators.

Mind you, this will also happen if you use any `*` permissions, since Bukkit knows nothing about those. But that is a more obvious difference, and more "fail safe" since it denies unexpectedly, rather than allowing unexpectedly.

I guess it comes down to what the WE permission system is supposed to do exactly, and how it is supposed to relate to the backend. Currently, it uses the backend only as a string database, and those strings mean completely different things to Bukkit and WE. It would be convenient, and probably safer, if WE at least guaranteed that anything denied by WE will also be denied by the backend. And this patch achieves that, to the best of my knowledge.

kenzierocks:
Unless someone demonstrates a need for this, we won't be adding this feature. To quote wizjany:
>  LP picks up our perms from our commands anyway, doesnt need bukkit registration. bukkit reg also breaks in many case that we would have to cover.

PR #337, opened 2015-10-12
More efficient chunk calculation of CuboidRegion

rlf:
The CuboidRegion currently creates 16x16 vectors of identical coordinates, and adds them to a set.
Could be more efficient.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/CuboidRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/CuboidRegion.java
@@ -300,10 +300,9 @@ public class CuboidRegion extends AbstractRegion implements FlatRegion {
         Vector min = getMinimumPoint();
         Vector max = getMaximumPoint();
 
-        for (int x = min.getBlockX(); x <= max.getBlockX(); ++x) {
-            for (int z = min.getBlockZ(); z <= max.getBlockZ(); ++z) {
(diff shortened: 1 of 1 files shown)


TomyLobo:
Have you compared the actual performance?

ewized:
What if you also statically import `ChunkStore.CHUNK_SHIFTS`?

rlf:
The static import would only be syntactical sugar - the optimizer would not care.
The code changes are quite small, not meritting a new test case in my oppinion, since it should be covered by any tests already verifying the getChunks() method.

The original method would do this:
```
  for all x,y within the region:
    create a vector object of (x >> 4, y >> 4) - object creation - heavy operation
    put this vector in a set (255/256 of these would be duplicates) - semi-heavy operation
  return the set containing ~ 1/256 of the original vectors
```

IMHO - this is a no-brainer - a simple fix with a very small risk

TomyLobo:
more over, the compiler will not care :)

sk89q:
Looks good. I will check this out and merge when I make a WE release in a few days.

DarkArc:
Just wanted to point out getChunkCubes() -- the method immediately following -- is almost the exact same method

sk89q:
That's a good point. Probably need to change that one too.

wizjany:
merged in 03f8843

TomyLobo:
@wizjany Did you add a unit test to cover this?

PR #294, opened 2014-05-24
//schematic list with pages

BangL:
New command argument: "page"
new syntax:
	//schematic list [page] [flags]

"schematic list" shows always page 1 now.

PR #338, opened 2015-12-28
Add paging to //schematics.

kenzierocks:
Re-implementation of #294. ~~Untested.~~ Working perfectly.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
@@ -24,8 +24,6 @@ import com.sk89q.minecraft.util.commands.CommandContext;
 import com.sk89q.minecraft.util.commands.CommandException;
 import com.sk89q.minecraft.util.commands.CommandPermissions;
 import com.sk89q.worldedit.EditSession;
-import com.sk89q.worldedit.util.io.file.FilenameException;
-import com.sk89q.worldedit.util.io.file.FilenameResolutionException;
 import com.sk89q.worldedit.LocalConfiguration;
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -244,14 +251,16 @@ public void formats(Actor actor) throws WorldEditException {
...
+            flags = "dnp",
             help = "List all schematics in the schematics directory\n" +
                     " -d sorts by date, oldest first\n" +
-                    " -n sorts by date, newest first\n"
+                    " -n sorts by date, newest first\n" +
+                    " -p <page> prints the requested page\n"
     )
     @CommandPermissions("worldedit.schematic.list")
-    public void list(Actor actor, CommandContext args) throws WorldEditException {
+    public void list(Actor actor, CommandContext args, @Switch('p') @Optional("1") int page) throws WorldEditException {
```
TomyLobo: hmm does that @Optional("1") mean "1 optional argument to -p" or does it mean "-p is optional and requires one argument".
In the first case, your code doesn't match your help blurb
In the second case, @Optional is veeeery strange...

kenzierocks: 3) `@Optional("1")` means the default value is `"1"`.

TomyLobo: oh, for some reason I missed those quotes. makes sense, don't mind me :)

kenzierocks:
@wizjany or @sk89q does this look good to merge?

PR #340, opened 2016-02-02
Initial work on WorldEdit Sponge

DarkArc:
This is my initial work on making a Sponge API based WorldEdit. 

This implementation does not try to fully implement purely with the Sponge API, instead this implementation attempts to leverage those parts of the Sponge API which are adequately developed such that WorldEdit can use them. In cases where this is not possible, the implementation tries to find a suitable functionality in SpongeForge, and then drops down to NMS/Forge code.

This should be considered an intermediary step in moving fully from Forge to the Sponge API. This is also being done due to some recent issues with WorldEdit and Sponge, where Sponge is attempting to calculate causation for the edits WorldEdit is performing. We can significantly alleviate strain on Sponge based servers by instead providing the cause here while performing our block change operations.

I should note, I've also made an effort to keep all NMS/SpongeForge code separated from any pure-sponge components. The thought being, checking, and updating a few minor areas will not be as significant of a workload as perusing the entire project for broken code.

TODO:
- [x] Finish command registration/processing code
- [x] Implement the ID Helper
- [ ] WorldEdit CUI
- [x] Fix the online players cache
- [x] Entity rotation code
- [ ] General cleanup
- [x] Fix some message outputs
- [x] Fix tile entity creation
- [x] Fix entity creation

Note: This PR is purely for centralizing discussion, and as such is setup against forge1.8 instead of master.


settings.gradle
```diff
@@ -1,3 +1,3 @@
 rootProject.name = 'worldedit'
 
-include 'worldedit-core', 'worldedit-bukkit', 'worldedit-forge'
\ No newline at end of file
+include 'worldedit-core', 'worldedit-bukkit', 'worldedit-forge', 'worldedit-sponge'
```
kenzierocks: While you're editing this, might as well add a newline.

kenzierocks:
What's `Ant` being used for? I'd like to keep everything in `Gradle`, even though it can natively load `Ant` build files.

worldedit-sponge/build.gradle
```diff
@@ -0,0 +1,99 @@
+apply plugin: 'eclipse'
```
kenzierocks: Could we merge some of this with the `worldedit-forge` build.gradle? I'll do it if you're open to the idea.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/NBTConverter.java
```diff
@@ -0,0 +1,236 @@
...
+import com.sk89q.jnbt.*;
+import net.minecraft.nbt.*;
+
+import java.util.*;
+import java.util.Map.Entry;
+
+/**
+ * Converts between JNBT and Minecraft NBT classes.
+ */
+final class NBTConverter {
```
kenzierocks: I wonder if we can rewrite this to adapt between `JNBT` and the Sponge Data API? That might work better going forwards.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.sponge;
+
+import com.sk89q.worldedit.util.PropertiesConfiguration;
+
+import java.io.File;
+
+public class SpongeConfiguration extends PropertiesConfiguration {
```
kenzierocks: It'd be great if we could use Sponge's built in config stuff instead of this.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
```diff
@@ -0,0 +1,267 @@
...
+import org.spongepowered.api.world.Location;
+import org.spongepowered.api.world.World;
+import org.spongepowered.mod.mixin.core.event.state.MixinEventServerAboutToStart;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * The Sponge implementation of WorldEdit.
+ */
+@Plugin(id = "SpongeWorldEdit.MOD_ID", name = "WorldEdit", version = "%VERSION%")
```
kenzierocks: The ID string is wrong. Should be without quotes if you wanted the `MOD_ID` field below.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeConfiguration.java
DarkArc: Possibly, I haven't looked much into Sponge's configuration API.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/NBTConverter.java
DarkArc: This may be possible for some things, but I don't think we can really do that as a whole, as it would remove our ability to interface with any tag (afaik).

worldedit-sponge/build.gradle
DarkArc: I really don't like working with gradle build scripts all that much, so if you can think of a nice way to do it I won't object. Although, it might be nice to keep them separate that way the Sponge and Forge versions of WorldEdit can differ.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/NBTConverter.java
kenzierocks: I think we should be able to serialize the whole thing as a DataContainer and read that but maybe not. If it looks like something we'd like to do we can cross that bridge then.

PR #341, opened 2016-02-07
Add FMP compatiblity

kenzierocks:
Since https://github.com/Chicken-Bones/ForgeMultipart/pull/272 is not being merged any time soon, and FMP compatibility would be really useful, this PR adds the ability to correctly set FMP blocks to WorldEdit.

### DIFF
--- a/worldedit-forge/src/main/resources/mcmod.info
+++ b/worldedit-forge/src/main/resources/mcmod.info
@@ -14,7 +14,8 @@
         "Forge@[${forgeVersion},)"
     ],
     "dependencies": [
-        "Forge@[${forgeVersion},)"
+        "Forge@[${forgeVersion},)",
+        "ForgeMultipart"
     ],
     "dependants": []
 }]
--- a/config/checkstyle/import-control.xml
+++ b/config/checkstyle/import-control.xml
@@ -48,6 +48,7 @@
     </subpackage>
 
     <subpackage name="forge">
+      <allow pkg="codechicken.multipart"/>
       <allow pkg="cpw"/>
       <allow pkg="net.minecraft"/>
       <allow pkg="net.minecraftforge"/>
--- a/worldedit-forge/build.gradle
+++ b/worldedit-forge/build.gradle
@@ -15,7 +15,8 @@ apply plugin: 'forge'
 
 dependencies {
     compile project(':worldedit-core')
-    testCompile group: 'org.mockito', name: 'mockito-core', version:'1.9.0-rc1'
+    compile group: 'codechicken', name: 'ForgeMultipart', version: '1.2.0.345', classifier: 'dev'
+    testCompile group: 'org.mockito', name: 'mockito-core', version: '1.9.0-rc1'
 }
 
 minecraft {
--- a/worldedit-forge/build.gradle
+++ b/worldedit-forge/build.gradle
@@ -13,9 +13,16 @@ buildscript {
 
 apply plugin: 'forge'
 
+repositories {
+    maven {
+        name 'forge'
+        url 'http://files.minecraftforge.net/maven/'
+    }
+}
+
 dependencies {
     compile project(':worldedit-core')
-    compile group: 'codechicken', name: 'ForgeMultipart', version: '1.2.0.345', classifier: 'dev'
+    compile group: 'codechicken', name: 'ForgeMultipart', version: '1.7.10-1.2.0.345', classifier: 'dev'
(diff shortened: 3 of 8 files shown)


DarkArc:
Is there a more general way we could do this, that doesn't add extra dependencies to WorldEdit? This feels more like something that should be an optional extension of WorldEdit, not so much a part of the platform.

What happens when there's another mod with special behavior we need to handle? Do we add yet another dependency, and even more code to handle that particular mod?

kenzierocks:
There might be a more general way to do this. If we add some events to WE to fire when TileEntities are placed, I could write a mod to hook into that.

worldedit-forge/src/main/resources/mcmod.info
```diff
@@ -14,7 +14,8 @@
         "Forge@[${forgeVersion},)"
     ],
     "dependencies": [
-        "Forge@[${forgeVersion},)"
+        "Forge@[${forgeVersion},)",
+        "ForgeMultipart"
```
wizjany: is FMP not part of forge? what happens when there's a forge server without FMP then, does WE not load due to a missing dependency?

luacs1998: FMP is not part of forge, and yes, the server will crash if you install WE without installing FMP this way,

kenzierocks: No, the server will not crash. This is a soft-depend, `requiredMods` is for hard.

DarkArc:
I've been thinking about this, in particular, with multipass. WorldEdit works with a lot of hard coded data in BlockType. 

I honestly think the way to go, in the long term, is to refactor our code so that users can configure our internal data (manually add IDs to things like the shouldPlaceLast list), and mods can hook in via our API and add special editing steps/behavior. 

Thus the editing for block types which use this tile entity could be overridden by a FMP WorldEdit Compatibility mod. Before this can happen though, we need to revamp our concept of block/item ids, it's just not ready for such a system.

kenzierocks:
As a holdover, it might be a good idea to start having WE events that fire either (`1`) on the implementation event bus or (`2`) on another WE event bus. That could help with making this PR a separate WE plugin. If someone would like to vote or discuss which of `1` or `2` is better, then I'm willing to PR it.

DarkArc:
Well, with the Sponge implementation, the block is set with a cause of WorldEdit.

kenzierocks:
This fix is for Forge, especially on clients using FMP, so Sponge doesn't factor much into this right now.

DarkArc:
Sponge relies on Forge, so we could potentially require Sponge if they want that bug fixed, rather than doing an invasive temporary fix.
		
        
		
		 ---- On Thu, 11 Feb 2016 20:15:32 -0800  notifications@github.com  wrote ----This fix is for Forge, especially on clients using FMP, so Sponge doesn't factor much into this right now.

wizjany:
I would much rather depend on FMP than throw yet more events imo.

kenzierocks:
The reason I'm against Sponge is that it currently doesn't support clients, so requiring someone to install it for singleplayer seems a little overkill and potentially dangerous.

DarkArc:
I still don't like the idea of depending on FMP, and adding this code... But I suppose it can be reverted after a more effective fix can be put in place.

wizjany:
the mod itself isn't even depending on FMP from what I understand, it's just a compile time dependency.

kenzierocks:
Correct.

DarkArc:
@wizjany I'm aware, I'm just not a fan of adding more special case code to WorldEdit itself. 

If no one else has any objection though, I mean, I guess go ahead and merge it. I'm not going to be terribly upset if this is merged or anything, I'd just rather see it done in a more open ended way.

kenzierocks:
We can look at this again if we need to come up with a better solution. This might be phased out in 1.8 anyways, since [CB is looking to hook into `setTileEntity` or similar.](https://github.com/Chicken-Bones/ForgeMultipart/issues/145#issuecomment-94252624)

Nentify:
Where can I get this build?

kenzierocks:
It's the latest 1.7.10 build on CurseForge.

Nentify:
That's 6.0, I had to compile myself to get the FMP integration. Thanks though.

kenzierocks:
Or you can get very latest off of https://builds.enginehub.org/

Nentify:
Doesn't work either

kenzierocks:
...yes it does? It's on the `forge-archive/1.7.10` branch. I just downloaded it now.

PR #342, opened 2016-02-16
Add support to /forestgen and //forest to use patterns for tree types

Hugobros3:
Just like for block patterns I wanted to be able to generate many type of trees in a single command. I did my best to make everything clean and respect the guidelines provided, I should stress that this is the first time I do a pull request (git is confusing) to a "large" project.


worldedit-core/src/main/java/com/sk89q/worldedit/util/GenericRandomList.java
```diff
@@ -0,0 +1,87 @@
...
+            if (r <= (offset + chance.getChance()) / max) {
+                return chance.getObject();
+            }
+            offset += chance.getChance();
+        }
+
+        throw new NullPointerException();
+    }
+
+    class Chance {
```
TomyLobo: This entire file looks overengineered.

Why not just list.remove(random.nextInt(list.size())) or something?

That'd be a generic function in some util class instead of an entire generic class

Hugobros3: Well, I based this part on the RandomPatern and RandomPatternParser classes, since as I see it it only makes sense to have a single way of doing things for both. Adding weighter things twice wouldn't work very well in the end since you can have percentages with a few digits and you'd end up with huge lists thrashing the garbage collection for no reason

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/TreeTypesParser.java
```diff
@@ -0,0 +1,74 @@
...
+        for (String token : input.split(",")) {
+            TreeType type;
+            double chance;
+
+            // Parse special percentage syntax
+            if (token.matches("[0-9]+(\\.[0-9]*)?%.*")) {
+                String[] p = token.split("%");
+
+                if (p.length < 2) {
+                    throw new InputParseException("Missing the type after the % symbol for '" + input + "'");
```
wizjany: this message is completely ignored.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/TreeTypesParser.java
```diff
@@ -0,0 +1,74 @@
...
+                }
+            } else {
+                chance = 1;
+                type = TreeGenerator.lookup(token);
+            }
+
+            if (type != null)
+                treeTypes.add(type, chance);
+            else
+                throw new InputParseException(String.format("Can't recognize tree type '%s' -- choose from: %s", token,
```
wizjany: same here.

Hugobros3: I just checked and they are taken into account, it's the same as the RandomPatternParser class already into the codebase really
http://image.noelshack.com/fichiers/2016/07/1455747479-proof.png

TomyLobo: I don't get why it says "null" there... must be a string literally containing "null" in that "token" variable...

TomyLobo:
You opened a pull request, which contains code changes (crazy ones, i might add)
Open an issue if you just want to report an issue.

wizjany:
I think you read that wrong @TomyLobo 
The merge was definitely messed up though.

Hugobros3:
I wanted to push that feature a while ago, trying to mix and match existing classes to follow the codestyle. Apparently people didn't like it and I forgot about it. I updated my branch yesterday ( something unrelated, I'm looking into porting WorldEdit for my own minecraft-like game ) and never thought to close my old PR.

I'm not sure why you'd say my merge was messed up though, what did I screw up ?

PR #343, opened 2016-02-21
Workaround for broken wooden/iron door rotation.

trichner:
- enables rotation for fence gates, all variants
- fixes rotation of wooden (only oak) / iron doors
- also works for double doors
- does NOT fix the newer doors
- exempts them from tests as they are now broken for those blocks


worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
```diff
@@ -110,6 +114,9 @@ public static BaseBlock transform(BaseBlock block, Transform transform, BlockReg
         return transform(block, transform, registry, block);
     }
 
+
+    private static Set<Integer> LEGACY_ROTATION = Sets.newHashSet(BlockID.WOODEN_DOOR,BlockID.IRON_DOOR);
```
kenzierocks: Formatting issue, needs a space after the comma.

worldedit-core/src/test/java/com/sk89q/worldedit/blocks/BlockDataTest.java
```diff
@@ -29,7 +29,7 @@
 
 public class BlockDataTest {
 
-    @Test
+    //@Test
```
kenzierocks: Use `@Ignore` instead of commenting it out.

TomyLobo: Use neither.
Instead make your code work with the test or provide a reason why it cannot.

TomyLobo:
You forgot to mention why a workaround is necessary and for what problem

kenzierocks:
@tomylobo Doors don't rotate in WorldEdit right now. Because they're double blocks rotation is a bit of a hack.

TomyLobo:
@kenzierocks got it. But this belongs in the description, not in a comment 2 pages down after someone happened to ask.

Btw, I realize the [flip/180-degree-rotation congruency check](https://github.com/sk89q/WorldEdit/blob/master/worldedit-core/src/test/java/com/sk89q/worldedit/blocks/BlockDataTest.java#L43-L48) is probably toast for the doors, unless you can fix flip as well.
I'm fine with adding a fine-grained exemption for the doors for now.

TomyLobo:
Hmm I just noticed that check is not testing whether flips are order-independent...
I have to add that once this PR is merged.

trichner:
Added exemptions for tests and generalized the rotation.

wizjany:
How much of this is still necessary? Pretty sure that single doors at least are working now.
Do double doors still have hinge issues then? Not quite sure what the issue was.

TomyLobo:
It's been 2.5 years, I'd just close it and see if anyone complains :)

me4502:
If this is still an issue, please update for latest source.

PR #344, opened 2016-03-17
Add RED_SANDSTONE_STAIRS to rotate function

Upd4ting:
This type of stairs wasn't getting rotated :)


wizjany:
1) the official implementation of the api no longer uses the BlockData class, it uses the bundled LegacyWorldData that's read from blocks.json
2) BlockData and the WorldData class are cross-checked when running post-compile testing, so changing just one breaks tests.
3) you didn't even bother adding them to the flip code.

PR #345, opened 2016-03-19
Initial work on WorldEdit Sponge

DarkArc:
This is my initial work on making a Sponge API based WorldEdit. 

This implementation does not try to fully implement purely with the Sponge API, instead this implementation attempts to leverage those parts of the Sponge API which are adequately developed such that WorldEdit can use them. In cases where this is not possible, the implementation tries to find a suitable functionality in SpongeForge, and then drops down to NMS/Forge code.

This should be considered an intermediary step in moving fully from Forge to the Sponge API. This is also being done due to some recent issues with WorldEdit and Sponge, where Sponge is attempting to calculate causation for the edits WorldEdit is performing. We can significantly alleviate strain on Sponge based servers by instead providing the cause here while performing our block change operations.

I should note, I've also made an effort to keep all NMS/SpongeForge code separated from any pure-sponge components. The thought being, checking, and updating a few minor areas will not be as significant of a workload as perusing the entire project for broken code.

TODO:
- [x] Finish command registration/processing code
- [x] Implement the ID Helper
- [x] WorldEdit CUI
- [x] Fix the online players cache
- [x] Entity rotation code
- [x] General cleanup
- [x] Fix some message outputs
- [x] Fix tile entity creation
- [x] Fix entity creation
- [x] Create a more proper "Sponge" configuration

Note: This PR is purely for centralizing discussion, and as such is setup against forge1.8 instead of master.

Previous discussion may be found on #340, due to the merge of feature/forge1.8, the request was automatically closed.


worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.sponge;
+
+import com.sk89q.worldedit.util.PropertiesConfiguration;
+
+import java.io.File;
+
+public class SpongeConfiguration extends PropertiesConfiguration {
```
kashike: Why not use configurate?

DarkArc: This was just taken from the Forge branch. I suppose I could add a more native configuration option as another todo item.

worldedit-sponge/build.gradle
```diff
@@ -0,0 +1,90 @@
...
+        maven { url = "https://oss.sonatype.org/content/repositories/snapshots/" }
+        jcenter()
+    }
+
+    dependencies {
+        classpath 'net.minecraftforge.gradle:ForgeGradle:2.1-SNAPSHOT'
+    }
+}
+
+apply plugin: 'net.minecraftforge.gradle.forge'
```
Minecrell: You should really use a Vanilla workspace here so it can also run on SpongeVanilla. Currently, probably 95% can already work on SpongeVanilla, and there is no point requiring SpongeForge for a Sponge implementation.

wizjany: 95% seems high when we still throw int block ids around everywhere in the api

kenzierocks: @minecrell said SpongeVanilla, not Sponge API. SpongeVanilla compatibility means we don't use Forge, not that we use SpongeAPI.

DarkArc: Actually there is a fairly high probability that no Forge specific functionality is used. I'll play around with it at some point, unless someone else wants to beat me to it.

wizjany: Er, if we use the spongevanilla workspace as minecrell says or whatever, does that also mean it's compatible with spongeforge then? i was under the impression that being compatible with both would mean only using the api (or maybe things in common at most - but that requires compiling two different ways?)

kashike: Being compatible with both just means not using any Forge-added methods (to `Block`, etc), or any of the Forge API.

DarkArc: @wizjany it should be equivalent to using NMS in Bukkit, at that point.

Minecrell: SpongeVanilla uses exactly the same SRG/MCP environment at runtime, just with the little difference that a) it will always be only on the server side and b) Forge modifications or additions to Minecraft classes are missing. From what I can tell, almost everything here is just Vanilla, but it's better to use a Vanilla workspace directly so you notice the usage of Forge methods directly.

Except very few places where Forge actually moves logic or modifies existing methods the code will always work on both platforms. It would be essentially exactly the same as the workspace used by SpongeCommon.

There is an example for a Vanilla server workspace at https://github.com/Minecrell/worldedit-spongevanilla/blob/spongevanilla/worldedit-forge/build.gradle

DarkArc:
I just killed the regeneration code, it seems to cause problems on Sponge anyways.

Minecrell:
@DarkArc By the way, there is no need to use reflection because SpongeVanilla also supports access transformers the same way as Forge does. :)

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlayer.java
```diff
@@ -0,0 +1,227 @@
...
+        sendColorized(msg, TextColors.LIGHT_PURPLE);
+    }
+
+    @Override
+    public void printError(String msg) {
+        sendColorized(msg, TextColors.RED);
+    }
+
+    private void sendColorized(String msg, TextColor formatting) {
+        for (String part : msg.split("\n")) {
```
Minecrell: Splitting this should not be necessary, texts can contain new lines

worldedit-sponge/src/main/ant/build.xml
```diff
@@ -0,0 +1,150 @@
+<project name="WorldEdit-Forge" default="main">
```
kashike: Uh, what is this ant build for?

DarkArc: Yeah, I have no idea, carry over from Forge, probably just need to kill it.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlayer.java
DarkArc: This may have implications on formatting though. If you note, it reformats after every new line. There might be some text that relies on that logic, and actually, I believe that even these formatted options may need to use desearalize. I've noticed some oddities.

worldedit-sponge/src/main/ant/build.xml
wizjany: wow didn't even realize that was still there

me4502:
Making a comment to see if I still have access to the repo.

TomyLobo:
> me4502 commented 3 hours ago • edited Collaborator

What makes you think you don't have access and how would a comment help determine that?

me4502:
@TomyLobo I didn't at the time, I was checking if I had the little 'Collaborator' badge.

TomyLobo:
I see
let me remove our off-topic comments then

build.gradle
```diff
@@ -90,7 +90,7 @@ subprojects {
         archives javadocJar
     }
 
-    if (!name.equals('worldedit-forge')) {
+    if (!(name.equals('worldedit-forge') || name.equals('worldedit-sponge'))) {
```
kenzierocks: is this the right logic? should `worldedit-sponge` already be included?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/CUIChannelHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import com.sk89q.worldedit.LocalSession;
+import org.spongepowered.api.Platform;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.network.*;
+
+import java.nio.charset.StandardCharsets;
+
+public class CUIChannelHandler implements RawDataListener {
+    public static final String CUI_PLUGIN_CHANNEL = "WECUI";
```
kenzierocks: Could we extract this to `-core`? That way all implementations use the same string.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeBiomeRegistry.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Nullable
+    @Override
+    public BiomeData getData(BaseBiome biome) {
+        return new SpongeBiomeData(IDHelper.resolveBiome(biome.getId()));
+    }
+
+    /**
+     * Cached biome data information.
```
kenzierocks: I wouldn't call this cached, it's just a wrapper.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/CUIChannelHandler.java
DarkArc: The problem is there isn't a common place to put it. Most of the implantations put it on the plugin. Static data on interfaces is a Java 8 thing, I guess we could use an abstract class, but that seems really unnecessary.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeBiomeRegistry.java
DarkArc: Fair enough

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlatform.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Nullable
+    @Override
+    public World matchWorld(World world) {
+        if (world instanceof SpongeWorld) {
+            return world;
+        } else {
+            for (org.spongepowered.api.world.World ws : Sponge.getServer().getWorlds()) {
+                if (ws.getName().equals(world.getName())) {
```
kenzierocks: Do we have uuid info to compare here?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlatform.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Override
+    public String getVersion() {
+        return mod.getInternalVersion();
+    }
+
+    @Override
+    public String getPlatformName() {
+        return "Sponge-Official";
```
kenzierocks: ...as opposed to what? Why the 'Official'?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlayer.java
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }
+
+    @Override
+    public void setPosition(Vector pos, float pitch, float yaw) {
+        org.spongepowered.api.world.Location<World> loc = new org.spongepowered.api.world.Location<>(
+                this.player.getWorld(), pos.getX(), pos.getY(), pos.getZ()
+        );
+
+        this.player.setLocationAndRotation(
```
kenzierocks: Funky formatting here.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlatform.java
DarkArc: We don't currently provide a concept of a world UUID in our abstraction:
https://github.com/sk89q/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorld.java
```diff
@@ -0,0 +1,307 @@
...
+        List<Entity> entities = new ArrayList<>();
+        for (org.spongepowered.api.entity.Entity entity : getWorld().getEntities()) {
+            entities.add(new SpongeEntity(entity));
+        }
+        return entities;
+    }
+
+    protected abstract void applyEntityData(org.spongepowered.api.entity.Entity entity, BaseEntity data);
+
+    private static final Cause ENTITY_SPAWN_CAUSE = Cause.source(SpawnCause.builder().type(SpawnTypes.PLUGIN).build()).build();
```
kenzierocks: Should we include our plugin instance in this?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldData.java
```diff
@@ -0,0 +1,47 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.sponge;
+
+import com.sk89q.worldedit.world.registry.BiomeRegistry;
+import com.sk89q.worldedit.world.registry.LegacyWorldData;
+
+/**
+ * World data for the Forge platform.
```
kenzierocks: Typo?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlatform.java
DarkArc: This is just convention:
https://github.com/sk89q/WorldEdit/blob/master/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitServerInterface.java#L156

---- On Wed, 18 May 2016 17:34:37 -0700 notifications@github.com wrote ----

In worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlatform.java:

> +        hookingEvents = true;
> +    }
> +
> +    @Override
> +    public SpongeConfiguration getConfiguration() {
> +        return mod.getConfig();
> +    }
> +
> +    @Override
> +    public String getVersion() {
> +        return mod.getInternalVersion();
> +    }
> +
> +    @Override
> +    public String getPlatformName() {
> +        return "Sponge-Official";
...as opposed to what? Why the 'Official'?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
```diff
@@ -0,0 +1,311 @@
...
+ * The Sponge implementation of WorldEdit.
+ */
+@Plugin(id = SpongeWorldEdit.MOD_ID, name = "WorldEdit", version = "%VERSION%")
+public class SpongeWorldEdit {
+
+    @Inject
+    private java.util.logging.Logger logger;
+
+    public static final String MOD_ID = "worldedit";
+    public static final String CUI_PLUGIN_CHANNEL = "WECUI";
```
kenzierocks: At least share this with the CUI class.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
```diff
@@ -0,0 +1,311 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * The Sponge implementation of WorldEdit.
+ */
+@Plugin(id = SpongeWorldEdit.MOD_ID, name = "WorldEdit", version = "%VERSION%")
+public class SpongeWorldEdit {
+
+    @Inject
+    private java.util.logging.Logger logger;
```
kenzierocks: Isn't it the Log4j logger that gets injected?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
```diff
@@ -0,0 +1,311 @@
...
+
+    @Listener
+    public void postInit(GamePostInitializationEvent event) {
+        logger.info("WorldEdit for Sponge (version " + getInternalVersion() + ") is loaded");
+    }
+
+    @Listener
+    public void serverAboutToStart(GameAboutToStartServerEvent event) {
+        if (this.platform != null) {
+            logger.warning("FMLServerStartingEvent occurred when FMLServerStoppingEvent hasn't");
```
kenzierocks: Typo.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
```diff
@@ -0,0 +1,311 @@
...
+     * Get the working directory where WorldEdit's files are stored.
+     *
+     * @return the working directory
+     */
+    public File getWorkingDir() {
+        return this.workingDir;
+    }
+
+    /**
+     * Get the version of the WorldEdit-for-Forge implementation.
```
kenzierocks: Typo.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/ThreadSafeCache.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public void run() {
+        Set<UUID> onlineIds = new HashSet<>();
+
+        for (Player player : Sponge.getServer().getOnlinePlayers()) {
+            onlineIds.add(player.getUniqueId());
+        }
+
+        this.onlineIds = new CopyOnWriteArraySet<>(onlineIds);
```
kenzierocks: Why not just build an immutable set if you're going to copy it anyways?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/ThreadSafeCache.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.UUID;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+/**
+ * Caches data that cannot be accessed from another thread safely.
+ */
+public class ThreadSafeCache implements Runnable {
+
+    private static final ThreadSafeCache INSTANCE = new ThreadSafeCache();
+    private Set<UUID> onlineIds = Collections.emptySet();
```
kenzierocks: This needs to be volatile or wrapped in `AtomicReference` to be really thread safe.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlatform.java
kenzierocks: Ok, this will do then.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongePlatform.java
kenzierocks: Ok.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldData.java
DarkArc: Yup

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/ThreadSafeCache.java
DarkArc: It can be modified, but isn't always. This saves the set from being copied at each point while constructing.

kenzierocks: Ah, I see.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
```diff
@@ -31,6 +31,7 @@
 import com.sk89q.worldedit.sponge.config.SpongeConfiguration;
 import com.sk89q.worldedit.sponge.nms.NMSHelper;
 import com.sk89q.worldedit.sponge.nms.SpongeNMSWorld;
+import org.apache.logging.log4j.Logger;
```
phroa: Docs actually say `org.slf4j.Logger` -- not sure if this `Logger` would work, though it *is* the right implementation class...

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/CUIChannelHandler.java
DarkArc: Actually, this could be moved to the WorldEdit class probably. However, lets move this in multipass / WorldEdit 7, if someone's trying to use one of the other variables it would be a breaking change.

build.gradle
kenzierocks: Never mind, misread.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
DarkArc: I'll change it for consistency sake.

worldedit-sponge/build.gradle
```diff
@@ -0,0 +1,91 @@
...
+        classpath 'net.minecrell:VanillaGradle:2.0.3-SNAPSHOT'
+        classpath 'net.minecraftforge.gradle:ForgeGradle:2.1-SNAPSHOT'
+    }
+}
+
+apply plugin: 'net.minecrell.vanilla.server.library'
+
+dependencies {
+    compile project(':worldedit-core')
+    compile 'org.spongepowered:spongeapi:4.+'
```
kenzierocks: Could you avoid this `+` here? If I remember correctly this breaks the auto-generated POM file (at least for validation by Central, if we care about that).

PR #350, opened 2016-05-17
Add support for 1.9 doors on MultiStageReorder

DonZitrO:
The MultiStageReorder algorithm uses the BlockType.shouldPlaceFinal set in order to attach the superior and inferior blocks of doors correctly. I also added the new doors to the BlockType.isRedstoneBlock collection. Those doors are problably missing from more collections or further configuration inside the BlockType class, but these two are the ones I'm sure of. This should allow to use fast=false and place the doors correctly.


DonZitrO:
As a bonus, after researching the code a bit, I also added the doors to the getBlockBagItem HashMaps.

PR #351, opened 2016-05-19
Fix rotation of item frames

FrozenBrain:
Minecraft 1.8 introduced a new tag named 'Facing' which holds the direction item frames face. Pre-1.8 'Dir' and 'Direction' were used instead.
See http://minecraft.gamepedia.com/Item_Frame#Data_values

Currently item frames pop off walls after rotating because they are facing the wrong direction. This commit fixes that issue.


wizjany:
Merged.

wizjany:
Merged.

PR #352, opened 2016-05-19
Fix MathUtils.dSin

FrozenBrain:
dSin(..) currently returns the **cosine** of angles that are not multiples of 90°.
This causes commands such as _//rotate 15_ to rotate the clipboard by 45° instead of 15°.


wizjany:
Merged.

PR #353, opened 2016-05-19
Fix secondary-ordinal direction vectors

FrozenBrain:
Some of the direction vectors in com.sk89q.worldedit.util.Direction are incorrect.
One example would be WEST_NORTHWEST which is currently equal to WEST_SOUTHWEST.
This commit fixes those vectors.


wizjany:
Merged.

PR #354, opened 2016-05-19
Apply transformations to NBT data of tile entities (fixes rotation of skulls)

FrozenBrain:
Skulls store their rotation in a NBT tag called 'Rot' if they're standing on the ground.
This commit modifies ExtentBlockCopy to set this tag to ensure that skulls are rotated correctly.
I'm not quite sure if this is the right place to do so though.


wizjany:
Merged.

PR #357, opened 2016-05-21
Minor improvements to Sponge implementation

Minecrell:
I made some minor improvements to the Gradle configuration and the main plugin class of the Sponge implementation that remove redundant parts (either because they're default or not needed for Sponge), or simplify the code:

- Use SpongeGradle for setting plugin version from Gradle
- Use Guice to manage configuration
- Use event filter annotation for interact event
- Set plugin description and URL
- Remove some redundant parts in the Gradle config

### DIFF
--- a/worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
+++ b/worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
@@ -32,15 +32,15 @@ import com.sk89q.worldedit.sponge.config.SpongeConfiguration;
 import com.sk89q.worldedit.sponge.nms.NMSHelper;
 import com.sk89q.worldedit.sponge.nms.SpongeNMSWorld;
 import org.slf4j.Logger;
-import org.spongepowered.api.Sponge;
 import org.spongepowered.api.block.BlockSnapshot;
 import org.spongepowered.api.block.BlockType;
 import org.spongepowered.api.block.BlockTypes;
 import org.spongepowered.api.command.CommandSource;
-import org.spongepowered.api.config.ConfigManager;
+import org.spongepowered.api.config.ConfigDir;
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.event.Listener;
 import org.spongepowered.api.event.block.InteractBlockEvent;
+import org.spongepowered.api.event.filter.cause.Root;
 import org.spongepowered.api.event.game.state.*;
 import org.spongepowered.api.item.inventory.ItemStack;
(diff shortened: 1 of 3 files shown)


worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
```diff
@@ -141,20 +139,16 @@ public void serverStarted(GameStartedServerEvent event) {
     private boolean ignoreLeftClickAir = false;
 
     @Listener
-    public void onPlayerInteract(InteractBlockEvent event) {
+    public void onPlayerInteract(InteractBlockEvent event, @Root Player spongePlayer) {
```
DarkArc: I did not know you could do this, that's awesome!

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
```diff
@@ -59,7 +58,9 @@
 /**
  * The Sponge implementation of WorldEdit.
  */
-@Plugin(id = SpongeWorldEdit.MOD_ID, name = "WorldEdit", version = "%VERSION%")
+@Plugin(id = SpongeWorldEdit.MOD_ID, name = "WorldEdit",
```
DarkArc: So is this necessary with the ID being set in the gradle build script?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
```diff
@@ -83,7 +84,11 @@ public static SpongeWorldEdit inst() {
     }
 
     private SpongePlatform platform;
+
+    @Inject
     private SpongeConfiguration config;
+
+    @Inject @ConfigDir(sharedRoot = false)
     private File workingDir;
```
DarkArc: Will this automatically run mkdirs?

Minecrell: IIRC this won't but the `@DefaultConfig` injection will.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorldEdit.java
Minecrell: Well you still need the `@Plugin` annotation with at least the plugin ID. The URL and description could be also defined in the Gradle file but considering it doesn't change every build we can also put it here.

DarkArc:
Merged in fd0e400bcc29016951eb5591f08da106d0f6c359

PR #359, opened 2016-06-18
Update WE Forge to 1.9.4/1965

kenzierocks:


kenzierocks:
This could use some testing, I tested `//regen` and placing blocks but there may be other things that broke (don't see how due to a lack of major code changes, but just in case).

DarkArc:
On 1.9.4 Sponge I noticed an exception with sign text being set via `//set sign|test`. Can you reproduce that on this branch of forge?

kenzierocks:
Yes, I can. I'm looking into it, probably related to signs using JSON.

DarkArc:
Yeah I tried forcing invalid Json parses into {"text": "my text"} but that didn't seem to work either...

kenzierocks:
Looks like Mojang removed the fallback for when JSON didn't parse in 1.9 or 1.9.4. It used to just take the text as-is, now it will propagate the exception. I'm going to add something to fix this in a few minutes.

DarkArc:
Don't forget it needs to handle old schematics as well.

---- On Sat, 18 Jun 2016 09:24:37 -0700 notifications@github.com wrote ----

Looks like Mojang removed the fallback for when JSON didn't parse in 1.9 or 1.9.4. It used to just take the text as-is, now it will propagate the exception. I'm going to add something to fix this in a few minutes.

kenzierocks:
@DarkArc turns out it wasn't Mojangs fault, it was ours. https://github.com/sk89q/WorldEdit/commit/867e59ba3d28a260ac4b40edf31ad23802337b33#diff-c3a3d1623b4c9c3ef80e84f2ce85ecccR67 put extra quotes in the generated JSON.

kenzierocks:
Sign fixed pushed to master, and this PR was rebased on top of it.

kenzierocks:
Merging before I start working on 1.10.

PR #360, opened 2016-06-19
Add new 1.10 blocks to blocks.json

Minecrell:
Not sure if that helps with something, but it's good to have it updated :)


wizjany:
was this generated or done by hand? i wasn't aware forge or even MCP was out for 1.10.

Minecrell:
I generated this using your class with a few minor changes so it runs on the Vanilla client - basically just logger calls and https://github.com/wizjany/ForgeUtils/blob/master/src/main/java/org/enginehub/util/forge/BlockRegistryDumper.java#L51 changed to `Block.REGISTRY`.

MCP bot isn't updated yet but the SRG mappings are available since yesterday.

wizjany:
merged

PR #362, opened 2016-07-10
Make shadowJar an artifact instead of depending on it

falukerites:


wizjany:
reasoning? our build system kind of depends on this.

falukerites:
The build will still depend on it

wizjany:
that does not really answer the question, so I am going to guess you don't have a good reason.

PR #363, opened 2016-08-16
Allow for plugins/mods to register a clipboard format.

kenzierocks:
This change adds more flexibility to the ClipboardFormat system. It allows mods/plugins to register a format with WorldEdit for saving and loading. This change also allows to adopt Sponge's built-in schematic saver/loader (in a Sponge PR right now) if we want.

My main concern for this PR is that it breaks binary compatibility for `ClipboardFormat` by renaming the implementation to `SupportedClipboardFormat`. I think another solution may need to be considered, like deprecating the current `ClipboardFormat` and relocating them somewhere else.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormats.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormats.java
@@ -1,0 +1,88 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.extent.clipboard.io;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public final class ClipboardFormats {
+
+    private static final Map<String, ClipboardFormat> aliasMap = new HashMap<>();
+    private static final List<ClipboardFormat> registeredFormats = new ArrayList<>();
+
+    public static void registerClipboardFormat(ClipboardFormat format) {
+        checkNotNull(format);
+
+        for (String key : format.getAliases()) {
+            aliasMap.put(key.toLowerCase(Locale.ENGLISH), format);
+        }
+        registeredFormats.add(format);
+    }
+
+    /**
+     * Find the clipboard format named by the given alias.
+     *
+     * @param alias
+     *            the alias
+     * @return the format, otherwise null if none is matched
+     */
+    @Nullable
+    public static ClipboardFormat findByAlias(String alias) {
+        checkNotNull(alias);
+        return aliasMap.get(alias.toLowerCase(Locale.ENGLISH).trim());
+    }
+
+    /**
+     * Detect the format of given a file.
+     *
+     * @param file
+     *            the file
+     * @return the format, otherwise null if one cannot be detected
+     */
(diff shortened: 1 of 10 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormats.java
```diff
@@ -0,0 +1,69 @@
...
+public final class ClipboardFormats {
+
+    private static final Map<String, ClipboardFormat> aliasMap = new HashMap<>();
+    private static final List<ClipboardFormat> registeredFormats = new ArrayList<>();
+
+    public static void registerClipboardFormat(ClipboardFormat format) {
+        checkNotNull(format);
+
+        for (String key : format.getAliases()) {
+            aliasMap.put(key.toLowerCase(Locale.ENGLISH), format);
```
wizjany: this gives later registered formats absolute priority, potentially making earlier registered formats inaccessible if aliases overlap...

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormats.java
```diff
@@ -0,0 +1,69 @@
+package com.sk89q.worldedit.extent.clipboard.io;
```
wizjany: missing header causing checkstyle to fail

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormats.java
kenzierocks: Not inaccessible - see `getAll`. But perhaps we should throw on conflict?

DarkArc: I'd be a fan of an exception on conflict in this case.

---- On Tue, 16 Aug 2016 17:02:16 -0700 notifications@github.com wrote ----

In worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormats.java:

> +import java.util.Locale;
> +import java.util.Map;
> +import javax.annotation.Nullable;
> +
> +import static com.google.common.base.Preconditions.checkNotNull;
> +
> +public final class ClipboardFormats {
> +
> +    private static final Map<String, ClipboardFormat> aliasMap = new HashMap<>();
> +    private static final List<ClipboardFormat> registeredFormats = new ArrayList<>();
> +
> +    public static void registerClipboardFormat(ClipboardFormat format) {
> +        checkNotNull(format);
> +
> +        for (String key : format.getAliases()) {
> +            aliasMap.put(key.toLowerCase(Locale.ENGLISH), format);
Not inaccessible - see getAll. But perhaps we should throw on conflict?

kenzierocks: Fixed, should we still try to register the other aliases from it? I think it's fine if it doesn't get registered at all.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormats.java
kenzierocks: Fixed, although GitHub doesn't seem to think so :)

boy0001:
Could use reflection to add to the ClipboardFormat enum.

kashike:
no

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SupportedClipboardFormat.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Set;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
+
+/**
+ * A collection of supported clipboard formats.
+ */
+public enum SupportedClipboardFormat implements ClipboardFormat {
```
DarkArc: Maybe this should be `BuildInClipboardFormat`?

boy0001:
mmk, but I'd rather not have binary compatibility broken.
```Java
public static ClipboardFormat addFormat(IClipboardFormat instance) {
    ClipboardFormat newEnum = ReflectionUtils.addEnum(ClipboardFormat.class, instance.getName());
    newEnum.format = instance;
    for (String alias : newEnum.getAliases()) {
        aliasMap.put(alias, newEnum);
    }
    return newEnum;
}
```

kenzierocks:
I'm pretty sure that either A) isn't possible or B) would not be very programmer friendly.

boy0001:
https://github.com/boy0001/FastAsyncWorldedit/wiki/Clipboard-API

kenzierocks:
That looks really sketchy to avoid breaking binary compatibility. It would be easier, faster, and safer to create a new interface that can have an implementation based on `ClipboardFormat`, and deprecate the enum.

boy0001:
Yep, really hacky, but given that it's minecraft, I don't take conventions very seriously. It allows existing mods use any added formats without needing to update to use a new interface. 

Anywho, it's just another option, and creating a new interface + keeping the existing ClipboardFormat class is probably the best solution.

me4502:
Implemented in #418. Thanks

PR #364, opened 2016-09-13
Fix region contain method

mickare:
The region contain method was broken.
A lot of subsquent issues are caused by this bug.
E.g. in a selection the entities in the last blocks at the positive-axis border are not selected.

`Max` block gives the impression of an exclusive point; however it is **inclusive**!
A position that is anywhere between of a 1x1x1 region, would return false in the old implementation.

By simply adding _1_ should solve the problem.

Greetings,
mike

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/CuboidRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/CuboidRegion.java
@@ -329,16 +329,16 @@ public class CuboidRegion extends AbstractRegion implements FlatRegion {
 
     @Override
     public boolean contains(Vector position) {
-        double x = position.getX();
-        double y = position.getY();
-        double z = position.getZ();
+        int x = position.getBlockX();
+        int y = position.getBlockY();
+        int z = position.getBlockZ();
 
         Vector min = getMinimumPoint();
         Vector max = getMaximumPoint();
 
-        return x >= min.getBlockX() && x < max.getBlockX() + 1
-                && y >= min.getBlockY() && y < max.getBlockY() + 1
-                && z >= min.getBlockZ() && z < max.getBlockZ() + 1;
+        return x >= min.getBlockX() && x <= max.getBlockX()
+                && y >= min.getBlockY() && y <= max.getBlockY()
+                && z >= min.getBlockZ() && z <= max.getBlockZ();
     }
 
     @Override
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/CuboidRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/CuboidRegion.java
@@ -336,9 +336,9 @@ public class CuboidRegion extends AbstractRegion implements FlatRegion {
         Vector min = getMinimumPoint();
         Vector max = getMaximumPoint();
 
-        return x >= min.getBlockX() && x <= max.getBlockX()
-                && y >= min.getBlockY() && y <= max.getBlockY()
-                && z >= min.getBlockZ() && z <= max.getBlockZ();
+        return x >= min.getBlockX() && x < max.getBlockX() + 1
+                && y >= min.getBlockY() && y < max.getBlockY() + 1
+                && z >= min.getBlockZ() && z < max.getBlockZ() + 1;
     }
 
     @Override


TomyLobo:
Nice spot.
The difference between ``((double) a) <= ((int) b)`` and ``((double) a) < ((int) b) + 1`` is indeed significant.
However, wouldn't it be better to just use ``getBlockX``/``Y``/``Z`` to obtain ``x``/``y``/``z`` and turn them into ``int``s?

mickare:
Yes, casting the position to int is better.

TomyLobo:
Btw, CuboidRegion.contains has been like that since 97529abacae4764e853f150de3bc3ae0a2cd4f6f
And if GitHub can be trusted, that was even before the departure from hmod, i.e. ages ago.

Why does this only surface now?
Do other places work around this issue maybe?

mickare:
> Why does this only surface now?

The problem that entities at the postive border are not copied exists now a long time. In fact i can't remember it did behave in any other way.

> Do other parts of worldedit and worldguard work around this issue maybe?

The thing is that `CuboidRegion` returned by `CuboidSelector` does not have the correct max point.
Each selection point set in `CuboidSelector` is created by the **block integer position** and it must be.

The issue here at hand is that the returned `CuboidSelector` only contains the block positions, not the exact positions.
The exact max-point would be `x + 0.99999...`.

mickare:
Adding 0.9999... to the max point in a virtual `BlockCuboidRegion` is ugly.

But i have another idea how to solve this, by not causing unwanted behavior. One moment plz...

mickare:
The idea is to add a new method `intersectsBlock` to the region interface.

Region.java
```java
     /**
     * Returns true based on whether the region intersects the block at the point.
     * 
     * @param position the position
     * @return true if intersect
     */
    boolean intersectsBlock(Vector position);
```

CuboidRegion.java
```java
    @Override
    public boolean contains(Vector position) {
        double x = position.getX();
        double y = position.getY();
        double z = position.getZ();

        Vector min = getMinimumPoint();
        Vector max = getMaximumPoint();

        return x >= min.getX() && x <= max.getX()
                && y >= min.getY() && y <= max.getY()
                && z >= min.getZ() && z <= max.getZ();
    }

    @Override
    public boolean intersectsBlock(Vector position) {
      int x = position.getBlockX();
      int y = position.getBlockX();
      int z = position.getBlockX();

      Vector min = getMinimumPoint();
      Vector max = getMaximumPoint();

      return x >= min.getBlockX() && x <= max.getBlockX()
              && y >= min.getBlockY() && y <= max.getBlockY()
              && z >= min.getBlockZ() && z <= max.getBlockZ();
    }
```
AbstractRegion.java & any other
```Java
    /**
     * Checks to see if a block intersects this region.
     */
    @Override
    public boolean intersectsBlock(Vector position) {
     return contains(position);
    }
```

mickare:
https://github.com/mickare/WorldEdit/commit/944cb965b0f089464b2f3631de4cbb6d16102e67

TomyLobo:
I meant check all the places in WorldEdit/WorldGuard, not rename the function to make sure it has no negative _or_ positive effect on parts you didn't check.

mickare:
Yeah, I would normally also propose to leave `contains` as it is. I can't see all effects this could have, e.g. there are many plugins which use the world edit api.

Unfortunately in Java7 my alternate fix-idea (https://github.com/mickare/WorldEdit/commit/944cb965b0f089464b2f3631de4cbb6d16102e67) to **add** a additional new method `intersectsBlock` to the interface `Region` is also a no-can-do. All implementations that don't extend `AbstractRegion` would need to implement it. This would force all plugins, which implement their own region, to recompile and add this method. *meeeeh*

With Java8 it would be possible to just add `intersectsBlock` as a default interface method. Java8 FTW! ;)
Ever thought about to switch to Java8?



Switching to German-Mode: Grüße aus Karlsruhe ^^

kenzierocks:
Yes, currently we only have Java 8 required for Sponge. But we're trying to get it required for other implementations as well.

TomyLobo:
Other plugins that use worldedit api (besides @sk89q's other plugins) use them at their own risk, especially if they dont use them according to spec.

DarkArc:
I wouldn't go quite that far @TomyLobo. That said, in this case, any breakages would be in truth bug fixes. That said, if this is being changed, we should also verify the logic on the polygon based regions, as I believe they suffer from the same issue.

Referencing the WorldGuard region methods should reveal solutions to all of these issues. 

That said, a good algorithm for finding the volume of concave polygons -- with this strange +1 point rule -- was never implemented, which may be a concern for both plugins.

TomyLobo:
iirc the polygons are defined by lines going from center to center, so they shouldnt suffer from the same problem.

mickare:
WorldGuard does the same as my initial "`+1`" pull request:
https://github.com/sk89q/WorldGuard/blob/master/worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/regions/ProtectedCuboidRegion.java#L124

me4502:
Thanks

PR #367, opened 2016-11-17
Update Sponge WorldEdit to 1.10 and 1.11

me4502:
This also adds an adapter system, akin to the Spigot version, that allows it to function across multiple versions.


me4502:
I just realized, this shouldn't work should it? It's a MCP named class file being loaded in SF & SV.. Yet it does? Anyone know why this could be?

me4502:
In my opinion it's a lot better than the alternative.

DarkArc:
Well, unlike bukkit, a Sponge update normally requires both API changes, and NMS changes. The adapter only helps with NMS, I feel like it's a false sense of cross version stability.

wizjany:
@DarkArc @me4502 status on this? as noted, WE 6 API still uses int IDs, so the adapter system is probably as good as it gets. are there any other concerns before pulling this?

me4502:
This doesn't currently work - so I wouldn't pull it just yet. I've been super busy, but basically the issue is that because 1.10 and 1.11 are similar, it's loading the 1.11 one on 1.10. It needs a way to determine which is the correct one to load, other than it erroring on class load.

me4502:
Moved to the sponge-adapter branch.

PR #349, opened 2016-05-14
Add spigot 1.9-R2 adapter

Cat7373:
package com.sk89q.worldedit.bukkit.adapter.impl;

    import java.lang.reflect.Field;
    import java.lang.reflect.Method;
    import java.util.ArrayList;
    import java.util.HashMap;
    import java.util.List;
    import java.util.Map;
    import java.util.Set;
    import java.util.logging.Level;
    import java.util.logging.Logger;

    import javax.annotation.Nullable;

    import org.bukkit.Bukkit;
    import org.bukkit.Location;
    import org.bukkit.Material;
    import org.bukkit.block.Biome;
    import org.bukkit.block.Block;
    import org.bukkit.craftbukkit.v1_9_R2.CraftServer;
    import org.bukkit.craftbukkit.v1_9_R2.CraftWorld;
    import org.bukkit.craftbukkit.v1_9_R2.block.CraftBlock;
    import org.bukkit.craftbukkit.v1_9_R2.entity.CraftEntity;
    import org.bukkit.event.entity.CreatureSpawnEvent;

    import com.google.common.base.Preconditions;
    import com.sk89q.jnbt.ByteArrayTag;
    import com.sk89q.jnbt.ByteTag;
    import com.sk89q.jnbt.CompoundTag;
    import com.sk89q.jnbt.DoubleTag;
    import com.sk89q.jnbt.EndTag;
    import com.sk89q.jnbt.FloatTag;
    import com.sk89q.jnbt.IntArrayTag;
    import com.sk89q.jnbt.IntTag;
    import com.sk89q.jnbt.ListTag;
    import com.sk89q.jnbt.LongTag;
    import com.sk89q.jnbt.NBTConstants;
    import com.sk89q.jnbt.ShortTag;
    import com.sk89q.jnbt.StringTag;
    import com.sk89q.jnbt.Tag;
    import com.sk89q.worldedit.blocks.BaseBlock;
    import com.sk89q.worldedit.bukkit.adapter.BukkitImplAdapter;
    import com.sk89q.worldedit.entity.BaseEntity;
    import com.sk89q.worldedit.internal.Constants;

    import net.minecraft.server.v1_9_R2.BiomeBase;
    import net.minecraft.server.v1_9_R2.BlockPosition;
    import net.minecraft.server.v1_9_R2.EntityTypes;
    import net.minecraft.server.v1_9_R2.NBTBase;
    import net.minecraft.server.v1_9_R2.NBTTagByte;
[...]

PR #348, opened 2016-05-12
Add spigot 1.9-R2 adapter

Tim-Schneider:

PR #368, opened 2016-11-17
Add Bukkit Adapter for 1.11

SupremeMortal:
Seems to be functional.
Here's the source: https://gist.github.com/SupremeMortal/da749b403c5376f9f89be9ea8f9cca33


wizjany:
#348 #349 #326

wizjany:
also not sure how I feel about pulling something from someone whose first attempt at fixing a changed signature was
```java
  @Nullable
  private static net.minecraft.server.v1_11_R1.Entity createEntityFromId(String paramString, World paramWorld) {
    throw new Error("Unresolved compilation problem: \n\tThe method createEntityByName(String, World) is undefined for the type EntityTypes\n");
  }
```

PR #369, opened 2016-11-18
Initial update to 1.11 Forge.

kenzierocks:
Tested:
- [x] `//set`, `//undo`, `//redo` -- the basics
- [x] `//regen` -- this broke last time
- [x] Tile entities

Also updates Gradle to 3.2.1, because keeping updated is good.


kenzierocks:
`//regen` seems to be working as before, but I'm going to rewrite it to work better (i.e. not generate trees all random-like)

kenzierocks:
This is now ready for review @wizjany @DarkArc.

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorld.java
```diff
@@ -243,73 +266,41 @@ public boolean regenerate(Region region, EditSession editSession) {
...
+        try {
+            for (BlockVector vec : region) {
+                editSession.setBlock(vec, from.getBlock(vec));
             }
+        } catch (MaxChangedBlocksException e) {
+            throw new RuntimeException(e);
+        } finally {
+            saveFolder.delete();
+            DimensionManager.setWorld(originalWorld.provider.getDimension(), null, server);
+            DimensionManager.setWorld(originalWorld.provider.getDimension(), originalWorld, server);
```
DarkArc: Why does this need nulled first?

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorld.java
```diff
@@ -243,73 +266,41 @@ public boolean regenerate(Region region, EditSession editSession) {
...
         }
 
-        return false;
+        return true;
```
DarkArc: :+1: on rewriting this method

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorld.java
kenzierocks: I'm not sure it needs to be, but it triggers the unload handlers this way.

PR #371, opened 2016-12-02
Fix /center

Griffrez:
We've fixed a bug related to the //center command.

We detected that for the four different world quadrants (+X, +Z; +X, -Z; -X, +Z; -X, -Z), the behavior of this command differed. Some of this behavior's particular cases showed up three times in the issue tracker:

- [Issue 1](http://dev.enginehub.org/youtrack/issue/WORLDEDIT-3450)
- [Issue 2](http://dev.enginehub.org/youtrack/issue/WORLDEDIT-3362)
- [Issue 3](http://dev.enginehub.org/youtrack/issue/WORLDEDIT-2986)

This pull request includes the changes necessary to have the expected behavior on all four quadrants.

Before the change:
before

The cobblestone portion represents the xOz center (the center of the cobblestone portion is the coordinate (0, Y, 0), Y is 85, but Y is irrelevant). The several squares around the center are the same region in the four different quadrants. As you can see, the behavior is not the same. From the upper right corner, going counterclockwise, the quadrants are (+X, +Z), (+X, -Z), (-X, -Z), (-X, +Z). The wrong behavior comes from reaching negative coordinates and the previous algorithm didn't correctly select the center coordinates for all possible use cases. The previous code already worked with odd-length sided shapes, but were incorrect when there were sides with even length.

After the change:
after

With the change, all quadrants work well.


worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -961,10 +961,15 @@ public int center(Region region, Pattern pattern) throws MaxChangedBlocksExcepti
         checkNotNull(pattern);
 
         Vector center = region.getCenter();
+        Vector vec1 = new Vector((int) center.getX(), (int) center.getY(), (int) center.getZ());
+        Vector vec2 = center.toBlockVector();
+        boolean condX = (region.getWidth() % 2) == 0;
+        boolean condZ = (region.getLength() % 2) == 0;
+        Vector vec3 = new Vector(vec2.getX() + (condX ? 0.5*Math.signum(vec2.getX()) : 0), vec2.getY(), vec2.getZ() + (condZ ? 0.5*Math.signum(vec2.getZ()) : 0));
```
kenzierocks: Could you pull this logic out to a method, and maybe explain it?

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -961,10 +961,15 @@ public int center(Region region, Pattern pattern) throws MaxChangedBlocksExcepti
...
         Region centerRegion = new CuboidRegion(
                 getWorld(), // Causes clamping of Y range
-                new Vector((int) center.getX(), (int) center.getY(), (int) center.getZ()),
-                center.toBlockVector());
+                vec1,
+                vec3);
```
kenzierocks: Merge this line with the one above.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
wizjany: also if this is getting fixed again, can we do it correctly this time instead of relying on coordinate roundoff. 
why are we relying on a +/- 0.5 to be rounded to the correct side when we could just supply the right value to begin with.
this line is entirely unreadable and obfuscates the fact that we're relying on [these lines](https://github.com/sk89q/WorldEdit/blob/03f8843c2021425e5279aa82f9be6177d1802eed/worldedit-core/src/main/java/com/sk89q/worldedit/regions/CuboidRegion.java#L349-L351) to do the inference for us - this math isn't actually determining the center region on its own, it's just determining the correct "half block" we should be on so it later gets rounded to the right side.

up201406036: I'm another contributor from the fork. 
Basically, the problem was on the third argument of the CuboidRegion constructor used to define centerRegion, where the vector's coordinates were not being properly rounded. In particular, if the x or z coordinates were -a.5 (a being a positive integer) (i.e. when the width or length of the region were even), the number was rounded to -a and not to -(a+1).
We added a new commit to the pull request where we simplified this bug correction.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -961,10 +961,11 @@ public int center(Region region, Pattern pattern) throws MaxChangedBlocksExcepti
         checkNotNull(pattern);
 
         Vector center = region.getCenter();
+        Vector vec2 = center.toBlockVector();
         Region centerRegion = new CuboidRegion(
                 getWorld(), // Causes clamping of Y range
                 new Vector((int) center.getX(), (int) center.getY(), (int) center.getZ()),
-                center.toBlockVector());
+                new Vector(Math.signum(vec2.getX())*Math.round(Math.abs(vec2.getX())), vec2.getY(), Math.signum(vec2.getZ())*Math.round(Math.abs(vec2.getZ()))));
```
kenzierocks: I'd still like this to be pulled out to its own method to reduce the noise on this line.

Griffrez: We've made a new commit complying with this request.

worldedit-core/src/main/java/com/sk89q/worldedit/math/MathUtils.java
```diff
@@ -105,4 +105,20 @@ public static double dSin(double degrees) {
...
+     * ("Round up")
+     *
+     * In other words, {@code Math.round(-3.5)}, as an example, rounds -3.5 to -3
+     * instead of the "typical" -4. The following function has the "typical"
+     * behavior for both positive and negative values.
+     *
+     * @param value the value
+     * @return the rounded value
+     */
+    public static double round(double value) {
```
kenzierocks: I wouldn't use the word typical, as there is [no typical way to round a fractional part of exactly `1/2`](https://math.stackexchange.com/questions/3448/rules-for-rounding-positive-and-negative-numbers). Perhaps we can just say `instead of to -4` and `The following function uses "to nearest even number" rounding`?

Griffrez: It doesn't round to the nearest even number. It rounds to the nearest integer. The difference is the behavior when the fractional part is exactly 0.5 or 1/2. But we'll change the description to better reflect that. We just meant "typical" as in the way it's taught in basic arithmetic, despite the ambiguity in computing contexts.

wizjany: except those are all different things and none are standard, especially when it comes to the applications in negatives here.
https://docs.oracle.com/javase/7/docs/api/java/math/RoundingMode.html

TomyLobo:
I'm just gonna say the test case in your screenshot is neither symmetric nor sufficient, even though it looks that way.
What I mean is that in the negative, your selection goes from (as far as i can see) block coordinates -2 to -5, while in the positive range, your selection goes from 1 to 4.

Have you made sure that this isn't the issue instead of negative/positive?

Griffrez:
@TomyLobo You are correct about the points used. To clear this up I'll place some screenshots with a new test condition that should offer a clear look into how the problem is the negative coordinates and not odd/even disparities.




Here, the points selected, for each region from the top right in a counterclockwise motion, are:

- (1.0, 1.0);(4.0, 4.0)
- (1.0, -2.0);(4.0, 5.0)
- (-1.0, -1.0);(-4.0, -4.0)
- (-2.0, 1.0);(-5.0, 4.0)

Also, for maximum reproduction, the y coordinate value is 85, but that is irrelevant, if the premise is indeed correct.

TomyLobo:
ok, what should happen is that it rounds down all the time, for each coordinate.
rounding a coordinate down finds the block coordinate for the block that encompasses it, regardless of sign.

what the old code does is region.getCenter(), which, for cuboid regions, is (for each component):
````java
(((double)min)+((double)max))/((int)2)
````

Then it does .toBlockVector on that, which basically is a no-op. it just changes the type, so you might as well skip it. The only thing that type does differently is hashCode and equals.

Now what's different is that ``(int) myDouble`` behaves differently depending on quadrant. What happens if you change that to floor?

TomyLobo:
@wizjany closed for inactivity or because this was solved elsewhere?

kashike:
https://github.com/sk89q/WorldEdit/commit/b2fb73582f8a11db03dfbce6da3d4383c72f1894

PR #372, opened 2016-12-06
Use an Adapter system for Sponge - Supports 1.10.2 and 1.11 at this time.

me4502:
This is the updated version of my adapter system for Sponge. It actually works now.


me4502:
@wizjany This is now ready. It works in a dev workspace.

PR #373, opened 2016-12-12
Added getWorldSelection.

hexosse:
This is an alias for getSelection.
It enables CraftScripts to get a world selection as it is not possible to use getSelection which have two default implementations.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
@@ -421,6 +421,27 @@ public class LocalSession {
     }
 
     /**
+     * Get the world selection.
+     *
+     * @return the current selection
+     */
+    public Region getWorldSelection() throws IncompleteRegionException {
+        return getSelection(getSelectionWorld());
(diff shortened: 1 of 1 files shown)


kenzierocks:
I don't understand what the problem is here. Rhino chooses which method to call based on the argument types, so calling either of them shouldn't be a problem.

hexosse:
Rhino cannot choose because there is two overloaded functions getSelection(LocalWorld world)  and getSelection(World world).

when i try to do :
```
var world = context.getSession().getSelectionWorld();
var region = context.getSession().getSelection(world);
```

There is an error as rhino cannot resolve the good function.

kenzierocks:
What error are you getting? According to the [documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Rhino/Scripting_Java#Calling_Overloaded_Methods), that should work.

hexosse:
Your right, after some more testing it seems that the problem does not come from WorldEdit but from FastAsyncWorldEdit.

Sorry for wasting your time. I'll continue to investigate on FastAsyncWorldEdit.

PR #374, opened 2016-12-17
Add //hemisphere

Griffrez:
We've added a feature: the possibility to generate hemispheres using the generation command //hemisphere.

Its usage is similar to //sphere's: `//hemisphere [-h] <block> <radius>[,<radius>,<radius>]`.
The raised argument is ignored since we felt it didn't make sense to be used with hemispheres.

We also added a -s flag to the //sphere and //hsphere commands, similar to how -h flag matches the //hsphere command.

We reused methods associated to the //sphere command to avoid duplicated code.

Using the 3 commands:
//hemisphere -h snowblock 8,4,2
//hemisphere -h snowblock 4
//hemisphere -h snowblock 4,5,6
generates the following, from left to right:

EDIT: "semisphere" -> "hemisphere"

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -24,6 +24,7 @@ import com.sk89q.worldedit.blocks.BlockID;
 import com.sk89q.worldedit.blocks.BlockType;
 import com.sk89q.worldedit.entity.BaseEntity;
 import com.sk89q.worldedit.entity.Entity;
+import com.sk89q.worldedit.entity.Player;
 import com.sk89q.worldedit.event.extent.EditSessionEvent;
 import com.sk89q.worldedit.extent.ChangeSetExtent;
 import com.sk89q.worldedit.extent.Extent;
@@ -1414,13 +1415,11 @@ public class EditSession implements Extent {
     * @param block The block pattern to use
     * @param radius The sphere's radius
     * @param filled If false, only a shell will be generated.
-    * @param hemi If true, only the top half will be generated.
-    * @param upsideDown If true and hemi is true, only the bottom half will be generated.
     * @return number of blocks changed
     * @throws MaxChangedBlocksException thrown if too many blocks are changed
     */
-    public int makeSphere(Vector pos, Pattern block, double radius, boolean filled, boolean hemi, boolean upsideDown) throws MaxChangedBlocksException {
-        return makeSphere(pos, block, radius, radius, radius, filled, hemi, upsideDown);
+    public int makeSphere(Vector pos, Pattern block, double radius, boolean filled) throws MaxChangedBlocksException {
+        return makeSphere(pos, block, radius, radius, radius, filled, false, false, null);
     }
 
     /**
@@ -1437,7 +1436,7 @@ public class EditSession implements Extent {
      * @return number of blocks changed
      * @throws MaxChangedBlocksException thrown if too many blocks are changed
      */
-    public int makeSphere(Vector pos, Pattern block, double radiusX, double radiusY, double radiusZ, boolean filled, boolean hemi, boolean upsideDown) throws MaxChangedBlocksException {
+    public int makeSphere(Vector pos, Pattern block, double radiusX, double radiusY, double radiusZ, boolean filled, boolean hemi, boolean upsideDown, @Nullable Vector directionVector) throws MaxChangedBlocksException {
         int affected = 0;
 
         radiusX += 0.5;
@@ -1454,6 +1453,10 @@ public class EditSession implements Extent {
 
         final boolean genUp = (!hemi || (hemi && !upsideDown));
         final boolean genDown = (!hemi || (hemi && upsideDown));
+
+        if (directionVector != null && upsideDown) {
+            directionVector = directionVector.multiply(-1);
+        }
 
         double nextXn = 0;
         forX: for (int x = 0; x <= ceilRadiusX; ++x) {
@@ -1485,32 +1488,44 @@ public class EditSession implements Extent {
                         }
                     }
 
-                    if(genUp) {
-                        if (setBlock(pos.add(x, y, z), block)) {
-                            ++affected;
-                        }
-                        if (setBlock(pos.add(-x, y, z), block)) {
-                            ++affected;
-                        }
-                        if (setBlock(pos.add(x, y, -z), block)) {
-                            ++affected;
-                        }
-                        if (setBlock(pos.add(-x, y, -z), block)) {
-                            ++affected;
+                    if (directionVector != null) {
+                        Vector vectors[] = {pos.add(x, y, z), pos.add(-x, y, z), pos.add(x, y, -z), pos.add(-x, y, -z), pos.add(x, -y, z), pos.add(-x, -y, z), pos.add(x, -y, -z), pos.add(-x, -y, -z)};
+
+                        for (Vector vector : vectors) {
+                            Vector vectorFromCenter = vector.subtract(pos);
+                            if((vectorFromCenter.dot(directionVector) >= 0) && (setBlock(vector, block))) {
+                                       ++affected;
+                            }
                         }
                     }
-                    if(genDown) {
-                        if (setBlock(pos.add(x, -y, z), block)) {
-                            ++affected;
+                    else {
+                        if (genUp) {
+                            if (setBlock(pos.add(x, y, z), block)) {
+                                ++affected;
+                            }
+                            if (setBlock(pos.add(-x, y, z), block)) {
+                                ++affected;
+                            }
+                            if (setBlock(pos.add(x, y, -z), block)) {
+                                ++affected;
+                            }
+                            if (setBlock(pos.add(-x, y, -z), block)) {
+                                ++affected;
+                            }
                         }
-                        if (setBlock(pos.add(-x, -y, z), block)) {
-                            ++affected;
-                        }
-                        if (setBlock(pos.add(x, -y, -z), block)) {
-                            ++affected;
-                        }
-                        if (setBlock(pos.add(-x, -y, -z), block)) {
-                            ++affected;
+                        if (genDown) {
+                            if (setBlock(pos.add(x, -y, z), block)) {
+                                ++affected;
+                            }
+                            if (setBlock(pos.add(-x, -y, z), block)) {
+                                ++affected;
+                            }
+                            if (setBlock(pos.add(x, -y, -z), block)) {
+                                ++affected;
+                            }
+                            if (setBlock(pos.add(-x, -y, -z), block)) {
+                                ++affected;
+                            }
                         }
                     }
                 }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -1432,7 +1432,7 @@ public class EditSession implements Extent {
      * @param filled If false, only a shell will be generated.
      * @param hemi If true, only the top half will be generated.
      * @param upsideDown If true and hemi is true, only the bottom half will be generated.
-     * @param directionVector If not null and hemi is true, the hemisphere's orientation will depend on the player's viewing angle.
+     * @param directionVector If not null and hemi is true, the hemisphere's direction will match to the vector.
      * @return number of blocks changed
      * @throws MaxChangedBlocksException thrown if too many blocks are changed
      */
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -24,7 +24,6 @@ import com.sk89q.worldedit.blocks.BlockID;
 import com.sk89q.worldedit.blocks.BlockType;
 import com.sk89q.worldedit.entity.BaseEntity;
 import com.sk89q.worldedit.entity.Entity;
-import com.sk89q.worldedit.entity.Player;
 import com.sk89q.worldedit.event.extent.EditSessionEvent;
 import com.sk89q.worldedit.extent.ChangeSetExtent;
 import com.sk89q.worldedit.extent.Extent;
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/GenerationCommands.java
```diff
@@ -122,8 +122,28 @@ public void cyl(Player player, LocalSession session, EditSession editSession, Pa
     }
 
     @Command(
+       aliases = { "/semisphere" },
+       usage = "<block> <radius>[,<radius>,<radius>] [raised?]",
```
kenzierocks: If you're ignoring raised you should remove it from the usage.

Griffrez:
The new commit should have captured all the "semi" misspells. It also adds the choice to invert the hemisphere through the use of a -i flag. This flag will not have any effect unless it's an hemisphere command or another sphere-related command with the -s flag.



From the topdown, left to right:
//semisphere -i snowblock 8,4,2
//semisphere -ih snowblock 4
//semisphere -i snowblock 4,5,6

//semisphere -h snowblock 8,4,2
//semisphere -h snowblock 4
//semisphere -h snowblock 4,5,6

(You can't really see the hollowness of the second set of hemispheres, but they're the same from the original post.)

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1478,29 +1482,33 @@ public int makeSphere(Vector pos, Pattern block, double radiusX, double radiusY,
...
-                    }
-                    if (setBlock(pos.add(-x, -y, z), block)) {
-                        ++affected;
-                    }
-                    if (setBlock(pos.add(x, -y, -z), block)) {
-                        ++affected;
-                    }
-                    if (setBlock(pos.add(-x, y, -z), block)) {
-                        ++affected;
+                    if(!hemi || (hemi && !upsideDown)) {
```
Griffrez: This can be simplified to (!hemi || !upsideDown) but maybe this way it's more readable. (And possibly Java optimizes it out anyway.)

kenzierocks: For readability, I would just put two booleans at the top, `makeTop` and `makeBottom`, and set them to whatever expression. Then use those in the `if` statements.

TomyLobo:
Well hemispheres can already be done by using a mask.
But this adds convenience of course.

How about the other 4 types of hemispheres, though?
You could use the north south east west up down parameter idiom with fallback to player direction if not specified.
You could even use a -p (for precise) flag that instead limits block placement at a plane perpendicular to the player's view and containing the sphere's center
I'm thinking radio dishes here :)

TomyLobo:
Do you think a separate "worldedit.generation.hemisphere" permission is needed?
It's not fundamentally more or less destructive than a regular sphere.

kashike:
I assume the separate permission is so that permission can be granted to use hemispheres and not full spheres.

TomyLobo:
Captain Obvious!

kenzierocks:
@kashike that doesn't really make any sense though, if you can make half a sphere you can make a whole sphere 😛

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1478,29 +1485,33 @@ public int makeSphere(Vector pos, Pattern block, double radiusX, double radiusY,
...
-                    }
-                    if (setBlock(pos.add(-x, -y, z), block)) {
-                        ++affected;
-                    }
-                    if (setBlock(pos.add(x, -y, -z), block)) {
-                        ++affected;
-                    }
-                    if (setBlock(pos.add(-x, y, -z), block)) {
-                        ++affected;
+                    if(genUp) {
```
kenzierocks: Could you add a space before the parenthesis?

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1478,29 +1485,33 @@ public int makeSphere(Vector pos, Pattern block, double radiusX, double radiusY,
...
+                        if (setBlock(pos.add(x, y, -z), block)) {
+                            ++affected;
+                        }
+                        if (setBlock(pos.add(-x, y, -z), block)) {
+                            ++affected;
+                        }
                     }
-                    if (setBlock(pos.add(-x, -y, -z), block)) {
-                        ++affected;
+                    if(genDown) {
```
kenzierocks: Same here (add a space).

TomyLobo:
If you want to go fully general mode, here's the formula to check which half a block is in, given a directional vector and a center:
````java
(blockPosition - center).dot(direction) <=> 0
````
Whether to use <, >, <= or >= depends on a number of factors. Just experiment to see what works best :)
The difference between </> and <=/>= should be the central ring (for axis-aligned directional vectors and block-aligned centers). I'd suggest always creating that central ring.

**Comment by Griffrez** [comment]
We've added the capability to use the player's viewing vector in order to change the hemisphere's orientation.



As examples we used the commands, from left to right:
//hemisphere stone 4
//hemisphere stone 4 -h
//hemisphere stone 4 -i
//hemisphere stone 4 -p
//hemisphere stone 4 -pi
//hemisphere stone 4 -ph

From the opposite side:

File: worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java [hunk 5]
```diff
@@ -1430,10 +1430,12 @@ public int makeSphere(Vector pos, Pattern block, double radius, boolean filled)
      * @param radiusY The sphere/ellipsoid's largest up/down extent
      * @param radiusZ The sphere/ellipsoid's largest east/west extent
      * @param filled If false, only a shell will be generated.
+     * @param hemi If true, only the top half will be generated.
+     * @param upsideDown If true and hemi is true, only the bottom half will be generated.
      * @return number of blocks changed
      * @throws MaxChangedBlocksException thrown if too many blocks are changed
      */
-    public int makeSphere(Vector pos, Pattern block, double radiusX, double radiusY, double radiusZ, boolean filled) throws MaxChangedBlocksException {
+    public int makeSphere(Vector pos, Pattern block, double radiusX, double radiusY, double radiusZ, boolean filled, boolean hemi, boolean upsideDown, @Nullable Vector directionVector) throws MaxChangedBlocksException {
```
kenzierocks: If you're going to add the parameter you should probably add it to the javadocs like the others :)

worldedit-core/src/main/java/com/sk89q/worldedit/command/GenerationCommands.java
```diff
@@ -122,67 +122,92 @@ public void cyl(Player player, LocalSession session, EditSession editSession, Pa
...
-        usage = "<block> <radius>[,<radius>,<radius>] [raised?]",
-        desc = "Generates a hollow sphere.",
-        help =
-            "Generates a hollow sphere.\n" +
-            "By specifying 3 radii, separated by commas,\n" +
-            "you can generate an ellipsoid. The order of the ellipsoid radii\n" +
-            "is north/south, up/down, east/west.",
-        min = 2,
-        max = 3
+            aliases = { "/hemisphere" },
```
kenzierocks: It looks like your formatted these annotations incorrectly. I'm not sure if we have an official formatter profile for IntelliJ or Eclipse though.

worldedit-core/src/main/java/com/sk89q/worldedit/command/GenerationCommands.java
```diff
@@ -122,67 +122,92 @@ public void cyl(Player player, LocalSession session, EditSession editSession, Pa
...
 
         Vector pos = session.getPlacementPosition(player);
-        if (raised) {
+        if (raised && !hemi) {
             pos = pos.add(0, radiusY, 0);
         }
 
-        int affected = editSession.makeSphere(pos, Patterns.wrap(pattern), radiusX, radiusY, radiusZ, !hollow);
+        Vector directionVector = null;
+        if (hemi && precisionMode) {
```
kenzierocks: Perhaps using the precision flag without the hemisphere flag should be an error? What do you think @TomyLobo and @wizjany?

Griffrez: What kind of error? A error value that then is interpreted by the sphere command method?

kenzierocks: Just a message to the user, saying that it won't have any effect, and then not performing the action. Not sure if we really need it.

worldedit-core/src/main/java/com/sk89q/worldedit/command/GenerationCommands.java
Griffrez: I suppose the complaint is that max should be 2 instead of 3?

kenzierocks: No, the indentation is incorrect, for all the annotations and some of the `case` statements. I know it's a minor thing but personally I'd prefer the whole project to stick to one style as much as possible 👍 

If you need to see what's wrong, take a look at the diff for the commit and notice the whitespace changes.

Griffrez: Yes, I see it now. Must have been a slight mistake when copy-pasting to my IDE (IntelliJ IDEA). But the max is also incorrect, for the hemisphere.

Griffrez: /hpyramid seems to suffer from this issue too, by the way.

worldedit-core/src/main/java/com/sk89q/worldedit/command/GenerationCommands.java
Griffrez: Okay. I'll do it too for the upsideDown argument. Should I use player.printError?

Griffrez: Doing, for example:
//sphere stone 4 -i
//sphere stone 4 -p
gives these two error messages:


Is this what you're looking for? (Nothing is generated)

Griffrez: One note is that if I do //sphere stone 4 -ip, it will only print the message related to the -p flag since it's tested for first. I can make it so it prints both, but it will clutter the method a little bit.

Griffrez: Nevermind, found a good way to do it without much clutter.

worldedit-core/src/main/java/com/sk89q/worldedit/command/GenerationCommands.java
```diff
@@ -154,7 +174,24 @@ public void hsphere(Player player, LocalSession session, EditSession editSession
     )
     @CommandPermissions("worldedit.generation.sphere")
     @Logging(PLACEMENT)
-    public void sphere(Player player, LocalSession session, EditSession editSession, Pattern pattern, String radiusString, @Optional("false") boolean raised, @Switch('h') boolean hollow) throws WorldEditException {
+    public void sphere(Player player, LocalSession session, EditSession editSession, Pattern pattern, String radiusString, @Optional("false") boolean raised, @Switch('h') boolean hollow, @Switch('s') boolean hemi, @Switch('i') boolean upsideDown, @Switch('p') boolean precisionMode) throws WorldEditException {
+        boolean error = false;
+        if(!hemi) {
+            if(precisionMode) {
+                player.printError("-p flag doesn't have effect when it's not an hemisphere.");
```
kenzierocks: `doesn't have an effect`, not `doesn't have effect`

`not generating a hemisphere`, not `it's not an hemisphere`

worldedit-core/src/main/java/com/sk89q/worldedit/command/GenerationCommands.java
```diff
@@ -154,7 +174,24 @@ public void hsphere(Player player, LocalSession session, EditSession editSession
     )
     @CommandPermissions("worldedit.generation.sphere")
     @Logging(PLACEMENT)
-    public void sphere(Player player, LocalSession session, EditSession editSession, Pattern pattern, String radiusString, @Optional("false") boolean raised, @Switch('h') boolean hollow) throws WorldEditException {
+    public void sphere(Player player, LocalSession session, EditSession editSession, Pattern pattern, String radiusString, @Optional("false") boolean raised, @Switch('h') boolean hollow, @Switch('s') boolean hemi, @Switch('i') boolean upsideDown, @Switch('p') boolean precisionMode) throws WorldEditException {
+        boolean error = false;
+        if(!hemi) {
```
kenzierocks: Spaces before parenthesis :)

sk89q:
Looks pretty cool.

I know you can do some of this stuff with the right combination of WE functions but it's pretty painful.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1478,29 +1488,45 @@ public int makeSphere(Vector pos, Pattern block, double radiusX, double radiusY,
...
-                    if (setBlock(pos.add(-x, -y, z), block)) {
-                        ++affected;
-                    }
-                    if (setBlock(pos.add(x, -y, -z), block)) {
-                        ++affected;
-                    }
-                    if (setBlock(pos.add(-x, y, -z), block)) {
-                        ++affected;
+                    if (directionVector != null) {
+                        Vector vectors[] = {pos.add(x, y, z), pos.add(-x, y, z), pos.add(x, y, -z), pos.add(-x, y, -z), pos.add(x, -y, z), pos.add(-x, -y, z), pos.add(x, -y, -z), pos.add(-x, -y, -z)};
```
TomyLobo: Storing everything in a temporary array becomes superfluous if you add a function that does the filtering.
@sk89q do we have a way to add temporary filters to whatever is behind setBlock already?

TomyLobo: I didn't really notice this, but why is there an "if (directionVector != null)" and still genUp and genDown?
You can implement the later using directionVector as well.
It'll be less code, too.
Btw, having everything in an array might not be that bad if that allows you to put all the logic in one piece of code instead of 6.

TheIntelloBox:
When will this command be added to WorldEdit ?

TheIntelloBox:
@me4502

octylFractal:
Closing this as it hasn't been updated in a while, linking it to #726 since that's basically the request for this. If someone wants to update this PR, feel free to make a new PR with your changes.

PR #375, opened 2016-12-24
hardcode direction for some blocks

boy0001:
The `half`, `rotation` and `axis` information is converted to direction. This includes blocks like signs, banners, logs, slabs, stairs, pillars etc.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledBlockData.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledBlockData.java
@@ -83,6 +83,23 @@ public class BundledBlockData {
         String data = Resources.toString(url, Charset.defaultCharset());
         List<BlockEntry> entries = gson.fromJson(data, new TypeToken<List<BlockEntry>>() {}.getType());
 
+        Vector[] range = new Vector[]{new Vector(0, 0, -1),
+                new Vector(0.5, 0, -1),
+                new Vector(1, 0, -1),
+                new Vector(1, 0, -0.5),
+                new Vector(1, 0, 0),
+                new Vector(1, 0, 0.5),
+                new Vector(1, 0, 1),
+                new Vector(0.5, 0, 1),
+                new Vector(0, 0, 1),
+                new Vector(-0.5, 0, 1),
+                new Vector(-1, 0, 1),
+                new Vector(-1, 0, 0.5),
+                new Vector(-1, 0, 0),
+                new Vector(-1, 0, -0.5),
+                new Vector(-1, 0, -1),
+                new Vector(-0.5, 0, -1)};
+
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledBlockData.java
```diff
@@ -87,6 +87,64 @@ private void loadFromResource() throws IOException {
...
+                    top.setDirection(new Vector(0, 1, 0));
+                }
+                if (bot != null && bot.getDirection() == null) {
+                    bot.setDirection(new Vector(0, -1, 0));
+                }
+                return;
+            }
+            SimpleState dir = entry.states.get("rotation");
+            if (dir != null && dir.valueMap() != null) {
+                Vector[] range = new Vector[]{new Vector(0, 0, -1),
```
wizjany: you can probably move this into a more permanent object instead of re-initializing it for each state. since vectors are immutable there's no danger of overwriting them.

on the other hand, i'm not sure how many blocks this will even apply to.

boy0001: 2 blocks :)

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledBlockData.java
```diff
@@ -83,10 +83,69 @@ private void loadFromResource() throws IOException {
...
+                return;
+            }
+            SimpleState axis = entry.states.get("axis");
+            if (axis != null && axis.valueMap() != null) { // Fix rotation for logs and such with axis information
+                SimpleStateValue x = axis.valueMap().get("x");
+                SimpleStateValue y = axis.valueMap().get("y");
+                SimpleStateValue z = axis.valueMap().get("z");
+                if (x != null) {
+                    x.setDirection(new Vector(1, 0, 0));
+                    axis.valueMap().put("-x", new SimpleStateValue(x).setDirection(new Vector(-1, 0, 0)));
```
wizjany:
```
Caused by: java.lang.UnsupportedOperationException
	at java.util.Collections$UnmodifiableMap.put(Collections.java:1457) ~[?:1.8.0_91]
	at com.sk89q.worldedit.world.registry.BundledBlockData.loadFromResource(BundledBlockData.java:137) ~[main/:?]
```

Did you test this code or just assume it would work?

same for below.

boy0001: @wizjany 
I copied it from FAWE and assumed it would still work. I've fixed it now.

wizjany:
Fixed in https://github.com/wizjany/ForgeUtils/commit/4422b4e4da6a3a64251d3069c7c48e036e216897

PR #377, opened 2016-12-24
Small changes (haha)

boy0001:
Expect anywhere between 10x and several thousand times less memory consumed by the change set.
It's also faster to write 13 bytes to a ByteArrayOutputStream than it is to use two tuple array lists.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -54,7 +54,7 @@ import com.sk89q.worldedit.function.util.RegionOffset;
 import com.sk89q.worldedit.function.visitor.*;
 import com.sk89q.worldedit.history.UndoContext;
 import com.sk89q.worldedit.history.change.BlockChange;
-import com.sk89q.worldedit.history.changeset.BlockOptimizedHistory;
+import com.sk89q.worldedit.history.changeset.ActualBlockOptimizedHistory;
 import com.sk89q.worldedit.history.changeset.ChangeSet;
 import com.sk89q.worldedit.internal.expression.Expression;
 import com.sk89q.worldedit.internal.expression.ExpressionException;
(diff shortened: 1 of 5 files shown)


kenzierocks:
As far as I know, WorldEdit is normally CPU bound for large operations, not memory bound. Is this faster or slower than the original?

boy0001:
The time an edit takes is not primarily memory dependent, and thus not memory bound. WorldEdit either completes an operation or crashes when it runs out of memory (which happens a lot).

Also, it's both faster and uses significantly less memory. Though the CPU bottleneck is setting blocks, recording changes are pretty insignificant (CPU wise).

wizjany:
>WorldEdit either completes an operation or crashes when it runs out of memory (which happens a lot).

or hits the server's timeout...
memory crashes are actually really rare.
that said it's mostly in the setting blocks into the world part, not the saving edit history, so i doubt this would bottleneck either way.

boy0001:
With memory crashes; each player can have 15 items in history, and Block(un)OptimizedHistory uses about 0.16 GB/1M blocks. As context, I can set a 45M cuboid in the 60s timeout (with //fast). OOMs *weren't* rare on multiplayer servers. Most large servers are now using modified WorldEdit (AWE/FAWE which put changes on disk) as this plugin is not feasible.

If you are fine with me restructuring the plugin, which I'm not sure you are, it could be made at least 10x faster, and operations properly batched (or partially async). Even without using NMS (to be fair there's already the adapter), you could probably see a 5 fold improvement for some operations just by optimizing overhead.

kenzierocks:
I think we're okay with some restructuring as long as it doesn't break the API. If you're going to break API, it needs to try and target the `seven` branch instead for WorldEdit 7.0.0.

As an aside, I don't really approve of FAWE because of the hacks it does to get the speeds it presents. It also frequently breaks the core of WorldEdit itself, which is unacceptable for an add-on.

sk89q:
If you want to restructure, I recommend discussing a plan first so we don't end up with a huge patch that won't work without major changes. 

As for the change itself, memory is indeed an issue sometimes and I have really old local branches experimenting with this and disk based history, so it was definitely going to end up on WE eventually if I had the time.

DarkArc:
I started working on a lot of these issues in the seven-new branch. I've just been burnt out of Minecraft development lately. In that branch, is as I recall is an API level of batched editing. Memory issues were also largely addressed by making BlockStates immutable (having 1 block state for all identical occurrences of stone).

I also began working on removing the hard dependency on integer IDs, without the performance penalty of switching over to full strings, and O(n) equality checks. The idea was, WorldEdit maintains its own integer to "native" block mapping, and things translate between their "native" types and the WE integer corresponding to that type.

PR #378, opened 2017-01-04
Fix rotation of lever

Postremus:
This fixes rotation (//rotate 90/180/270) of levers on walls.


wizjany:
#375 
blocks.json should never be hand-modified since it's machine generated. changes will get lost on the next update.

Postremus:
Okay, thanks for the heads up.

PR #382, opened 2017-02-23
[Forge] Update to 1.11.2

kenzierocks:
This might not be necessary, but I think it's a good idea to keep up with the Forge releases anyways.


worldedit-forge/build.gradle
```diff
@@ -26,11 +26,11 @@ repositories {
...
 }
 
-ext.forgeVersion = "13.19.1.2189"
+ext.forgeVersion = "13.20.0.2237"
 
 minecraft {
-    version = "1.11-${project.forgeVersion}"
-    mappings = "snapshot_20161209"
+    version = "1.11.2-${project.forgeVersion}"
+    mappings = "snapshot_20170221"
```
liach: You can use `snapshot_20170223`

kenzierocks:
If no one's reported a problem, I don't think so. I have no clue if the 1.11 version works on 1.11.2 since I've been out of the loop for a little bit.

wizjany:
Tell me about it. Fine with me anyway. Haven't heard of any issues.

PR #384, opened 2017-03-23
Add "worldedit.showchatcoords" permission to show coordinates in chat.

RickyB98:
This is useful for people who stream on the Internet who don't want WE coordinates to show up in chat. They can revoke themselves the above permission.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/Polygonal2DRegionSelector.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/Polygonal2DRegionSelector.java
@@ -167,7 +167,7 @@ public class Polygonal2DRegionSelector extends com.sk89q.worldedit.regions.Polyg
 
     @Override
     public void explainPrimarySelection(Actor player, LocalSession session, Vector pos) {
-        if (player.hasPermission("worldedit.hidechatcoords")) {
+        if (player.hasPermission("worldedit.showchatcoords")) {
             player.print("Starting a new polygon at " + pos + ".");
         } else {
             player.print("Starting a new polygon.");
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/Polygonal2DRegionSelector.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/Polygonal2DRegionSelector.java
@@ -167,7 +167,11 @@ public class Polygonal2DRegionSelector extends com.sk89q.worldedit.regions.Polyg
 
     @Override
     public void explainPrimarySelection(Actor player, LocalSession session, Vector pos) {
-        player.print("Starting a new polygon at " + pos + ".");
+        if (player.hasPermission("worldedit.hidechatcoords")) {
+            player.print("Starting a new polygon at " + pos + ".");
+        } else {
+            player.print("Starting a new polygon.");
+        }
 
         session.dispatchCUIEvent(player, new SelectionShapeEvent(getTypeID()));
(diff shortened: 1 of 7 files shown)


kenzierocks:
Why not just use a client side chat filter?

RickyB98:
Chat filtering client mods might not catch all the messages. The filtering pattern might be difficult to find and messages might change over time, making even the working patterns no longer valid. New messages might be added in the future, too, so new patterns should be found to cover those. And after all, people might not want to use modded clients at all for various reasons. Solving the problem server-side is probably the easiest and most maintainable way for the goal. It also doesn't affect users who haven't "opted-out" for it on purpose, which makes sure nothing has to be reconfigured because of the update.

wizjany:
but why?

worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/Polygonal2DRegionSelector.java
```diff
@@ -167,7 +167,11 @@ public boolean selectSecondary(Vector position, SelectorLimits limits) {
 
     @Override
     public void explainPrimarySelection(Actor player, LocalSession session, Vector pos) {
-        player.print("Starting a new polygon at " + pos + ".");
+        if (player.hasPermission("worldedit.hidechatcoords")) {
```
wizjany: um?

RickyB98:
That was left from some previous edits, sorry about that, I pushed a new commit. Anyway, of course this comes from a question I was asked at the server I work at. It's a Factions server and the owner is a Youtuber, hence showing coordinates of "private" places in chat exposes them to the public. Of course playing the "why?" game won't get us anywhere. I thought these commits didn't break anything about the plugin for anyone, while giving the option to other people who might want to use it.

DarkArc:
This feels to me more like something that should be toggled with a commad, than set by a permission. That's just my two cents.

RickyB98:
Perhaps, yes. Permissions looked to me more like a quasi-permanent solution, because I suppose they don't want to toggle it every time they log on or the server restarts. If the idea is accepted and we want to do this file-based, so that opt-outers get saved to a file, sure enough I'm okay to recode it.

wizjany:
as @DarkArc said, using permissions is a really one-way approach, and will cause breakage when people who don't use worldedit.* permissions (i.e. anyone smart) upgrade. it also means users have no way to control it themselves, which means it's imposed on them by the server admin (which only works in cases like you described where the youtuber is the admin)

A toggle would be fine, and if you're worried about having to re-set it upon log-in, note that non-transient localsession variables are persisted. https://github.com/sk89q/WorldEdit/blob/6.1/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java#L88

RickyB98:
Despite the fact that this situation is most likely if the youtuber *is* the admin, otherwise I doubt it they would have WorldEdit permissions, and despite the fact that permission changes might happen anyway with some updates, I'll try and code it so that it becomes a command-based toggleable option. My only doubt is what permission node would you assign to this `//togglecoords` command? (or suggest a better name!) And wouldn't that break permissions anyway? And of course, how do you normally deal with that?

wizjany:
Same way ``//toggleeditwand`` works probably. Default on, perm to turn off.
https://github.com/sk89q/WorldEdit/blob/6.1/worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java#L289
That leaves behavior the same by default, with an added permission to change that.

kenzierocks:
There's significant changes required for this to be merged, plus I think this would be easier to integrate post-translation support, since we would already know what messages contain vectors due to the pass-down of arguments to the translation layer, and could probably implement the check in a single spot. I'm closing this for now.

PR #385, opened 2017-04-03
Exclude players and complex parts from region pastes

caseif:
This PR adds a check to `ForwardExtentCopy` to block player and complex part entities from being pasted with a region. In addition to such behavior not making sense, it is also breaking on Sponge, which actively blocks spawning these entities.

This can be safely back-merged into `feature/sponge-5.1.0`, which does not function properly without this change (albeit with a merge conflict due to the entity-copying routine being wrapped in an if-block).

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitEntityType.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitEntityType.java
@@ -24,6 +24,7 @@ import com.sk89q.worldedit.util.Enums;
 import org.bukkit.entity.Ambient;
 import org.bukkit.entity.Animals;
 import org.bukkit.entity.Boat;
+import org.bukkit.entity.ComplexEntityPart;
 import org.bukkit.entity.Entity;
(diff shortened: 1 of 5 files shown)


me4502:
IMO it'd make sense to have a method that's called `isPasteableEntity` or something, rather than having a method for all of them. This way if it changes in the future, or changes between platforms, it'll be fine.

caseif:
Do you mean in `EntityType`, or in `ForwardExtentCopy` accepting an `Entity` parameter?

caseif:
Agreed, I'll update it shortly.

wizjany:
Agree with @me4502 here. Also it seems for whatever reason that sponge is the only thing that breaks with this...?

sk89q:
Agree too

caseif:
@me4502 You're suggesting an `EntityType#isPasteable` method, or am I misunderstanding?

@wizjany I agree that hard-failing is a dumb behavior, but that's a separate issue to take up with Sponge.

me4502:
Superseded by https://github.com/sk89q/WorldEdit/commit/50f60da69fefc325c9b6186ff03f8fd48ab1ab55

PR #386, opened 2017-04-05
Update .gitignore

vinyldarkscratch:


kenzierocks:
Why do we need to add this?

vinyldarkscratch:
Eclipse creates `.launch` files that contain all of the project's launch options.  This may even include login information if someone is testing their mods by launching Minecraft within Eclipse using their regular account (which, believe me, I've seen many do without realizing).  On top of that, Gradle configures all of the needed launch options for Eclipse projects, and as such any launch options should be added there instead of within Eclipse itself, assuming that everyone should have those options.  😛

vinyldarkscratch:
Will close this for now and submit a new PR for other updates!

PR #387, opened 2017-05-14
Add Support for Minecraft 1.12

md-5:
Minecraft 1.12 requires Java 8 and bundles Guava 21. This change allows usage in that scenario. Unfortunately it's hard/impossible to avoid breakage as Mojang jumped over two years of Guava versions, which is longer than their deprecation cycle.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ buildscript {
 
     configurations.all {
         resolutionStrategy {
-            force 'com.google.guava:guava:17.0'
+            force 'com.google.guava:guava:21.0'
         }
     }
 
--- a/worldedit-core/build.gradle
+++ b/worldedit-core/build.gradle
@@ -5,7 +5,7 @@ dependencies {
     compile 'de.schlichtherle:truezip:6.8.3'
     compile 'rhino:js:1.7R2'
     compile 'org.yaml:snakeyaml:1.9'
-    compile 'com.google.guava:guava:17.0'
+    compile 'com.google.guava:guava:21.0'
     compile 'com.sk89q:jchronic:0.2.4a'
     compile 'com.google.code.findbugs:jsr305:1.3.9'
     compile 'com.thoughtworks.paranamer:paranamer:2.6'
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/ReplaceParser.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/ReplaceParser.java
@@ -30,7 +30,7 @@ import com.sk89q.worldedit.function.pattern.Pattern;
 import com.sk89q.worldedit.util.command.argument.CommandArgs;
 import com.sk89q.worldedit.util.command.composition.SimpleCommand;
 
-import static com.google.common.base.Objects.firstNonNull;
+import static com.google.common.base.MoreObjects.firstNonNull;
 
 public class ReplaceParser extends SimpleCommand<Contextual<? extends RegionFunction>> {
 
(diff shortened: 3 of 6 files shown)


me4502:
Is Guava shaded into WorldEdit currently? If not, this should hopefully not break anything that relies on it as a library.

md-5:
Guava is not shaded into WorldEdit (for Bukkit at least, but I imagine this is the same for at least Forge and derivatives also since MC bundles it). This won't break stuff depending on WorldEdit, but it will make WorldEdit require Guava >=21

stuntguy3000:
Does this PR actually result in a working WorldEdit?

```
[13:39:05 ERROR]: Could not load 'plugins\worldedit-bukkit-6.1.7-SNAPSHOT.jar' in folder 'plugins'
org.bukkit.plugin.InvalidPluginException: java.lang.NoClassDefFoundError: com/sk89q/worldedit/entity/Player
        at org.bukkit.plugin.java.JavaPluginLoader.loadPlugin(JavaPluginLoader.java:133) ~[spigot-1.12-pre2-SNAPSHOT.jar:git-Spigot-429b0e3-5195487]
        at org.bukkit.plugin.SimplePluginManager.loadPlugin(SimplePluginManager.java:326) ~[spigot-1.12-pre2-SNAPSHOT.jar:git-Spigot-429b0e3-5195487]
        at org.bukkit.plugin.SimplePluginManager.loadPlugins(SimplePluginManager.java:248) [spigot-1.12-pre2-SNAPSHOT.jar:git-Spigot-429b0e3-5195487]
        at org.bukkit.craftbukkit.v1_12_R1.CraftServer.loadPlugins(CraftServer.java:300) [spigot-1.12-pre2-SNAPSHOT.jar:git-Spigot-429b0e3-5195487]
        at net.minecraft.server.v1_12_R1.DedicatedServer.init(DedicatedServer.java:205) [spigot-1.12-pre2-SNAPSHOT.jar:git-Spigot-429b0e3-5195487]
        at net.minecraft.server.v1_12_R1.MinecraftServer.run(MinecraftServer.java:544) [spigot-1.12-pre2-SNAPSHOT.jar:git-Spigot-429b0e3-5195487]
        at java.lang.Thread.run(Unknown Source) [?:1.8.0_121]
Caused by: java.lang.NoClassDefFoundError: com/sk89q/worldedit/entity/Player
        at java.lang.Class.forName0(Native Method) ~[?:1.8.0_121]
        at java.lang.Class.forName(Unknown Source) ~[?:1.8.0_121]
        at org.bukkit.plugin.java.PluginClassLoader.<init>(PluginClassLoader.java:64) ~[spigot-1.12-pre2-SNAPSHOT.jar:git-Spigot-429b0e3-5195487]
        at org.bukkit.plugin.java.JavaPluginLoader.loadPlugin(JavaPluginLoader.java:129) ~[spigot-1.12-pre2-SNAPSHOT.jar:git-Spigot-429b0e3-5195487]
        ... 6 more
```

md-5:
(In response to now deleted comment):

Yes, this PR does result in a working worldedit - it was used to produce the WorldEdit and WorldGuard versions on our test server - https://github.com/SpigotMC/SpigotCraft/tree/master/plugins

wizjany:
i assume since mojang is updating on their end that means forge will have the updated one too? this was originally done because bukkit (spigot) and forge had different versions for some time since iirc spigot updated to something higher than what mojang/forge was using.

md-5:
Bukkit has always either used lower (<= MC 1.7) or the same version (>= MC1.8) that Mojang has been using (never higher).

Yes, this change has been done at a Mojang level, so I would assume that change will be carried across to 1.12 servers based on Mojang code.

Brokkonaut:
Wouldn't it be preferable if guava is shaded into WE to allow it to work on older MC versions as well?

wizjany:
shading all of guava would be pretty wasteful. if you look at the WG side we actually did "shade" a few classes that we used from the new guava when bukkit was still on the old one for compatibility, but those were hand picked.

Black-Hole:
One alternative would be to implement Objects.firstNull() without any references to Guava.
Since there is no MoreObjects class in Guava 17.0 and Objects.firstNull() is removed in Guava 21.0.
That way WorldEdit could be compatible with Minecraft 1.12 and very old versions.

wizjany:
that's what #389 did

PR #388, opened 2017-05-16
Update blocks.json to 1.12

boy0001:
used https://github.com/wizjany/ForgeUtils

### DIFF
--- a/worldedit-core/src/main/resources/com/sk89q/worldedit/world/registry/blocks.json
+++ b/worldedit-core/src/main/resources/com/sk89q/worldedit/world/registry/blocks.json
@@ -1367,7 +1367,7 @@
     "legacyId": 26,
     "id": "minecraft:bed",
     "unlocalizedName": "tile.bed",
-    "localizedName": "Bed",
+    "localizedName": "tile.bed.name",
     "states": {
       "facing": {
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/resources/com/sk89q/worldedit/world/registry/blocks.json
```diff
@@ -1700,7 +1700,7 @@
       "slipperiness": 0.6,
       "renderedAsNormalBlock": false,
       "liquid": false,
-      "solid": true,
+      "solid": false,
```
Minecrell: This looks wrong, cobweb is solid in the Minecraft server

boy0001: It appears to have changed in 1.12

kashike: No, it hasn't.

Minecrell: Yes, it hasn't changed. I've generated the file myself and this was still true. How did you generate it? You may have used the wrong method depending on your mappings.

PR #389, opened 2017-05-20
updated guava + blocks.json

boy0001:
updated guava + blocks.json for 1.12]
This is backwards compatible and won't break the plugin for older versions


TomyLobo:
No descriptioin provided => no reason to merge?

PR #390, opened 2017-06-08
Sponge: Add support for Minecraft 1.12

Minecrell:
No changes at all in adapter class, just recompiled with Minecraft 1.12. The adapter for 1.11.2 would probably have worked without any changes..

Adapter PR: TODO


Minecrell:
I guess it doesn't make sense to merge the binary files directly, so I'll leave a link to the adapter PR instead: https://github.com/me4502/WorldEdit-Sponge-Adapters/pull/1 (why are they on separate repositories anyway?)

PR #393, opened 2017-07-29
Compability to Java 9

FilipeRamalho:
New version-string scheme causes an ArrayIndexOutOfBoundException when trying to access ver[1] after the String split. Fixed by checkingif the scheme contains a dot (Old scheme) or not and then acting according to that. http://openjdk.java.net/jeps/223

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/Java8Detector.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/Java8Detector.java
@@ -22,10 +22,9 @@ import com.sk89q.worldedit.WorldEdit;
 
 public final class Java8Detector {
 
-    private static int major;
-
     public static void notifyIfNot8() {
 
+        int major;
         try {
             Class.forName("java.lang.Runtime.Version");
             major = 9;
--- a/worldedit-sponge/build.gradle
+++ b/worldedit-sponge/build.gradle
@@ -13,7 +13,7 @@ plugins {
 
 dependencies {
     compile project(':worldedit-core')
-    compile 'org.spongepowered:spongeapi:7.0.0-SNAPSHOT'
+    compile 'org.spongepowered:spongeapi:6.0.0'
     testCompile group: 'org.mockito', name: 'mockito-core', version:'1.9.0-rc1'
 }
 


kenzierocks:
Alternatively, we could drop support for Java 7 and remove this class. Has Minecraft dropped it yet?

FilipeRamalho:
It doesn't seem to have dropped it https://help.mojang.com/customer/en/portal/articles/2636196-minecraft---java-6-and-7-support a

DarkArc:
@kenzierocks https://help.mojang.com/customer/en/portal/articles/2636196-minecraft---java-6-and-7-support looks like it's on its way out... We're probably good to just drop Java 7 support, and the checker all together at this point... A Java 7 install should be fairly rare these days

wizjany:
MC 1.12 is the first version that requires Java 8, but WorldEdit should preferably be backwards compatible just a bit more.

As for the PR, it could probably use a bit less code duplication. I've also taken a quick look at the spec (http://openjdk.java.net/jeps/223) and it seems that just doing Integer.parseInt won't work once you have updates/security/whatever in the version numbers. (e.g., "9+100" would fail the contains("."), but then also fail the parseInt).

14mRh4X0r:
It would probably suffice to check for the existence of `java.lang.Runtime.Version`, and do the current check if the class doesn't exist.

FilipeRamalho:
So i looked into the new version scheme and i figured out that the best solution would be to check the first charachter of ` System.getProperty("java.version") `. 

**After 9 it's 9
Before 9 it's 1. and then the Major version**


- _To check 9 see if it starts with 9_
- _To check 8 see if it starts with 1.8_
- _To check 7 see if it starts with 1.7_
- And so on...


Today I am not able to code that differences, I will try to make a commit tomorrow (this time with an IDE 😛 )

wizjany:
Can anyone tell me if there's an issue with @14mRh4X0r's suggestion? Everyone seems to really really like parsing strings and hoping they are deterministic.

FilipeRamalho:
There seems to be no issue with that. I am not sure how test for the existence of certain jdk classes, but if he is able to code, the i go all for it.

wizjany:
```java
try {
    Class.forName("java.lang.Runtime.Version");
} catch (ClassNotFoundException ...
```

FilipeRamalho:
Ok i will try that. Tomorrow i submit a new commit with better major check

FilipeRamalho:
@14mRh4X0r 
> You could then even use Runtime.Version to compare later versions of  Java.

Well that won't work until we only support Java 9 and that won't be any time soon.  [See this commit](https://github.com/sk89q/WorldEdit/pull/393/commits/7b4a4c3558ea5810b7b7de08c49d2eea3ce3fe21)

FilipeRamalho:
I am not sure how, but the last commit broke Sponge

"Execution failed for task ':worldedit-sponge:compileJava'."

kenzierocks:
I implemented a simpler version of this that is more targeted at the intention of this class. Thank you for raising the issue!

PR #394, opened 2017-08-02
Compability to Java 9 (cleaner then #393)

IsaiahPatton:
Fixes compability with Java 9.
This is a cleaner version of #393


wizjany:
You say cleaner but it still looks like there's a bunch of assumptions being made.
Assumptions are the reason it doesn't work with java 9 in the first place, so it would be great if we could cut down on those.

IsaiahPatton:
Did Java change the _ to +?

wizjany:
i don't know i didn't fully read the spec, just looked at a table of examples http://i.imgur.com/LFwNvyY.png

wizjany:
why are you still just bandaiding this? we have two pull requests now that are just bandaids.

IsaiahPatton:
There is a "squash" option when pulling.

wizjany:
What does squashing have to do with anything?
I'm asking about the method that this and #393 are attempting to employ to "fix" the issue.

IsaiahPatton:
I've fixed all version formats from http://i.imgur.com/LFwNvyY.png

wizjany:
I feel like I'm talking to a brick wall here.

PR #395, opened 2017-09-02
Fix affine inverse

boy0001:
The AffineTransform is used for rotation. This fixes a minor issue with the inverse not being correct for specific angles, which can cause minor issues with craftscripts and deforms. 

Easiest to replicate the problem using code:
```Java
public static void main(String[] args) {
    AffineTransform af = new AffineTransform().rotateX(-90).rotateY(-90);
    AffineTransform inv = af.inverse();

    Vector v = new Vector(100, 100, 100);
    System.out.println(inv.apply(af.apply(v)) + " != " + v);
}
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/math/transform/AffineTransform.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/math/transform/AffineTransform.java
@@ -169,13 +169,13 @@ public class AffineTransform implements Transform {
                 (m01 * m12 - m11 * m02) / det,
                 (m01 * (m22 * m13 - m12 * m23) + m02 * (m11 * m23 - m21 * m13)
                         - m03 * (m11 * m22 - m21 * m12)) / det,
-                (m12 * m20 - m22 * m10) / det,
+                (m20 * m12 - m10 * m22) / det,
                 (m00 * m22 - m20 * m02) / det,
-                (m02 * m10 - m12 * m00) / det,
+                (m10 * m02 - m00 * m12) / det,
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/math/transform/AffineTransform.java
```diff
@@ -165,17 +165,17 @@ public AffineTransform inverse() {
         double det = this.determinant();
         return new AffineTransform(
                 (m11 * m22 - m21 * m12) / det,
-                (m21 * m01 - m01 * m22) / det,
+                (m02 * m21 - m22 * m01) / det,
```
kenzierocks: this `m02` is literally the only change you made, can you revert all the other multiplications to their original order to shrink the diff?

boy0001: yeh

TomyLobo: I think he meant all of them, including this line.
Essentially you're only changing a 1 to a 2 and the diff should reflect that, so people dont have to play "where's waldo" to see what's going on :)

TomyLobo:
yay, perfect, thanks for the patch and for spotting this in the first place

TomyLobo:
@boy0001 if you want to PR your example as a test case, that would be nice.
Should probably be under worldedit-core/src/test/java/com/sk89q/worldedit/math/transform/AffineTransformTest.java

PR #396, opened 2017-09-02
Fix FallbackRegistrationListener

boy0001:
The FallbackRegistrationListener is used when Bukkit's SimpleCommandMap isn't present (or has been modified). e.g. Thermos, or if some plugin inserts a custom command map. 

Easiest way to replicate these conditions is to set the initial command map to null:
https://github.com/sk89q/WorldEdit/blob/master/worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/CommandRegistration.java#L69


sk89q:
I believe it's also used for the dynamic /blah.js commands.

boy0001:
It should be `/cs blah.js <arguments>`

wizjany:
it's also available by using just /<name>.js

PR #398, opened 2017-11-02
Patched calculation exploit

Legoman99573:
The famous calculation to crash the server has finally been patched somewhat. It may lag your server until you get the answer if you use the same command to attempt to crash the server.

Best bet still is to be careful who you give the calc perms to.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
@@ -63,12 +63,11 @@ import com.sk89q.worldedit.util.formatting.component.CommandListBox;
 import com.sk89q.worldedit.util.formatting.component.CommandUsageBox;
 import com.sk89q.worldedit.world.World;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
+import java.util.concurrent.*;
 
 import static com.sk89q.minecraft.util.commands.Logging.LogMode.PLACEMENT;
+import static java.lang.Integer.*;
 
 /**
  * Utility commands.
@@ -123,7 +122,7 @@ public class UtilityCommands {
         Pattern pattern = we.getBlockPattern(player, args.getString(0));
         double radius = Math.max(1, args.getDouble(1));
         we.checkMaxRadius(radius);
-        int depth = args.argsLength() > 2 ? Math.max(1, args.getInteger(2)) : Integer.MAX_VALUE;
+        int depth = args.argsLength() > 2 ? Math.max(1, args.getInteger(2)) : MAX_VALUE;
 
         Vector pos = session.getPlacementPosition(player);
         int affected = 0;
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
```diff
@@ -515,17 +514,22 @@ public void remove(Actor actor, CommandContext args) throws WorldEditException,
...
-            actor.print("= " + expression.evaluate());
-        } catch (EvaluationException e) {
-            actor.printError(String.format(
-                    "'%s' could not be parsed as a valid expression", input));
-        } catch (ExpressionException e) {
-            actor.printError(String.format(
-                    "'%s' could not be evaluated (error: %s)", input, e.getMessage()));
-        }
+    public void calc(final Actor actor, @Text final String input) throws CommandException {
+        // Dev should have threaded it years ago to prevent the exploit from happening.
```
ryantheleach: Cut the snark, comments like this won't make it into the final and makes it less likely they will pull your PR.

worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
```diff
@@ -515,17 +514,22 @@ public void remove(Actor actor, CommandContext args) throws WorldEditException,
...
+        Executors.newSingleThreadExecutor().execute(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    actor.print("Calculating...");
```
ryantheleach: Not 100% sure if sending this async is fine.

worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
```diff
@@ -63,12 +63,11 @@
 import com.sk89q.worldedit.util.formatting.component.CommandUsageBox;
 import com.sk89q.worldedit.world.World;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
ryantheleach: Don't change imports for the sake of it, follow the code conventions.

worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
```diff
@@ -63,12 +63,11 @@
...
+import java.util.concurrent.*;
 
 import static com.sk89q.minecraft.util.commands.Logging.LogMode.PLACEMENT;
+import static java.lang.Integer.*;
```
ryantheleach: Does worldedit elsewhere use static imports for max_value?

ryantheleach:
"The famous calculation"

If you are going to post a PR, could you at least post how to test it.

DarkArc:
I'm guessing the "famous calculation" is the one mentioned in the comments on this video https://m.youtube.com/watch?v=mOKtWJ26Fyk

First I'm hearing anything of this. 

IMO, might be better to just add a separate permission for a simple calculator that only accepts the pattern `[0-9 +-/*]`. Short of implementing a maximum execution time system, I think that makes the most sense to "properly" patch this; as the author notes, this fix can still cause significant lag -- that really shouldn't be something you have to worry about when handing out a calculator.

wizjany:
> Best bet still is to be careful who you give the calc perms to.

Which is why we fixed this by adding the permission node when this was first reported 2 years ago.

Also all the things @ryantheleach said. Except chat is fine async I'm pretty sure. But you shouldn't send "Calculating..." unless it's actually taking a while. Take a look at the way WG does report and profile.

Legoman99573:
@wizjany all i did was make another thread, which keeps the server from crashing the first time. need a little more than a permission node to fully fix it.

worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
Legoman99573: newSingleThreadExecutor() opens a new thread for the execution. Didnt try to make it sound like an async event. if 1+1 was used, it would print no problem

PR #399, opened 2017-11-02
Calculation now threaded (Fixes for loop exploit)

Legoman99573:
My last commit wasnt as nice, but this one has fixes and  ❤️  s now.

Each thing added in comments will tell you what it does and how its executed including stopping the thread when its complete. This only affect //calc command which is good for `for` loops, which is dangerous without being in a thread.

This does not run asynchronously, this runs in a seperate thread to prevent the server from not responding.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
@@ -63,11 +63,13 @@ import com.sk89q.worldedit.util.formatting.component.CommandListBox;
 import com.sk89q.worldedit.util.formatting.component.CommandUsageBox;
 import com.sk89q.worldedit.world.World;
 
-import java.util.*;
-import java.util.concurrent.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.Executors;
 
 import static com.sk89q.minecraft.util.commands.Logging.LogMode.PLACEMENT;
-import static java.lang.Integer.*;
 
 /**
  * Utility commands.
@@ -81,11 +83,11 @@ public class UtilityCommands {
     }
 
     @Command(
-        aliases = { "/fill" },
-        usage = "<block> <radius> [depth]",
-        desc = "Fill a hole",
-        min = 2,
-        max = 3
+            aliases = { "/fill" },
+            usage = "<block> <radius> [depth]",
+            desc = "Fill a hole",
+            min = 2,
+            max = 3
     )
     @CommandPermissions("worldedit.fill")
     @Logging(PLACEMENT)
@@ -109,11 +111,11 @@ public class UtilityCommands {
     }
 
     @Command(
-        aliases = { "/fillr" },
-        usage = "<block> <radius> [depth]",
-        desc = "Fill a hole recursively",
-        min = 2,
-        max = 3
+            aliases = { "/fillr" },
+            usage = "<block> <radius> [depth]",
+            desc = "Fill a hole recursively",
+            min = 2,
+            max = 3
     )
     @CommandPermissions("worldedit.fill.recursive")
     @Logging(PLACEMENT)
@@ -122,7 +124,7 @@ public class UtilityCommands {
         Pattern pattern = we.getBlockPattern(player, args.getString(0));
         double radius = Math.max(1, args.getDouble(1));
         we.checkMaxRadius(radius);
-        int depth = args.argsLength() > 2 ? Math.max(1, args.getInteger(2)) : MAX_VALUE;
+        int depth = args.argsLength() > 2 ? Math.max(1, args.getInteger(2)) : Integer.MAX_VALUE;
 
         Vector pos = session.getPlacementPosition(player);
         int affected = 0;
@@ -137,11 +139,11 @@ public class UtilityCommands {
     }
 
     @Command(
-        aliases = { "/drain" },
-        usage = "<radius>",
-        desc = "Drain a pool",
-        min = 1,
-        max = 1
+            aliases = { "/drain" },
+            usage = "<radius>",
+            desc = "Drain a pool",
+            min = 1,
+            max = 1
     )
     @CommandPermissions("worldedit.drain")
     @Logging(PLACEMENT)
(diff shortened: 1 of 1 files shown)


Legoman99573:
If there is any questions, let me know. I get the perm exists, but doesnt prevent OPs from running it.

worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
```diff
@@ -82,11 +83,11 @@ public UtilityCommands(WorldEdit we) {
     }
 
     @Command(
-        aliases = { "/fill" },
-        usage = "<block> <radius> [depth]",
-        desc = "Fill a hole",
-        min = 2,
-        max = 3
+            aliases = { "/fill" },
```
kenzierocks: Please undo all this formatting, it's completely unnecessary.

worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
```diff
@@ -510,30 +511,36 @@ public void remove(Actor actor, CommandContext args) throws WorldEditException,
...
-            actor.print("= " + expression.evaluate());
-        } catch (EvaluationException e) {
-            actor.printError(String.format(
-                    "'%s' could not be parsed as a valid expression", input));
-        } catch (ExpressionException e) {
-            actor.printError(String.format(
-                    "'%s' could not be evaluated (error: %s)", input, e.getMessage()));
-        }
+    public void calc(final Actor actor, @Text final String input) throws CommandException {
+        // executing as a thread should keep the server from crashing when using for loops.
```
kenzierocks: This isn't a proper solution, and may still cause crashes if enough commands are executed. A better solution would be to timeout calculations after a certain time.

worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
```diff
@@ -510,30 +511,36 @@ public void remove(Actor actor, CommandContext args) throws WorldEditException,
...
+                try {
+                    Expression expression = Expression.compile(input);
+                    actor.print(input + " = " + expression.evaluate());
+                } catch (EvaluationException e) {
+                    actor.printError(String.format("'%s' could not be parsed as a valid expression", input));
+                } catch (ExpressionException e) {
+                    actor.printError(String.format("'%s' could not be evaluated (error: %s)", input, e.getMessage()));
+                }
+                // shuts the thread down so its not sitting somewhere aimlessly after executing fully
+                Thread.currentThread().stop();
```
kenzierocks: Thread.stop is completely deprecated and irresponsible to use when using an executor. Rely on the executor service to shut down the thread using ExecutorService.shutdown.

kenzierocks:
>This does not run asynchronously, this runs in a seperate thread

This is basically the definition of asynchronous code.

worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
Legoman99573: ok, will change the code and i am not sure how the formatting is happening.

Legoman99573:
I will work on changing the code up.

kenzierocks:
There's no need for a temporary solution - if someone is OP, they should be expected to have the ability to cause damage to the server. This is what granular permissions are for -- if someone shouldn't have these potentially server-destroying abilities, they should not have the permission for it.

ekyl:
I'd like to urge that this update is forced and/or enthusiastically
suggested to server owners. When I discovered the exploit in February, I
immediately told servers to update their version to fix it. Nobody did that
at all, and when I accidentally let the knowledge of the exploit get out of
my control, it prompted plugin creation just to block the command...

Anyways, I'm glad to see this is formally being fixed. I apologize for any
chaos it's created. I do think I deserve to say to the developer who shot
down my plea for a patch to older versions, after I told them before
reporting the exploit to anyone else: I told you so.

By the way, does replying to these emails work? I guess I'll find out.

Legoman99573:
I have gotten the message. I am using the fix against the latest worldedit update.

ekyl:
I still find it ridiculous that server owners would sooner write a [custom plugin](https://www.spigotmc.org/resources/worldedit-command-crash-fix.35962/) than just update their WorldEdit version.

Also: after nearly 2 years... Finally, some closure :)

Legoman99573:
@ekyl is correct, the //calc exploit has basically gotten out of control at this point. Everyone knows it and they will use it any chance they get to it.

kenzierocks:
I think "out of control" and "everyone knows it" are exaggerations that are not helping your case. If it was such a big problem, we'd be getting many more reports, but as it is, most servers simply don't allow unknown players access to these powerful aspects of worldedit. This is a pretty small problem, especially with the permission node in place.

The only people affected by this are people who run servers with OPs that apparently cannot be trusted to not use a command that would kill a server. I assume that if they're OPs, they also have access to commands like `/setblock`, or `//set`, or `//fill`, which are also server-ruining if used properly. To put it simply, if this is a problem for you, this will not be the only problem you have. I'm open to fixing it, but this simply isn't that big of a deal.

Legoman99573:
True, but this is a bad thing for people who dont know how worldedit works and someone comes on and uses the command.

kenzierocks:
So, you give all players permission to use every command when they join? That's just asking for trouble. The default is not set up like this, and if you OP every player who joins, then you should expect your server to be taken down eventually.

Legoman99573:
I didnt say i give players all perms. I am saying that a 16GB server may handle it no problem, but take shared hosts as an example, most only allow 4 or 8 GB of RAM and the command can crash alot faster on a 4GB server than an 8GB server

ekyl:
@kenzierocks I agree completely, but it's all about idiot proofing (especially when most of the users demonstrate that they are indeed idiots!)

Legoman99573:
I will probably close the PR as this is getting nowhere at all. All we did was make an argument about a change to partially fix the exploit. The perm only fixed 50% of it imo, but i guess we all have different thoughts of a "fixed exploit".

kenzierocks:
Please see #400, I think you'll be happy with what I've done.

PR #400, opened 2017-11-02
Attach a configurable timeout to expression evaluation

kenzierocks:
This solves the problems mentioned in #399 in a more elegant way, and in a wider scope.


worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/Expression.java
```diff
@@ -115,9 +129,30 @@ public double evaluate(double... values) throws EvaluationException {
...
+            Future<Double> result = evalThread.submit(new Callable<Double>() {
+                @Override
+                public Double call() throws Exception {
+                    return root.getValue();
+                }
+            });
+            try {
+                return result.get(WorldEdit.getInstance().getConfiguration().calculationTimeout, TimeUnit.MILLISECONDS);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
```
wizjany: I haven't worked with futures much, but doesn't this exception get thrown on the main thread?
I would think that `Future.get(...)` will interrupt the thread where the task is being run? I might be totally wrong though.

sk89q: It gets thrown if something completely unrelated to this code interrupts this current thread.

kenzierocks: Yes, what @sk89q said - this is just proper handling of the interrupt should we happen to get hit with one. It's an exceptional case.

wizjany: Ah ok I was not at all off. Just seemed weird how we're throwing and rethrowing runtime exceptions like candy.
Also for some reason all the expression tests are failing now. Aside from that this looks pretty good. Definitely better than the last one.

sk89q:
It's only for getting the value and not for stopping execution.

DarkArc:
Oh, that's far less exciting then; shame :\

Legoman99573:
This is wht i was expecting sort of, but this seems like it should work just fine.

kenzierocks:
@wizjany I added a platform to fix the problem, but perhaps a system property would work better?

worldedit-core/src/test/java/com/sk89q/worldedit/internal/expression/ExpressionPlatform.java
```diff
@@ -0,0 +1,77 @@
+package com.sk89q.worldedit.internal.expression;
```
wizjany: missing header causing checkstyle failures on build

kenzierocks: . -.

will fix tonight

kenzierocks: fixed :D

me4502:
Is this ready to be merged?

kenzierocks:
I think so, but perhaps I should add some tests for the timeout.

me4502:
Could you please rebase this, and add tests if you think it's best to do so?

Legoman99573:
@me4502 yeah, my PR broke it :/

kenzierocks:
I'll take a look by next week.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/Expression.java
```diff
@@ -69,6 +78,22 @@
 public class Expression {
 
     private static final ThreadLocal<Stack<Expression>> instance = new ThreadLocal<>();
+    private static final ExecutorService evalThread = Executors.newCachedThreadPool(
+            new ThreadFactoryBuilder()
+                    .setDaemon(true)
+                    .setNameFormat("worldedit-expression-eval-%d")
+                    .build());
+
+    // A workaround for testing -- we can't easily create a LocalConfiguration
```
kenzierocks: @me4502 Want to explain why `LocalConfiguration` can't be created here -- it uses `BlockType#getId`, which relies on the underlying `Platform` to provide registries, etc. But this test is in `-core`, where we have no `Platform` implementation. I think the best solution is to remove references to `BlockType` -- we should just be writing the string IDs there now. I'm not sure if that change should be in this PR as well though.

me4502: Hmm, I'll take a look into uncoupling BlockType#getId() from the Platform, it doesn't overly make sense for it to depend on it.

me4502:
Thanks

TomyLobo:
Cool, might as well get rid of the 256 iteration limit now, or at least make it configurable.
What do you think?

kenzierocks:
Yes, we can do that.

PR #401, opened 2017-11-03
Make DinnerPermsResolver group prefix configurable

lucko:
Allows you to adapt the DinnerPerms resolver to lookup group memberships using an alternative node prefix.

Not all permission plugins implement the `group.<name>` format - and on the other hand, using that format for stuff like WorldGuard regions can create conflicts with other usages of the default `group.` format.

The current format is still used by default.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/wepif/DinnerPermsResolver.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/wepif/DinnerPermsResolver.java
@@ -32,15 +32,27 @@ import java.util.List;
 
 public class DinnerPermsResolver implements PermissionsResolver {
 
+    // the default group prefix
     public static final String GROUP_PREFIX = "group.";
+
     protected final Server server;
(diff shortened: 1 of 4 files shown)


wizjany:
why are the other things that use the "group.x" format (which has become standard since dinnerperms was added to bukkit) not also configurable then? :thinking:

lucko:
I'm sure some of them are. :)

The issue is more that it's quite common to use the WorldGuard `GroupDomain` as more of a `PermissionDomain` - although the latter doesn't exist, and then handle access to the region using the permission manager, as opposed to adding lots of groups to the region. (I appreciate this isn't a WorldEdit issue, but as I'm sure you're aware, WorldGuard uses WEPIF to evaluate group domains)

This is more powerful in a number of cases, as it allows you to grant access to a domain in specific context. (whether that be per-server, per-world or whatever.)

lucko:
Closed following our discussion on IRC. :)

PR #402, opened 2017-11-23
Calculation Config missing and typo [bugfix/timed-calc]

Legoman99573:
Added missing configuration option as its not generated when you reset the config in the bugfix/timed-calc branch, plus also fixed a typo in the comments on top of the configuration file.


kenzierocks:
Good catch, thank you.

Legoman99573:
No problem ;)

PR #403, opened 2017-12-14
Fix //snow doesn't put snow on all types of leaves

SlimSoftware:
For example, //snow didn't work on Oak leaves while it did work as you would expect on Dark Oak leaves. Fixed by adding leaves2 to translucent block list and checking if the block to put snow on is a leaf block when executing //snow.


me4502:
Thanks

PR #408, opened 2018-03-25
Fix loading and converting signs in old schematics

Brokkonaut:
Fixes https://dev.enginehub.org/youtrack/issue/WORLDEDIT-3642

* Sign texts not starting with { are never valid json objects, so don't try to parse
* Else try to parse as json, but fall back to read sign text as string if not parseable


Brokkonaut:
Any news on this? Is there anything I should change until it can be pulled?

PR #411, opened 2018-03-25
Clean up remnants of old dependencies

Orhideous:
Related commit: 46dff16a977eb9244d8893dbf58a066d2e600c87

This also boy0001/FastAsyncWorldedit#856


wizjany:
this breaks backwards compatibility, that's why that method was copied to our codebase.

PR #412, opened 2018-05-05
Prevent double transforming for entities

Qveshn:
The bug was in command //paste after command //rotate \<angle>.
Entities in clipboard were rotated by 2*\<angle>


Qveshn:
I know that I have made "bad" solution. I only have showed you "basic" solution. I've just pointed to the place of the code where there was a problem.
You should look more closely at the code.

For example, I do not understand the concept of next lines:
>> currentTransform = currentTransform.combine(transform);
I do not understand, why we need to remember "double" transform.
Also I do not understand the concept of "lastVisitor = blockVisitor;"

So, you should look more closely at the code.

Qveshn:
I have done new pull request https://github.com/sk89q/WorldEdit/pull/414
Both requests have the same effect on code.
This one has minimum code changes, but I think that https://github.com/sk89q/WorldEdit/pull/414 is more nice

PR #415, opened 2018-05-19
Customizable command log format (Date+Time are now available)

Qveshn:
**Added customizable command log format (Date+Time are now available):**
Added the option "format" to config files (config.yml + worldedit.properties). By default, current date and time are also displayed.
The formatting function is implemented by analogy with the class java.util.logging.SimpleFormatter
Notes: I also added comments in default config.yml. I'm not sure that this option and description will be understandable for most users, but it allows to customize the format of the command log =\

**This feature was tested**


me4502:
Thanks

PR #416, opened 2018-05-19
Fix entities double transformation copy-rotate-paste and stack

Qveshn:
The bug was during command //paste after command //rotate angle.
Entities in clipboard were rotated by 2*angle during paste command
Also, the bug was during command //stack n
Selected region was copied n-times with shift, but entities were shifted one region more

**Tested with commands stack,copy,rotate,paste**
1. //stack 3 up (checked field "repetitions")
2. //copy -e //rotate 45 //rotate 45 //paste (checked transformation)


Qveshn:
I have made new PR https://github.com/sk89q/WorldEdit/pull/416
This PR is more correct and simpler
=)

me4502:
Thanks

PR #419, opened 2018-07-09
Fix ascent and descent commands.

Cebtenzzre:
For seven years and two days, the algorithm for ascent and descent has had an off-by-one error, and started at the wrong level.
Also, the message for failed descent has apparently never been right.
I figured I'd follow Java's conventions (as I see them) and have a body for the loop, so instead of incrementing the counter within the loop condition, I refactored the loop and put in a break statement.  I prefer the shorter version, but POLA wins in a foreign code base.

I tested this on my server and it seems to work fine.


me4502:
What was the old behaviour that was broken, and what does the new behaviour do in its place?

I’m not at a computer right now, but from quick inspection it should be logically equivalent except for the typo fix in the message.

Cebtenzzre:
Yeah, the main issue was that it told you that you had moved successfully even when you hadn't.  Sorry for not being clear about the important details xD

sk89q:
Jesus seven years 

🕐🕝🕘

me4502:
Thanks

PR #422, opened 2018-09-02
Add 1.13 chunk format for snapshots

Brokkonaut:
Fixes WORLDEDIT-3714

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/storage/ChunkStore.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/storage/ChunkStore.java
@@ -27,6 +27,7 @@ import com.sk89q.worldedit.Vector2D;
 import com.sk89q.worldedit.world.DataException;
 import com.sk89q.worldedit.world.World;
 import com.sk89q.worldedit.world.chunk.AnvilChunk;
+import com.sk89q.worldedit.world.chunk.AnvilChunk13;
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/storage/ChunkStore.java
```diff
@@ -78,7 +79,32 @@ public static BlockVector2D toChunk(Vector position) {
...
+                }
+            }
+        }
+
+        if (tag == null) {
+            throw new ChunkStoreException("Missing root 'Level' tag");
+        }
+
+        int dataVersion = rootTag.getInt("DataVersion");
+        if (dataVersion >= 1519) { // data version of Minecraft 1.13
```
kenzierocks: Perhaps extract `1519` to a constant named `DATA_VERSION_MC_1_13`?

Then you won't need the comment, and it'll be easier to read in-line.

Brokkonaut: Ok I changed that.

Brokkonaut:
Parts of the new format are documented here: https://minecraft.gamepedia.com/Chunk_format
For the interpretation of the BlockStates array I used the decompiled sources (class DataBits in CB)

I tested it with many large selections in my current 1.13 world and had no problems restoring things, so I think I got everything correct.

me4502:
Thanks

PR #423, opened 2018-09-04
fix sponge schematic offset

Gerrygames:
fixes https://dev.enginehub.org/youtrack/issue/WORLDEDIT-3690


me4502:
Is this correct to Sponge Schematics generated by other plugins?

Gerrygames:
I don‘t know any other plugin that generates Sponge Schematics. The Sponge Schematic does not support the WEOffset anyway, so it would not be included in Schematics from other plugins. I did test this fix with Schematics saved by World Edit and Schematics gnerated with my MCEdit to Sponge Schematic Converter.

me4502:
Yeah fair, thanks

PR #425, opened 2018-10-01
Implement some optional methods in ForgeWorld

kenzierocks:
A few optimization-related methods and weather weren't implemented in ForgeWorld. Implementing these will ensure a consistent experience, although I don't see any of the weather methods currently used.


me4502:
Thanks

PR #426, opened 2018-10-01
Cleanup javadoc warnings

kenzierocks:
Compiling with Java 10 produced some Javadoc warnings because they're technically written in HTML, and some characters are forbidden without proper encoding.

This PR simply cleans up these warnings.


me4502:
Thanks

PR #427, opened 2018-10-02
Update maze.js to work on modern WorldEdit

dashaw92:
Changes:
 - Created helper function `add` to replace calls to `Location#add` with as the method no longer exists in modern WorldEdit.
 - Convert locations in calls to `setBlock` to `Vector`s.


contrib/craftscripts/maze.js
```diff
@@ -23,19 +23,19 @@ importPackage(Packages.com.sk89q.worldedit.blocks);
...
-usage += "• e places air blocks\n";
-usage += "• a places air blocks only\n";
-usage += "• v creates a vertical maze\n";
+usage += "� i adds an entry and an exit\n";
+usage += "� y places the entry and the exit randomly\n";
+usage += "� f adds a floor\n";
+usage += "� c adds a ceiling\n";
+usage += "� e places air blocks\n";
+usage += "� a places air blocks only\n";
+usage += "� v creates a vertical maze\n";
```
TomyLobo: I think you messed up the encoding on this file.
The encoding is supposed to be UTF-8, now it's Windows-1252

contrib/craftscripts/maze.js
```diff
@@ -23,19 +23,19 @@ importPackage(Packages.com.sk89q.worldedit.blocks);
...
-usage += "• e places air blocks\n";
-usage += "• a places air blocks only\n";
-usage += "• v creates a vertical maze\n";
+usage += "• i adds an entry and an exit\n";
+usage += "• y places the entry and the exit randomly\n";
+usage += "• f adds a floor\n";
+usage += "• c adds a ceiling\n";
+usage += "• e places air blocks\n";
+usage += "• a places air blocks only\n";
+usage += "• v creates a vertical maze\n";
```
TomyLobo: Now you have both a broken U+0095 control character and the bullet point in there.
Also, you have an UTF-8 BOM in there that wasn't there before.
I suggest you use Notepad++ to edit this and go to View -> Show Symbols -> All characters to show the broken characters.

dashaw92:
With all due respect, I'm just going to move on. This is a script that's been untouched for 5 years, and I took the time to update it so it works on 1.13 builds of WorldEdit. Not really sure how having broken source files ([and still using them as working examples](http://wiki.sk89q.com/wiki/WorldEdit/Scripting#Bundled_scripts)) is acceptable by the standards portrayed in this PR, but a file encoding issue is not. Have a nice day.

contrib/craftscripts/maze.js
TomyLobo: Also, make sure you go to Encoding -> Convert to UTF-8 (without BOM!) before you save.

TomyLobo:
The problem is that I can't make a judgment call on the code changes itself.
I haven't been following API changes in years.

However, I do know that this file wasn't and still isn't encoded properly. That I can tell.
* UTF-8 BOM is a "MAY" in the standard, meaning a conforming parser can, but does not have to ignore it.
  So, even if it works with the current parser, there can be problems in the future.
* Initially, you had invalid UTF-8 sequences in there. In addition to being unreadable, this is going to make some parsers outright reject your code
* Your last iteration has unicode C1 control character U+0095 around the bullet point characters.
  Since these are control characters, they are invalid in some places.
  One such place I'm aware of is in an XML 1.1 document.

So in summary, while your code may or may not look fine, it definitely isn't encoded fine.
I cannot just greenlight it and break the next architectural overhaul with invisible characters.

You didn't see them, I only saw them because GitHub showed the lines as changed.
Do you expect the next person to be so lucky?

Now I want you to pick one of these options:
* You keep your current stance, think I'm a big idiot for insisting on correct encoding and keep contributing code like that to other places and have them react similarly.
  That'll probably get this PR rejected.
* You fix these encoding issues yourself
  That'll probably get this PR accepted.
* I help you fix the encoding issues
  That'll probably get this PR accepted.

I don't have any strong preference one way over the other. Which way do you prefer?

dashaw92:
No, I fully understand that this is not your fault. You are correct by rejecting this due to file encoding issues if that is how other PRs are accepted on this project. I just simply don't feel like spending any more time on this, after all it's just a maze generator, and I have a working copy now, which was my original goal. Feel free to reject this PR, I have no intention of resolving any more issues with it here.

PR #428, opened 2018-10-04
Switch to a license plugin for headers

kenzierocks:
Instead of using checkstyle and manually fixing any differences between headers, automatically generating them with a license plugin will allow for a much smoother process.


me4502:
Thanks

PR #430, opened 2018-10-12
Chunk Batching

kenzierocks:
This change adds an extent to batch changes by chunk, and also by region (a la region files). This results in vastly improved performance (approximately 20k blocks per second more when testing in dev).


ryantheleach:
@kenzierocks > " (approximately 20k blocks per second more when testing in dev)."

Mind telling us out of what? so that we can see it percentage wise?

worldedit-core/src/main/java/com/sk89q/worldedit/extent/reorder/ChunkBatchingExtent.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+public class ChunkBatchingExtent extends AbstractDelegateExtent {
+
+    /**
+     * Comparator optimized for sorting chunks by the region file they reside
+     * in. This allows for file caches to be used while loading the chunk.
+     */
+    private static final Comparator<Vector2D> REGION_OPTIMIZED_SORT =
+            Comparator.<Vector2D, Vector2D>comparing(vec -> vec.divide(32).floor(), Vector2D.COMPARING_GRID_ARRANGEMENT)
+                    .thenComparing(Vector2D.COMPARING_GRID_ARRANGEMENT);
```
DarkArc: Smart, I tried doing it just based off the chunk, didn't seem to be as beneficial. Granted that was also in Sponge and I think the overhead of cause tracking was obliterating any optimizations :cry:.

kenzierocks: Yea, I'm not sure how much it speeds us up to sort by region, but the overall change was from `50k` to `70k` on my machine, or about 40% faster overall. Obviously this may vary, but I think it should be effective overall.

@ryantheleach I missed your comment but here's the answer :sweat_smile:

ryantheleach: Thanks.

PR #431, opened 2018-10-14
Refactor vector system to be cleaner

kenzierocks:
- Move Vector, etc. into `.math` package
- Drop many methods that will be auto-promoted anyways, eg. with
`divide(int)` and `divide(double)` the first is now gone.
- Take Block vectors into their own class hierarchy
- Make it clear throughout the API what takes blockvectors
- many more improvements


me4502:
Do you think it'd be worth replacing constructors with an `of` method or whatever so that we can return the ONE/ZERO/etc vectors when possible? Or even in the future implement some level of caching.

kenzierocks:
Fixed merge conflicts with `master`, may conflict with #432.

kenzierocks:
Squashed & ready for final review.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/adapter/BukkitImplAdapter.java
```diff
@@ -112,7 +112,7 @@
      * @param pos The position
      * @param nbtData The NBT Data
      */
-    void sendFakeNBT(Player player, Vector pos, CompoundTag nbtData);
+    void sendFakeNBT(Player player, Vector3 pos, CompoundTag nbtData);
```
me4502: Should this be BlockVector3? It'd require an adapter change too though.

kenzierocks: Yea, we need an adapter change anyways to change to `Vector3`. Will update.

PR #432, opened 2018-10-15
Added multi-pass fixing for MultiStageReorder, and a Fast-Mode equivalent

me4502:
Requires https://github.com/EngineHub/worldedit-adapters/pull/6


kenzierocks:
Perhaps we should drop `fixLighting` then? It's unused AFAIK.

me4502:
It may be worth making the notifyAndLightBlock just notify, and use a chunk-wide lighting fixer tbh. Also we should make sure that when we schedule a relight the game does it async, as I'm pretty sure async lighting exists.

DarkArc:
@kenzierocks maybe, could be nice from an API perspective to provide a fixLighting for an area, and also a fixAnyFastModeResults kind of thing as well though, so there may be some benefit to leave it

@me4502 sounds reasonable

me4502:
One thing I'm currently stuck on here, is how should we enable/disable this? Just always have it enabled? But then that kinda removes the need of having the multistagereorder extent. @DarkArc what have you found from using this kind of system? As this is based on your old one.

DarkArc:
I always had it enabled IIRC. The way to disable the system is to mak expose configuration for how much work it's able to do per tick, and then your mechanism for disabling batching is to just say, you have no containts per tick, so you can do everything (or something along those lines). This prevents you from maintaining two different systems.

Also the configuration needs to be available at the API level. Like there should bea way to get WE's queuing, but you should also be able to provide your own instance of a queue manager with your own timing specifications as a plugin author.

Also, if this doesn't make sense let me know I'm still half asleep 😅 This is kind of the visionI had for multipass when writing the initial logic though.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -149,6 +149,38 @@
...
+            return this.name;
+        }
+
+        public static ReorderMode getFromName(String name) {
+            for (ReorderMode mode : values()) {
+                if (mode.getName().equalsIgnoreCase(name)) {
+                    return mode;
+                }
+            }
+            return null;
```
kenzierocks: I would prefer to use an `Optional` instead of `null` here, I normally would expect this to throw.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -237,25 +271,76 @@ private Extent wrapExtent(Extent extent, EventBus eventBus, EditSessionEvent eve
...
+        }
+        if (reorderMode == ReorderMode.MULTI_STAGE && reorderExtent == null) {
+            throw new IllegalArgumentException("An EditSession without a reorder extent tried to use it for reordering!");
+        }
+        flushSession();
+
+        this.reorderMode = reorderMode;
+        switch (reorderMode) {
+            case MULTI_STAGE:
+                if (fastModeExtent != null) {
```
kenzierocks: I think this `if` block and the one in `case NONE:` really belong together outside the switch statement, in `if (reorderMode != FAST)`. Same with disabling the `reorderExtent`. It's not really the right use for `switch`.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -237,25 +271,76 @@ private Extent wrapExtent(Extent extent, EventBus eventBus, EditSessionEvent eve
...
+     * @param reorderMode The reorder mode
+     */
+    public void setReorderMode(ReorderMode reorderMode) {
+        if (reorderMode == ReorderMode.FAST && fastModeExtent == null) {
+            throw new IllegalArgumentException("An EditSession without a fast mode tried to use it for reordering!");
+        }
+        if (reorderMode == ReorderMode.MULTI_STAGE && reorderExtent == null) {
+            throw new IllegalArgumentException("An EditSession without a reorder extent tried to use it for reordering!");
+        }
+        flushSession();
```
kenzierocks: Surround this in `commitRequired()` like the other `flushSession()` call? Or maybe we should just make `flushSession()` do the check?

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -149,6 +149,38 @@
...
+     * MULTI_STAGE = Multi stage reorder, may not be great with mods.
+     * FAST = Use the fast mode. Good for mods.
+     * NONE = Place blocks without worrying about placement order.
+     */
+    public enum ReorderMode {
+        MULTI_STAGE("multi"),
+        FAST("fast"),
+        NONE("none");
+
+        private String name;
```
kenzierocks: Perhaps change this to `displayName`, likewise with `getFromName`? Enums already have a `name` and I think it's a little confusing to have a field with the same name.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/reorder/MultiStageReorder.java
```diff
@@ -94,58 +99,76 @@ public void setEnabled(boolean enabled) {
...
+        return enabled && stages.stream().anyMatch(stage -> stage.size() > 0);
+    }
+
+    /**
+     * Gets the stage priority of the block.
+     *
+     * @param block The block
+     * @return The priority
+     */
+    public int getPlacementPriority(BlockStateHolder block) {
```
kenzierocks: I think it might be nice to turn this into an enum, `PlacementPriority`:
```java
enum PlacementPriority {
  FIRST,
  LAST,
  LATE,
  FINAL
}
```

This should be a little more readable. Another option would be to just extract the numbers to fields, so there's no magic numbers around. It's hard to tell what `1` and `2` mean in `srcPriority == 1 || srcPriority == 2` below.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/reorder/MultiStageReorder.java
```diff
@@ -94,58 +99,76 @@ public void setEnabled(boolean enabled) {
...
-        } else if (Blocks.shouldPlaceFinal(block.getBlockType())) {
-            // Place signs, reed, etc even later
-            stage3.add(location, block);
-            return !existing.equalsFuzzy(block);
-        } else if (Blocks.shouldPlaceLast(existing.getBlockType())) {
+        BlockState existing = getBlock(location);
+        int priority = getPlacementPriority(block);
+        int srcPriority = getPlacementPriority(existing);
+
+        if (srcPriority == 1 || srcPriority == 2) {
```
kenzierocks: This is starting to feel like we should have a `RemovalPriority` as well -- this obviously doesn't mirror any placement priority, because it could be refactored to `<` or `>` in that case, but both `0` and `3` are excluded here.

Or maybe you did intend to include 3, and this should be `srcPriority > 0`.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/reorder/MultiStageReorder.java
```diff
@@ -201,11 +224,14 @@ public Operation resume(RunContext run) throws WorldEditException {
                 }
             }
 
-            stage1.clear();
-            stage2.clear();
-            stage3.clear();
+            if (blocks.isEmpty()) {
```
kenzierocks: Is this in preparation for multiple-tick operations? I don't see how this would ever be false, there's no break for the outer `while` loop above.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/reorder/MultiStageReorder.java
```diff
@@ -94,58 +99,76 @@ public void setEnabled(boolean enabled) {
...
                 blocks.add(pt);
                 blockTypes.put(pt, entry.getBlock());
             }
+        }
 
+        @Override
+        public Operation resume(RunContext run) throws WorldEditException {
             while (!blocks.isEmpty()) {
                 BlockVector3 current = blocks.iterator().next();
                 if (!blocks.contains(current)) {
```
kenzierocks: Can this ever be false? You create a new iterator, get the first element, and check if it's still in the same collection.

kenzierocks: Additionally, this isn't very useful, because if this is `false`, nothing will change -- `blocks` doesn't get modified.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/world/FastModeExtent.java
```diff
@@ -83,16 +87,36 @@ public void setEnabled(boolean enabled) {
...
+            }
+
+            return false;
         } else {
             return world.setBlock(location, block, true);
         }
     }
 
+    public boolean commitRequired() {
+        return !dirtyChunks.isEmpty() || !positions.isEmpty();
```
kenzierocks: Don't forget to include `enabled` in this check, like the others -- just in case.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/reorder/MultiStageReorder.java
me4502: Yeah I don't know what this was for

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
me4502: I don't think they make sense outside, as for example a FAST extent is perfectly capable of working without a reorder extent existing

worldedit-core/src/main/java/com/sk89q/worldedit/extent/reorder/MultiStageReorder.java
me4502: I've made some CLEAR levels of priority and made them get used - this fixes a few issues.

PR #433, opened 2018-10-21
Flush or disable buffers in tools

kenzierocks:
`/brush` and `/repl` implementations did not flush the queue previously. This appeared to work even with the reorder extent because most blocks weren't delayed, but with chunk batching it caused total breakage. This ensures that the buffers will be flushed, or disables them entirely for single-block changes.

Note: this PR also includes an `AutoCloseable`-implementing `EditSession` for ease of flushing. It's in a separate commit, so it can be easily reverted if deemed inappropriate.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/AreaPickaxe.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/AreaPickaxe.java
@@ -83,7 +83,7 @@ public class AreaPickaxe implements BlockTool {
         } catch (MaxChangedBlocksException e) {
             player.printError("Max blocks change limit reached.");
         } finally {
-            editSession.flushQueue();
+            editSession.flushSession();
             session.remember(editSession);
         }
 
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/AreaPickaxe.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/AreaPickaxe.java
@@ -62,29 +62,29 @@ public class AreaPickaxe implements BlockTool {
             return true;
         }
 
-        EditSession editSession = session.createEditSession(player);
-        editSession.getSurvivalExtent().setToolUse(config.superPickaxeManyDrop);
+        try (EditSession editSession = session.createEditSession(player)) {
+            editSession.getSurvivalExtent().setToolUse(config.superPickaxeManyDrop);
 
-        try {
-            for (int x = ox - range; x <= ox + range; ++x) {
-                for (int y = oy - range; y <= oy + range; ++y) {
-                    for (int z = oz - range; z <= oz + range; ++z) {
-                        Vector pos = new Vector(x, y, z);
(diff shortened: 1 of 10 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/AreaPickaxe.java
```diff
@@ -62,29 +62,29 @@ public boolean actPrimary(Platform server, LocalConfiguration config, Player pla
             return true;
         }
 
-        EditSession editSession = session.createEditSession(player);
-        editSession.getSurvivalExtent().setToolUse(config.superPickaxeManyDrop);
+        try (EditSession editSession = session.createEditSession(player)) {
```
me4502: It may be worth moving the current stuff here into the other try block just so we don't have two

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BlockReplacer.java
```diff
@@ -52,17 +52,19 @@ public boolean canUse(Actor player) {
     public boolean actPrimary(Platform server, LocalConfiguration config, Player player, LocalSession session, com.sk89q.worldedit.util.Location clicked) {
         BlockBag bag = session.getBlockBag(player);
 
-        EditSession editSession = session.createEditSession(player);
-
-        try {
-            Vector position = clicked.toVector();
-            editSession.setBlock(position, pattern.apply(position));
-        } catch (MaxChangedBlocksException ignored) {
+        try (EditSession editSession = session.createEditSession(player)) {
```
me4502: Same here

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/RecursivePickaxe.java
```diff
@@ -66,17 +66,17 @@ public boolean actPrimary(Platform server, LocalConfiguration config, Player pla
...
-
-        try {
-            recurse(server, editSession, world, clicked.toVector().toBlockVector(),
-                    clicked.toVector(), range, initialType, new HashSet<>());
-        } catch (MaxChangedBlocksException e) {
-            player.printError("Max blocks change limit reached.");
-        } finally {
-            editSession.flushQueue();
-            session.remember(editSession);
+        try (EditSession editSession = session.createEditSession(player)) {
```
me4502: Same here

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/SinglePickaxe.java
```diff
@@ -49,15 +49,11 @@ public boolean actPrimary(Platform server, LocalConfiguration config, Player pla
             return true;
         }
 
-        EditSession editSession = session.createEditSession(player);
-        editSession.getSurvivalExtent().setToolUse(config.superPickaxeDrop);
-
-        try {
+        try (EditSession editSession = session.createEditSession(player)) {
```
me4502: Same here

PR #434, opened 2018-10-31
Flush / unbuffer more tools

kenzierocks:
#433 missed some tools, not sure why. I checked every use of `session#createEditSession` in this one, there might be some edge cases still not hit.


me4502:
Thanks

PR #435, opened 2018-10-31
Add tracing for unflushed EditSessions

kenzierocks:
With the new chunk batching, flushing is more important, as evidenced by #433 and #434. This adds a debug helper to allow developers to find leaks in tools that might otherwise appear to work fine.

Note: we may want to _enable_ this by default, as although it adds time to construction and slows down GC by a tiny amount due to the finalizer, it's probably not a huge deal because of how little EditSessions are created (only a few hundred even on busy servers).

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -234,6 +234,16 @@ public class EditSession implements Extent, AutoCloseable {
         return event.getExtent();
     }
 
+    // pkg private for TracedEditSession only
+
+    ChunkBatchingExtent getChunkBatchingExtent() {
+        return chunkBatchingExtent;
+    }
+
+    MultiStageReorder getReorderExtent() {
+        return reorderExtent;
+    }
+
     /**
      * Turns on specific features for a normal WorldEdit session, such as
      * {@link #enableQueue() queuing} and {@link #setBatchingChunks(boolean)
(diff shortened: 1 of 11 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -234,6 +234,16 @@ private Extent wrapExtent(Extent extent, EventBus eventBus, EditSessionEvent eve
         return event.getExtent();
     }
 
+    // pkg private for TracedEditSession only
+
+    ChunkBatchingExtent getChunkBatchingExtent() {
```
me4502: Would it be better to just have a `commitRequired()` method here? That way we could have something semi-generic that also allows for custom extents.

kenzierocks: Sure, I'll do that.

kenzierocks: Sure, I'll do that.

worldedit-core/src/main/java/com/sk89q/worldedit/util/PropertiesConfiguration.java
```diff
@@ -75,6 +75,7 @@ public void load() {
         loadExtra();
 
         profile = getBool("profile", profile);
+        traceUnflushedSessions = getBool("traceUnflushedSessions", traceUnflushedSessions);
```
me4502: Other options appear to be in the lowercase - separated format

PR #437, opened 2018-12-05
Add check for unchanged materials in BlockBagExtent

JOO200:
I have written my own BlockBag and don't want to handle unchanged materials.
Maybe it is possible to change the behavior of the BlockBagExtent :)

Current behavoir:
Set new Block: Check for the necessary item in the inventory (fetchBlock). Store the old block after getting the new one.
That's okay, better take the placed item first. But if you want to change a block with the same material, you have two unnecessary operations.


me4502:
Thanks

PR #438, opened 2018-12-23
Modularise Patterns & Masks

me4502:


worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledItemRegistry.java
```diff
@@ -19,22 +19,10 @@
...
-import com.sk89q.worldedit.world.item.ItemType;
-import com.sk89q.worldedit.world.item.ItemTypes;
-
-import javax.annotation.Nullable;
-
 /**
  * A item registry that uses {@link BundledItemRegistry} to serve information
  * about items.
  */
 public class BundledItemRegistry implements ItemRegistry {
```
me4502: TODO Probably remove this class - is it even used?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/BiomeMaskParser.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public Mask parseFromInput(String input, ParserContext context) throws InputParseException {
+        if (!input.startsWith("$")) {
+            return null;
+        }
+
+        Set<BaseBiome> biomes = new HashSet<>();
+        String[] biomesList = input.substring(1).split(",");
+        BiomeRegistry biomeRegistry = worldEdit.getPlatformManager().queryCapability(Capability.GAME_HOOKS).getRegistries().getBiomeRegistry();
```
kenzierocks: This makes me feel like we should have a `BiomeTypes` like we have `ItemTypes` above. Possibly another PR?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/BiomeMaskParser.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public Mask parseFromInput(String input, ParserContext context) throws InputParseException {
+        if (!input.startsWith("$")) {
+            return null;
+        }
+
+        Set<BaseBiome> biomes = new HashSet<>();
+        String[] biomesList = input.substring(1).split(",");
```
kenzierocks: Might be better to split using `Splitter` and use just an iterator, rather than have an intermediate array.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/BlockCategoryMaskParser.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class BlockCategoryMaskParser extends InputParser<Mask> {
+
+    public BlockCategoryMaskParser(WorldEdit worldEdit) {
+        super(worldEdit);
+    }
+
+    @Override
+    public Mask parseFromInput(String input, ParserContext context) throws InputParseException {
+        if (input.startsWith("##")) {
```
kenzierocks: Use `!input.startsWith` to align with the parser above?

worldedit-core/src/main/java/com/sk89q/worldedit/internal/registry/InputParser.java
```diff
@@ -33,10 +36,18 @@
...
 
     public abstract E parseFromInput(String input, ParserContext context) throws InputParseException;
 
+    /**
+     * Gets a list of suggestions of input to this parser.
+     *
+     * @return a list of suggestions
+     */
+    public List<String> getSuggestions() {
+        return Lists.newArrayList();
```
kenzierocks: I would prefer a use of `Collections.emptyList()` here, there's no need for this array list.

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/ItemRegistry.java
```diff
@@ -19,19 +19,6 @@
 
 package com.sk89q.worldedit.world.registry;
 
-import com.sk89q.worldedit.blocks.BaseItem;
-
-import javax.annotation.Nullable;
-
 public interface ItemRegistry {
```
kenzierocks: Is there even a point to this interface now? It's totally empty.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/BiomeMaskParser.java
me4502: Internally biomes haven't been converted to a registry system in 1.13, I was waiting on that to see how it's done. Not sure if it's done in 1.14 yet.

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/ItemRegistry.java
me4502: Just fixed it up, it's now used like it should be

PR #439, opened 2018-12-24
Fix Metrics injection for Sponge distributive

gamerforEA:
[crash-2018-12-24_16.03.01-server.txt](https://github.com/sk89q/WorldEdit/files/2707208/crash-2018-12-24_16.03.01-server.txt)

Server crashed at start, because Google Guice can't find implementation for Metrics interface. I replaced it to Metrics2 class.

### DIFF
--- a/worldedit-sponge/build.gradle
+++ b/worldedit-sponge/build.gradle
@@ -17,7 +17,7 @@ repositories {
 
 dependencies {
     compile project(':worldedit-core')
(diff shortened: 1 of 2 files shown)


me4502:
Thanks,

Worldedit 7 is for 1.13, so there’s no guarantees at all that it’ll work, however this is something I would have to do anyway when sponge for 1.13 does eventually arrive.

PR #440, opened 2018-12-24
Fix BlockType and ItemType registration for Sponge distributive

gamerforEA:
[2018-12-24-6.log](https://github.com/sk89q/WorldEdit/files/2707671/2018-12-24-6.log)
Plugin initialization failed due to double registration from BlockTypes/ItemTypes classes and Sponge registry.

P.S. I know, that WorldEdit 7 is for 1.13, not 1.12.2, but I think that this problem will be relevant for future versions too.


me4502:
Thanks

PR #443, opened 2018-12-28
session.getRegionSelector(BukkitAdapter.adapt(world)).getPrimaryPosition(); but where is .getSecondaryPosition()??

FlashLight34:
https://github.com/sk89q/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/regions/RegionSelector.java#L104


me4502:
Do not open pull requests to ask questions.

The answer is that this is a generic type for all shapes. Not all shapes have a "secondary" position. If you know that it's cuboid, cast it to a CuboidRegionSelector.

PR #444, opened 2019-01-01
Update maze.js

orthoplex64:
My friend wanted to use this script but it was broken; seems to be working after these changes.
Changes are based on the commit that updated all the other scripts.


me4502:
Thanks

PR #445, opened 2019-01-21
Make biome changes undoable

Brokkonaut:


worldedit-core/src/main/java/com/sk89q/worldedit/extent/ChangeSetExtent.java
```diff
@@ -65,6 +68,13 @@ public ChangeSetExtent(Extent extent, ChangeSet changeSet) {
         return super.setBlock(location, block);
     }
 
+    @Override
+    public boolean setBiome(BlockVector2 position, BaseBiome biome) {
+        BaseBiome previous = getBiome(position);
+        changeSet.add(new BiomeChange(position, previous, new BaseBiome(biome)));
```
me4502: Is there a reason you’re wrapping the current biome?

Brokkonaut: BaseBiome is not immutable, so when it comes from external sources it might be modified after calling setBiome. So I thought it would be safer to create a internal copy.

me4502: Oh fair

me4502:
Thanks

PR #446, opened 2019-02-11
Added offset to ClipboardPattern.

wizjany:
Takes input like '//set #clipboard@-1,0,1' which shifts the pattern over.
It also now extends RepeatingExtentPattern, which was previously unused.

In action: https://gfycat.com/EnormousWeirdInvisiblerail

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/ClipboardPatternParser.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/ClipboardPatternParser.java
@@ -28,31 +28,53 @@ import com.sk89q.worldedit.extension.input.ParserContext;
 import com.sk89q.worldedit.extent.clipboard.Clipboard;
 import com.sk89q.worldedit.function.pattern.ClipboardPattern;
 import com.sk89q.worldedit.function.pattern.Pattern;
-import com.sk89q.worldedit.internal.registry.SimpleInputParser;
+import com.sk89q.worldedit.internal.registry.InputParser;
+import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.session.ClipboardHolder;
 
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/ClipboardPatternParser.java
```diff
@@ -28,31 +28,53 @@
...
-    public Pattern parseFromSimpleInput(String input, ParserContext context) throws InputParseException {
+    public Pattern parseFromInput(String input, ParserContext context) throws InputParseException {
+        if (!input.startsWith("#clipboard") && !input.startsWith("#copy")) {
+            return null;
+        }
         LocalSession session = context.requireSession();
 
+        int offsetPart;
+        BlockVector3 offset = BlockVector3.ZERO;
+        if ((offsetPart = input.indexOf('@')) >= 0) {
```
TomyLobo: I'd just use split here instead of indexOf+substring.
Makes for more readable code.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/ClipboardPatternParser.java
```diff
@@ -28,31 +28,53 @@
...
+            if (input.length() <= offsetPart + 1) {
```
TomyLobo: Don't know if you noticed, but this check only triggers with input.equals("#copy@") and input.equals("#clipboard@"), both of which would trigger the next error condition as well.

TomyLobo:
Look good to me 👍
I hope you don't mind that I defer merging to someone who actually knows htf the new framework works :)

wizjany:
yea ofc, I will probably merge myself. Just PR for feedback since I'm relatively new to 1.13 stuff, and I've been tarnished by C# patterns and syntax.

PR #447, opened 2019-02-12
Add some cool new state things.

wizjany:
Masks based on block state:
`^[state=value,...]`
in action: https://gfycat.com/kindheartedcolossalfowl

Patterns that change type while retaining (valid) states:
`//set ^type`
in action: https://gfycat.com/ExcitableShorttermCobra (note `^^` is now `^`)
or change states while retaining original type:
`//set ^[state=value,...]`
or both:
`//set ^type[state=value,...]`
in action: https://gfycat.com/WindyEvenAmberpenshell (syntax changed)

Drain waterlogged blocks:
`//drain -w <radius>`
in action: https://gfycat.com/forkedsilkyerne

Use all states in block category:
`//set ##*wool` (note the *. without it, only default states are used)

Combining patterns with chance:
https://i.imgur.com/rHMlJXp.png
(previously, only single blocks could be used with chances)

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/BlockCategoryPatternParser.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/BlockCategoryPatternParser.java
@@ -28,9 +28,11 @@ import com.sk89q.worldedit.function.pattern.RandomPattern;
 import com.sk89q.worldedit.internal.registry.InputParser;
 import com.sk89q.worldedit.world.block.BlockCategories;
 import com.sk89q.worldedit.world.block.BlockCategory;
+import com.sk89q.worldedit.world.block.BlockState;
 import com.sk89q.worldedit.world.block.BlockType;
 
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 public class BlockCategoryPatternParser extends InputParser<Pattern> {
@@ -46,17 +48,34 @@ public class BlockCategoryPatternParser extends InputParser<Pattern> {
 
     @Override
     public Pattern parseFromInput(String input, ParserContext context) throws InputParseException {
-        if(!input.startsWith("##")) {
+        if (!input.startsWith("##")) {
             return null;
         }
-        BlockCategory category = BlockCategories.get(input.substring(2).toLowerCase());
+        String tag = input.substring(2).toLowerCase();
+        boolean anyState = false;
+        if (tag.startsWith("*")) {
+            tag = tag.substring(1);
+            anyState = true;
+        }
+
+        BlockCategory category = BlockCategories.get(tag);
         if (category == null) {
-            throw new InputParseException("Unknown block tag: " + input.substring(2));
+            throw new InputParseException("Unknown block tag: " + tag);
         }
         RandomPattern randomPattern = new RandomPattern();
 
-        for (BlockType blockType : category.getAll()) {
-            randomPattern.add(new BlockPattern(blockType.getDefaultState()), 1.0 / category.getAll().size());
+        Set<BlockType> blocks = category.getAll();
+        if (blocks.isEmpty()) {
+            throw new InputParseException("Block tag " + category.getId() + " had no blocks!");
+        }
+
+        if (anyState) {
+            blocks.stream().flatMap(blockType -> blockType.getAllStates().stream()).forEach(state ->
+                randomPattern.add(new BlockPattern(state), 1.0));
+        } else {
+            for (BlockType blockType : blocks) {
+                randomPattern.add(new BlockPattern(blockType.getDefaultState()), 1.0);
+            }
         }
 
         return randomPattern;
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/BlockStateMask.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/BlockStateMask.java
@@ -1,0 +1,70 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.function.mask;
+
+import com.google.common.collect.Maps;
+import com.sk89q.worldedit.blocks.Blocks;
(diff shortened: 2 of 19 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/BlockStateMask.java
```diff
@@ -0,0 +1,52 @@
...
+    private final Map<String, String> states;
+
+    public BlockStateMask(Extent extent, Map<String, String> states) {
+        super(extent);
+        this.states = states;
+    }
+
+    @Override
+    public boolean test(BlockVector3 vector) {
+        Map<String, Object> existingStates = getExtent().getBlock(vector).getStates().entrySet().stream()
```
kenzierocks: I think this could use caching, it's going to be called often. Or perhaps there's another performant way to do it I can't think of.

worldedit-core/src/main/java/com/sk89q/worldedit/function/pattern/ExtentBufferedCompositePattern.java
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * <p>Note that all patterns passed which are ExtentPatterns should use the same extent as the one
+     * passed to this constructor, or block changes may not be realized by those patterns.</p>
+     *
+     * @param extent the extent to buffer changes to
+     * @param patterns the patterns to apply, in order
+     */
+    public ExtentBufferedCompositePattern(Extent extent, Pattern... patterns) {
+        super(extent);
+        checkArgument(patterns.length == 0, "patterns cannot be empty");
```
kenzierocks: This is backwards, you want the expected condition to be here.

worldedit-core/src/main/java/com/sk89q/worldedit/function/pattern/ExtentBufferedCompositePattern.java
```diff
@@ -0,0 +1,66 @@
...
+        this.patterns = patterns;
+    }
+
+    @Override
+    public BaseBlock apply(BlockVector3 position) {
+        BaseBlock lastBlock = null;
+        for (Pattern pattern : patterns) {
+            lastBlock = pattern.apply(position);
+            try {
+                getExtent().setBlock(position, lastBlock);
```
kenzierocks: This seems odd. Why are you replacing the block over and over?

worldedit-core/src/main/java/com/sk89q/worldedit/function/pattern/TypeApplyingPattern.java
```diff
@@ -0,0 +1,51 @@
...
+        super(extent);
+        this.blockState = blockState;
+    }
+
+    @Override
+    public BaseBlock apply(BlockVector3 position) {
+        BlockState oldBlock = getExtent().getBlock(position);
+        BlockState newBlock = blockState;
+        for (Entry<Property<?>, Object> entry : oldBlock.getStates().entrySet()) {
+            //noinspection unchecked
```
kenzierocks: I typically like to use the annotation to cut down on javac warnings too.

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/BlockStateMask.java
me4502: Realistically how often will it be called on a single position?

Perhaps do a "last position cache"

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockCategories.java
```diff
@@ -32,8 +32,8 @@
     public static final BlockCategory BIRCH_LOGS = register("minecraft:birch_logs");
     public static final BlockCategory BUTTONS = register("minecraft:buttons");
     public static final BlockCategory CARPETS = register("minecraft:carpets");
-    public static final BlockCategory CORAL = register("minecraft:coral");
-    public static final BlockCategory CORAL_PLANTS = register("minecraft:coral_plants");
+    public static final BlockCategory CORALS = register("minecraft:corals");
```
me4502: Was this generated using MCUtils? It's very odd that it has changed. If this has changed across versions we may need to rethink how it works :(

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/BlockCategoryPatternParser.java
```diff
@@ -46,17 +48,36 @@ public BlockCategoryPatternParser(WorldEdit worldEdit) {
...
-        for (BlockType blockType : category.getAll()) {
-            randomPattern.add(new BlockPattern(blockType.getDefaultState()), 1.0 / category.getAll().size());
+        Set<BlockType> blocks = category.getAll();
+        if (blocks.isEmpty()) {
+            throw new InputParseException("Block tag " + category.getId() + " had no blocks!");
+        }
+
+        if (anyState) {
+            for (BlockState blockState : blocks.stream()
+                    .flatMap(blockType -> blockType.getAllStates().stream()).collect(Collectors.toList())) {
```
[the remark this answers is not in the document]
wizjany: because LINQ doesn't have foreach :P

worldedit-core/src/main/java/com/sk89q/worldedit/function/pattern/TypeApplyingPattern.java
wizjany: intellij did this because it's inline. moved it to method annotation instead.

worldedit-core/src/main/java/com/sk89q/worldedit/function/pattern/ExtentBufferedCompositePattern.java
wizjany: because the next (Extent)Pattern needs the newly changed block to apply to. since we don't want to change the block in the world, we change it in the buffer. so the buffer changes over and over until we have our final type/states/etc, and then return that to the caller.

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockCategories.java
wizjany: done by hand since ##coral was returning 0 blocks and i checked the wiki. it's very possible that other things have/will change. (I was just using coral for testing since it's conveniently waterlogged by default.) definitely agree it would be nice to do this some other way (can we get all valid tags from the server on platform init?)

worldedit-core/src/main/java/com/sk89q/worldedit/function/pattern/TypeApplyingPattern.java
wizjany: oh, i could do the cast separately too (to a new local var). not sure which is more idiomatic.

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/BlockStateMask.java
wizjany: pretty much never on the same position, since masks are generally used when iterating over regions.
is there a better way to do this? converting properties to strings is unfortunately kinda necessary since two different blocktypes have different property objects for the same name. we could cache properties per blocktype i suppose?

wizjany: so instead of checking a `Map<String, String>` each time, we could have a `Map<BlockType, Map<Property<>, Object>`, which allows us to do ```
BlockState block = getExtent().getBlock(vector);
cache.computeIfAbsent(block.getType(), ...).entrySet().stream().allMatch(entry -> block.getState(entry.getKey()) == entry.getValue());
```
that means only computing which properties to check once per block type, which will save a lot.

PR #448, opened 2019-02-15
WIP Multi region CUI support

wizjany:
This PR is mostly just for brainstorming/gathering ideas on API and usages.

So, a while ago mutli-region support was added to the WE CUI
https://github.com/Mumfrey/WorldEditCUI/issues/16
since blood wanted to display multiple GP claims at once to the client.

I've drafted some preliminary API and one (very hacky POC) usage:
https://gfycat.com/CautiousFrailAmericancrocodile

This could also be used in WG to show regions (like GP does now) and so on.

PS - currently this targets 1.12 since the client mod might not be updated for a bit.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/Brush.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/Brush.java
@@ -23,6 +23,8 @@ import com.sk89q.worldedit.EditSession;
 import com.sk89q.worldedit.MaxChangedBlocksException;
 import com.sk89q.worldedit.Vector;
 import com.sk89q.worldedit.function.pattern.Pattern;
+import com.sk89q.worldedit.regions.CuboidRegion;
+import com.sk89q.worldedit.regions.Region;
 
 /**
  * A brush is a long-range build tool.
@@ -40,4 +42,15 @@ public interface Brush {
      */
     public void build(EditSession editSession, Vector position, Pattern pattern, double size) throws MaxChangedBlocksException;
 
+    /**
+     * Get the region that represents the area that would be affected if called at the given position.
+     *
+     * @param session the editsession to build in
+     * @param position the brush target
+     * @param size the brush size
+     * @return the bounds of the affected area
+     */
+    default public Region getBounds(EditSession session, Vector position, double size) {
+        return CuboidRegion.fromCenter(position, (int) Math.ceil(size));
+    }
 }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/multi/ConvexPolyhedralCUIMultiRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/multi/ConvexPolyhedralCUIMultiRegion.java
@@ -1,0 +1,68 @@
+package com.sk89q.worldedit.internal.cui.multi;
+
+import com.sk89q.worldedit.BlockVector2D;
+import com.sk89q.worldedit.LocalSession;
+import com.sk89q.worldedit.Vector;
+import com.sk89q.worldedit.extension.platform.Actor;
+import com.sk89q.worldedit.internal.cui.SelectionMinMaxEvent;
+import com.sk89q.worldedit.internal.cui.SelectionPoint2DEvent;
+import com.sk89q.worldedit.internal.cui.SelectionPointEvent;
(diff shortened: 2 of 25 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/Brush.java
```diff
@@ -40,4 +42,15 @@
...
 
+    /**
+     * Get the region that represents the area that would be affected if called at the given position.
+     *
+     * @param session the editsession to build in
+     * @param position the brush target
+     * @param size the brush size
+     * @return the bounds of the affected area
+     */
+    default public Region getBounds(EditSession session, Vector position, double size) {
```
me4502: 1.12 doesn't guarantee Java 8

wizjany: this isn't really meant to go to 1.12. just using it as a test bed.
also the idea is that third-party brushes will stay compatible, not 100% sure how that works though.

kenzierocks: Unrelated -- for an interface, you don't need `public`.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/multi/ConvexPolyhedralCUIMultiRegion.java
```diff
@@ -0,0 +1,68 @@
...
+        this.region = region;
+    }
+
+    @Override
+    public void describeCUI(LocalSession session, Actor player) {
+        super.describeCUI(session, player);
+        Collection<Vector> vertices = region.getVertices();
+        Collection<Triangle> triangles = region.getTriangles();
+
+        Map<Vector, Integer> vertexIds = new HashMap<Vector, Integer>(vertices.size());
```
kenzierocks: No need for the second set of generics -- just use `<>`.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/multi/ConvexPolyhedralCUIMultiRegion.java
```diff
@@ -0,0 +1,68 @@
...
+        int lastVertexId = -1;
+        for (Vector vertex : vertices) {
+            vertexIds.put(vertex, ++lastVertexId);
```
kenzierocks: Why the `-1` and prefixed `++` rather than a more standard `0` and suffixed `++`?

worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/multi/MultiRegionClearEvent.java
```diff
@@ -0,0 +1,21 @@
...
+public class MultiRegionClearEvent implements CUIMultiEvent {
+    private final String regionID;
+
+    public MultiRegionClearEvent(String regionID) {
+        this.regionID = regionID;
+    }
+
+    @Override
+    public String getTypeId() {
+        return "+s";
```
kenzierocks: Is there a special meaning to `+s`? I would rather have something like `+clear`.

kenzierocks: On review of other files, it looks like a copy-paste mistake from `Shape`?

wizjany: the protocol was decided without our input :P "s" once meant selection. so sending "s|cuboid" meant "start a cuboid selection". the "+" part is for the multiregion protocol, so it doesn't clash with the normal WE selection CUI.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/multi/ConvexPolyhedralCUIMultiRegion.java
wizjany: all of these are copied verbatim from TomyLobo's old code, with the exception of the Wrapped event constructor. will definitely update style and so on in another pass.

wizjany:
closing til we have a working CUI and a better idea of how the puzzle pieces should fit.

PR #449, opened 2019-02-16
Refactor registries, and added a registry for biomes

me4502:
This removes a lot of adapter code from Bukkit and Sponge, and also just overly simplifies how biomes work in WorldEdit.


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
```diff
@@ -304,6 +306,27 @@ public static GameMode adapt(org.bukkit.GameMode gameMode) {
...
+    }
+
+    public static Biome adapt(BiomeType biomeType) {
+        if (!biomeType.getId().startsWith("minecraft:")) {
+            throw new IllegalArgumentException("Bukkit only supports vanilla biomes");
+        }
+        try {
+            return Biome.valueOf(biomeType.getId().substring(10).toUpperCase());
+        } catch (IllegalArgumentException e) {
+            return null;
```
kenzierocks: This is weird -- either we should return `null` in all error cases, or throw an error. You do the former here, and the latter above.

me4502: This is the only one using the Enum#valueOf method, which throws an exception. I don't believe the others do?

kenzierocks: Oh, weird. Looks like it'd be consistent then.

worldedit-core/src/main/java/com/sk89q/worldedit/LocalConfiguration.java
```diff
@@ -138,6 +77,73 @@
...
+                BlockTypes.DARK_OAK_BUTTON,
+                BlockTypes.JUNGLE_BUTTON,
+                BlockTypes.OAK_BUTTON,
+                BlockTypes.SPRUCE_BUTTON,
+                BlockTypes.CACTUS,
+                BlockTypes.SUGAR_CANE,
+                // ores and stuff
+                BlockTypes.BEDROCK
+        );
+        return blockTypes.stream().filter(type -> !Objects.isNull(type)).map(BlockType::getId).toArray(String[]::new);
```
kenzierocks: 95% sure there's a `isNotNull`.

me4502:
So realistically yes.

I forgot weather existed, so I’ll do that. As for fluids, we can’t populate it from Spigot as they don’t have a system for it. I assume once it actually gets used within the game they’ll add one.

As for the categories... getting that to generate in MCUtils is fairly complex as they’re loaded from JSON files, and it’s not a clean process.

wizjany:
> As for the categories... getting that to generate in MCUtils is fairly complex as they’re loaded from JSON files, and it’s not a clean process.

Ah yea, forgot about data packs entirely.

PR #451, opened 2019-03-03
Add better control over expression timeouts.

wizjany:
- //timeout command can be used to change player's current timeout.
- Config now also has a max timeout, can be bypassed with permission
- Timeout of < -1 will let expressions run indefinitely.
- Said expressions won't run on a separate thread, slightly reducing the overhead from context switching. For large //gen commands, for example, this can actually increase speed.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -81,6 +81,7 @@ import com.sk89q.worldedit.history.changeset.BlockOptimizedHistory;
 import com.sk89q.worldedit.history.changeset.ChangeSet;
 import com.sk89q.worldedit.internal.expression.Expression;
 import com.sk89q.worldedit.internal.expression.ExpressionException;
+import com.sk89q.worldedit.internal.expression.runtime.ExpressionTimeoutException;
 import com.sk89q.worldedit.internal.expression.runtime.RValue;
 import com.sk89q.worldedit.math.BlockVector2;
 import com.sk89q.worldedit.math.BlockVector3;
@@ -1879,9 +1880,43 @@ public class EditSession implements Extent, AutoCloseable {
         return count.getDistribution();
     }
 
-    public int makeShape(final Region region, final Vector3 zero, final Vector3 unit, final Pattern pattern, final String expressionString, final boolean hollow) throws ExpressionException, MaxChangedBlocksException {
+    /**
+     * Generate a shape for the given expression.
+     *
+     * @param region the region to generate the shape in
+     * @param zero the coordinate origin for x/y/z variables
+     * @param unit the scale of the x/y/z/ variables
+     * @param pattern the default material to make the shape from
+     * @param expressionString the expression defining the shape
+     * @param hollow whether the shape should be hollow
+     * @return number of blocks changed
+     * @throws ExpressionException
+     * @throws MaxChangedBlocksException
+     */
+    public int makeShape(final Region region, final Vector3 zero, final Vector3 unit,
+                         final Pattern pattern, final String expressionString, final boolean hollow)
+            throws ExpressionException, MaxChangedBlocksException {
+        return makeShape(region, zero, unit, pattern, expressionString, hollow, WorldEdit.getInstance().getConfiguration().calculationTimeout);
+    }
+
+    /**
+     * Generate a shape for the given expression.
+     *
+     * @param region the region to generate the shape in
+     * @param zero the coordinate origin for x/y/z variables
+     * @param unit the scale of the x/y/z/ variables
+     * @param pattern the default material to make the shape from
+     * @param expressionString the expression defining the shape
+     * @param hollow whether the shape should be hollow
+     * @param timeout the time, in milliseconds, to wait for each expression evaluation before halting it. -1 to disable
+     * @return number of blocks changed
+     * @throws ExpressionException
+     * @throws MaxChangedBlocksException
+     */
+    public int makeShape(final Region region, final Vector3 zero, final Vector3 unit,
+                        final Pattern pattern, final String expressionString, final boolean hollow, final int timeout)
+            throws ExpressionException, MaxChangedBlocksException {
         final Expression expression = Expression.compile(expressionString, "x", "y", "z", "type", "data");
-        expression.optimize();
 
         final RValue typeVariable = expression.getVariable("type", false);
         final RValue dataVariable = expression.getVariable("data", false);
@@ -1889,6 +1924,7 @@ public class EditSession implements Extent, AutoCloseable {
         final WorldEditExpressionEnvironment environment = new WorldEditExpressionEnvironment(this, unit, zero);
         expression.setEnvironment(environment);
 
+        final int[] timedOut = {0};
         final ArbitraryShape shape = new ArbitraryShape(region) {
             @Override
             protected BaseBlock getMaterial(int x, int y, int z, BaseBlock defaultMaterial) {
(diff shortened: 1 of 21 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/LocalConfiguration.java
```diff
@@ -67,6 +67,7 @@
     public int navigationWandMaxDistance = 50;
     public int scriptTimeout = 3000;
     public int calculationTimeout = 100;
+    public int maxCalculationTimeout = 300;
```
kenzierocks: May want to verify post-config load that this is `>= calculationTimeout`, to prevent weird config setups from existing.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/Expression.java
```diff
@@ -116,7 +116,15 @@ private Expression(List<Token> tokens, String... variableNames) throws Expressio
         root = Parser.parse(tokens, this);
     }
 
+    public double evaluate() throws EvaluationException {
```
kenzierocks: No point to this if you keep the varargs.

worldedit-core/src/test/java/com/sk89q/worldedit/internal/expression/ExpressionTest.java
```diff
@@ -64,7 +64,7 @@ public void testEvaluate() throws ExpressionException {
         assertEquals(atan2(3, 4), simpleEval("atan2(3, 4)"), 0);
 
         // check variables
-        assertEquals(8, compile("foo+bar", "foo", "bar").evaluate(5, 3), 0);
+        assertEquals(8, compile("foo+bar", "foo", "bar").evaluate(5D, 3D), 0);
```
kenzierocks: There's no need to change these, with the vararg / timeout separation in place.

wizjany: yea, as mentioned in the commit message, I decided to leave the change for clarity.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1879,16 +1880,51 @@ public int makeForest(BlockVector3 basePosition, int size, double density, TreeG
...
+     * @param timeout the time, in milliseconds, to wait for each expression evaluation before halting it. -1 to disable
+     * @return number of blocks changed
+     * @throws ExpressionException
+     * @throws MaxChangedBlocksException
+     */
+    public int makeShape(final Region region, final Vector3 zero, final Vector3 unit,
+                        final Pattern pattern, final String expressionString, final boolean hollow, final int timeout)
+            throws ExpressionException, MaxChangedBlocksException {
         final Expression expression = Expression.compile(expressionString, "x", "y", "z", "type", "data");
-        expression.optimize();
```
TomyLobo: why'd you throw that out?

TomyLobo:
I see you are storing the player-specific timeout in all kinds of objects.
Do you do that to make a timeout setting persist even after the user changed it, for instance in a mask?
If yes, won't that be counter-intuitive?
If not, aren't you adding complexity to the API for no reason?

worldedit-core/src/main/java/com/sk89q/worldedit/session/SessionManager.java
```diff
@@ -192,6 +183,22 @@ public synchronized LocalSession get(SessionOwner owner) {
...
 
+    private boolean shouldBoundLimit(int defaultLimit, int maxLimit, int currentLimit) {
+        // If the maximum isn't infinite, ensure the default isn't larger than it.
+        if (maxLimit > -1) {
+            if (defaultLimit < 0) {
+                if (currentLimit < 0 || currentLimit > maxLimit) {
+                    return true;
+                }
+            } else {
+                if (currentLimit == -1 || currentLimit > maxLimit) {
```
TomyLobo: Something is messed up here... one limit is tested == -1, the other limit is tested < 0, yet another is tested > -1.
Also, defaultLimit is never checked for being larger than the maximum, so the comment is wrong.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/Expression.java
```diff
@@ -127,34 +131,44 @@ public double evaluate(double... values) throws EvaluationException {
...
             Thread.currentThread().interrupt();
             throw new RuntimeException(e);
+        } catch (TimeoutException e) {
+            result.cancel(true);
+            throw new ExpressionTimeoutException("Calculations exceeded time limit.");
         } catch (ExecutionException e) {
             Throwable cause = e.getCause();
             if (cause instanceof ReturnException) {
-                return ((ReturnException) cause).getValue();
+                throw (ReturnException) cause;
```
TomyLobo: There's no need to cast before throwing.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/Expression.java
```diff
@@ -127,34 +131,44 @@ public double evaluate(double... values) throws EvaluationException {
...
             }
             if (cause instanceof RuntimeException) {
                 throw (RuntimeException) cause;
             }
             throw new RuntimeException(cause);
```
TomyLobo: Can yo add a case for EvaluationException, so we don't break Exception checking as much by introducing the Future?
```
             if (cause instanceof EvaluationException) {
                throw (EvaluationException) cause;
            }
```

TomyLobo:
to summarize my comments:
1. (not introduced by your changes) shouldBoundLimit's check for negative values is inconsistent
2. (not introduced by your changes) Please add a case for EvaluationException to evaluateRootTimed
3. Please re-add the calls to `expression.optimize();` or state why they can no longer be there (I assume that's a leftover debug change)

(Yeah I should have used the review feature, sorry for the mail spam)

worldedit-core/src/main/java/com/sk89q/worldedit/session/SessionManager.java
wizjany: config loading checks `Math.max(-1, config.getInt(...` so if it's less than zero, it must be -1. but yea that relies on the config's check. we can make it explicit here. default CAN be larger than max, IF the player has override perms. we actually are bounding current to max here. this method is only called if the player doesn't have override perms.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
wizjany: good catch.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/Expression.java
wizjany: will do.

wizjany:
> 
> 
> I see you are storing the player-specific timeout in all kinds of objects.
> Do you do that to make a timeout setting persist even after the user changed it, for instance in a mask?
> If yes, won't that be counter-intuitive?
> If not, aren't you adding complexity to the API for no reason?
> Can't you just read the timeout value from the session somehow?

I just noticed there's some inconsistency here. Yes, the idea is that if you set an expression mask and later change the timeout, the new timeout will be used for the old mask. This allows you to set an expression mask, and if it takes too long, change the timeout without setting your mask again. I think this is more intuitive than a user doing //timeout and the timeout on their old mask not changing.

I originally passed an IntSupplier into the Expression itself, which would be called at expression runtime to get the current value of the timeout.
Right now, the operation factory brush holds the session so it always uses the current timeout, but the expression mask doesn't. That said, storing the session in a lot of places is kind of ugly. But even if I pass an IntSupplier into that OperationFactoryBrush/ExpressionMask (where a LocalSession isn't available), I'd have to close the session into the IntSupplier anyway.

wizjany:
btw @TomyLobo unrelated to this PR and for my personal curiosity, did you have some use cases for the [g]megabuf functions? I've been playing around with expressions recently and trying to think of how they might be used.

TomyLobo:
It's been a while since I last did anything with minecraft, and especially WorldEdit.
The names for the functions are inspired by (read: shamelessly stolen from) [AVS](https://en.wikipedia.org/wiki/Advanced_Visualization_Studio)'s evallib scripting engine :)

They can be useful for generating complex randomized structures where you need some kind of storage for the random values, for instance.

I just now, looking at the code, remember that I used this to create [voronoi cells](https://en.wikipedia.org/wiki/Voronoi_diagram).
The related functions closest/gclosest help with finding the closest point.
The result looks kinda like [Giant's Causeway](https://en.wikipedia.org/wiki/Giant%27s_Causeway).

TomyLobo:
Hmm, i even stole some implementation details from AVS, like the 2-tier system and even the size of the 2nd-tier cells :)
Doesn't matter though, since AVS has been open source since 2005, long before my involvement with WorldEdit.

PR #453, opened 2019-03-16
Add RequestExtent to be used when a delayed EditSession is needed.

wizjany:
For example, if you set a mask that takes an extent (many of them),
and then move to another world, the mask will test blocks in the old
world and return bad results.

~~Downside: ThreadLocal is slow?~~

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/session/request/RequestExtent.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/session/request/RequestExtent.java
@@ -1,0 +1,109 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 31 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/session/request/RequestExtent.java
```diff
@@ -0,0 +1,109 @@
...
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+public class RequestExtent implements Extent {
+
+    private EditSession extent;
+
+    protected Extent getExtent() {
+        if (extent == null) {
```
TomyLobo: Since you're caching it now, are you sure the EditSession's lifetime is guaranteed to be greater than that of the RequestExtent?
What if the player leaves?

The problem with lazy creation is that problems with it will only show up much much later.

TomyLobo: Oh I just saw you null that field again in `.commit()`. I don't know WE's new structure very well. Is that the function that is  called at the end of a time slice or something?

wizjany: Yea, this doesn't actually work yet. You miss out of the discord discussion, but essentially this extent tends to be used for masks which mainly do gets, not sets, so this extent isn't actually committed anywhere. The "correct" place to check would be when `Request#setEditSession` and `Requests#reset` are called.

PR #454, opened 2019-03-31
Remove redundant block vector method to avoid confusion

boy0001:
BlockVector class has both getX and getBlockX etc. which is redundant

This removes getX in favor of getBlockX
(as getX is used by Vector2/3)


me4502:
I’m not sure how I feel about this - it makes sense to keep them both for consistency reasons tbh

DarkArc:
We can't do this without an ABI break as well... I feel like we've done enough of that for a while at this point. People are going to get mad.

boy0001:
For consistency, wouldn't getX/Y/Z return a double, not an int?

PR #456, opened 2019-04-03
Remove "switch for efficiency on typical cases"

boy0001:
It's not an efficient tradeoff for how common it is.

Also change the BlockVector hashcode, because it was better before. idk
why you reverted it in the new vector system update.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/math/BlockVector2.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/math/BlockVector2.java
@@ -60,18 +60,6 @@ public final class BlockVector2 {
     }
 
     public static BlockVector2 at(int x, int z) {
-        switch (x) {
-            case 0:
-                if (z == 0) {
-                    return ZERO;
-                }
-                break;
-            case 1:
-                if (z == 1) {
-                    return ONE;
-                }
-                break;
-        }
         return new BlockVector2(x, z);
     }
 
@@ -525,10 +513,7 @@ public final class BlockVector2 {
 
     @Override
     public int hashCode() {
-        int hash = 17;
-        hash = 31 * hash + Integer.hashCode(x);
-        hash = 31 * hash + Integer.hashCode(z);
-        return hash;
+        return (x << 16) ^ z;
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


wizjany:
Wasn't so much reverted as the classes were rewritten. I wasn't part of this decision making process so I'm not sure if there's any reason for the hash code or the switch though.

boy0001:
looks like me4502 was going to try adding object pooling at some point. Aside from specific cases, I wouldn't think that'd have much benefit either since most vectors are short lived and probably optimized away by the jvm.

The current hashcode is just a generic java approach of multiplying it by a prime number. Objectively it'd be marginally slower and result in more collisions.

worldedit-core/src/main/java/com/sk89q/worldedit/math/BlockVector3.java
```diff
@@ -35,27 +35,16 @@
     public static final BlockVector3 UNIT_X = new BlockVector3(1, 0, 0);
     public static final BlockVector3 UNIT_Y = new BlockVector3(0, 1, 0);
     public static final BlockVector3 UNIT_Z = new BlockVector3(0, 0, 1);
+    public static final BlockVector3 UNIT_MINUS_X = new BlockVector3(-1, 0, 0);
+    public static final BlockVector3 UNIT_MINUS_Y = new BlockVector3(0, -1, 0);
+    public static final BlockVector3 UNIT_MINUS_Z = new BlockVector3(0, 0, -1);
```
wizjany: does it make sense to add the negative units to Vector as well?

boy0001: I didn't see anywhere need the other ones.

worldedit-core/src/main/java/com/sk89q/worldedit/math/BlockVector2.java
```diff
@@ -60,18 +60,6 @@ public static BlockVector2 at(double x, double z) {
     }
 
     public static BlockVector2 at(int x, int z) {
-        switch (x) {
```
kenzierocks: This & the other switch removals should be reverted, performance tests show that this isn't any more efficient. The change also increases memory usage, making this a net negative change.

PR #458, opened 2019-04-04
Allow copy/pasting biomes.

wizjany:
Copy takes a -b flag to copy biomes.
Paste takes a -b flag to paste biomes (if available).
This allows flexibility to create/load schematics with/without biomes
(when schematic biome support is added).

Also added a -m mask flag to paste to set a source mask, and a -e flag
to skip pasting entities if they are loaded.

Uses the Schematic format v2: https://github.com/SpongePowered/Schematic-Specification/pull/13 to save and load to/from schematics.
Note that there's currently some blockers there (eg y-axis in biomes) that I spoke to gabizou about, so this should technically work once he gets to updating/finalizing the spec.


wizjany:
Point for consideration: at the moment, you have to opt-in to copying entities (`//copy -e`), but opt-out of pasting them (`//paste -e`). This is rather user-unfriendly, but maintains previous behavior.

I think it might be worth changing this so that biomes and entities work the same and are consistent across copying and pasting

worldedit-core/src/main/java/com/sk89q/worldedit/function/biome/ExtentBiomeCopy.java
```diff
@@ -0,0 +1,73 @@
...
+import com.sk89q.worldedit.math.BlockVector3;
+import com.sk89q.worldedit.math.transform.Transform;
+import com.sk89q.worldedit.world.biome.BiomeType;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Copies the biome from one extent to another.
+ */
+public class ExtentBiomeCopy  implements FlatRegionFunction {
```
me4502: two spaces :O

wizjany:
note to self: update clipboard brush flags

PR #459, opened 2019-04-11
Remove the system property check for uuidOverride

boy0001:
as it caused a server timeout...
https://pastebin.com/z5ktU8r3


kenzierocks:
How did this _cause_ it? Is your system property table being hash-DoS'd or something? This is essentially a no-cost operation.

boy0001:
it's a deadlock, so most likely a synchronization issue. It didn't happen to me, it happened to someone else, I just have the log. 

Easier to just remove something that isn't used than debug what caused the deadlock.

wizjany:
as stupid as this scenario is, i'm guessing that was originally added to debug sessions and has never been used since.

PR #460, opened 2019-04-17
Fix 1.13 entity direction code, port old schematics

kenzierocks:
1.13 changed the `Facing` tag on us.

I added some code to the MCEdit reader to port old values to 1.13 as well.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
@@ -144,25 +144,24 @@ public class ExtentEntityCopy implements EntityFunction {
                         .putInt("TileZ", newTilePosition.getBlockZ());
 
                 if (hasDirection || hasLegacyDirection || hasFacing) {
-                    int d;
+                    Direction direction;
                     if (hasDirection) {
-                        d = tag.asInt("Direction");
+                        direction = MCDirections.fromPre13Hanging(tag.asInt("Direction"));
                     } else if (hasLegacyDirection) {
-                        d = MCDirections.fromLegacyHanging((byte) tag.asInt("Dir"));
+                        direction = MCDirections.fromPre13Hanging(
+                            MCDirections.fromLegacyHanging((byte) tag.asInt("Dir"))
+                        );
                     } else {
-                        d = tag.asInt("Facing");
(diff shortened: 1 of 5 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
```diff
@@ -144,25 +144,24 @@ private BaseEntity transformNbtData(BaseEntity state) {
...
-
                     if (direction != null) {
                         Vector3 vector = transform.apply(direction.toVector()).subtract(transform.apply(Vector3.ZERO)).normalize();
                         Direction newDirection = Direction.findClosest(vector, Flag.CARDINAL);
 
                         if (newDirection != null) {
-                            byte hangingByte = (byte) MCDirections.toHanging(newDirection);
-                            builder.putByte("Direction", hangingByte);
-                            builder.putByte("Facing", hangingByte);
+                            builder.putByte("Direction", (byte) MCDirections.toPre13Hanging(newDirection));
```
me4502: Wait does it store both?

kenzierocks: What do you mean? I didn't change `Direction` because it's not a Minecraft tag, so if it's used by mods I don't think they'll be changing it.

wizjany: https://i.imgur.com/jbGhDOU.png

kenzierocks: Hmm, I added a schematic handler for that -- perhaps if that does it, we don't need this anymore then?

wizjany: yea, if it's done on load it probably doesn't need to be done here anymore.

i think in general it's good to keep compatibility things (including eventually datafixeruppers maybe) in the schematic reading code, as that stuff is actually pretty threadsafe and could probably be done on another thread at some point (including the file i/o itself - loading big schematics can be slow as well).

wizjany: I removed the superfluous tags in my additions to this branch. #463

PR #462, opened 2019-04-22
feat(text): added modern text formatting to worldedit, and switch a few things over to it

me4502:
WorldGuard update branch: https://github.com/EngineHub/WorldGuard/compare/feature/text-update


worldedit-forge/build.gradle
```diff
@@ -19,6 +19,7 @@ def forgeVersion = "25.0.76"
 dependencies {
     compile project(':worldedit-core')
     compile 'org.apache.logging.log4j:log4j-slf4j-impl:2.11.2'
+    compile 'net.kyori:text-serializer-gson:2.0.0'
```
kenzierocks: We should drop this dependency here, it's shaded in core.

build.gradle
```diff
@@ -82,6 +82,15 @@ artifactory {
 artifactoryPublish.skip = true
 
 subprojects {
+    repositories {
+        mavenCentral()
+        maven { url "http://maven.sk89q.com/repo/" }
+        maven { url "http://repo.maven.apache.org/maven2" }
```
kenzierocks: Remove this repo? It's just `mavenCentral()` again.

me4502: Oh, yeah - no idea how this got added

config/checkstyle/import-control.xml
```diff
@@ -16,6 +16,7 @@
   <allow pkg="net.royawesome.jlibnoise"/>
   <allow pkg="org.json.simple" />
   <allow pkg="org.slf4j"/>
+  <allow pkg="net.kyori.text"/>
```
kenzierocks: We can actually drop this, since we use the relocated package everywhere.

worldedit-core/src/main/java/com/sk89q/worldedit/util/formatting/component/TextComponentProducer.java
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.util.formatting.component;
+
+import com.sk89q.worldedit.util.formatting.text.Component;
+import com.sk89q.worldedit.util.formatting.text.TextComponent;
+
+public class TextComponentProducer {
```
kenzierocks: Can we add some more utility-calls to this class? It looks like there is a lot of need to pass raw String & TextComponentProducer, and I think those are worth adding as their own `append` overrides.

me4502: I'll add some more yeah - I originally didn't as I wanted it to explicitly error when I hadn't ported something over.

PR #464, opened 2019-04-26
Piston command system

kenzierocks:
This PR switches WorldEdit to the [Piston](https://github.com/EngineHub/Piston) command system, and (might) remove all other command systems.

TODO:
- [x] Fix some minor bugs in Piston related to help text, and whatever else comes up in testing
- [x] Port a few non-standard commands (brushes and `//set`)
- [x] Add quoted string parsing to WE
- [x] Test EVERY command in WorldEdit
  - [x] BiomeCommands
  - [x] BrushCommands
    - [x] PasteBrushCommands
    - [x] ApplyBrushCommands
  - [x] ChunkCommands
  - [x] ClipboardCommands
  - [x] GeneralCommands
  - [x] GenerationCommands
  - [x] HistoryCommands
  - [x] NavigationCommands
  - [x] RegionCommands
  - [x] SchematicCommands
  - [x] ScriptingCommands
  - [x] SelectionCommands
  - [x] SnapshotCommands
  - [x] SnapshotUtilCommands
  - [x] SuperPickaxeCommands
  - [x] ToolCommands
  - [x] ToolUtilCommands
  - [x] UtilityCommands
  - [x] WorldEditCommands
- [x] (Maybe) Remove all other command systems
- [x] Release Piston, update this PR with it


me4502:
http://paste.enginehub.org/gN5sOA I'm getting this - not sure if I've done something wrong?

kenzierocks:
@me4502 fixed, needed to not use the context classloader when searching.

kenzierocks:
Piston is released, this PR is ready for a final review set, and I think we are able to remove the old command systems, if we are so inclined.

wizjany:
The original system in `com/sk89q/minecraft/util/commands` is still in use by WG/CB, so we can't remove that until those are updated I guess.
The Intake-like system in `com/sk89q/worldedit/util/command` is probably safe to remove. It was never really public as Intake itself was meant to be the public library version of that. Note that this excludes the AsyncCommandHelper system (used by both WE + WG atm), which isn't part of the Intake/parametric framework, but actually just `util` :P

PR #465, opened 2019-04-26
Print //calc input before its result

TheMolkaPL:
Currently, `//calc` history is unreadable when many calculations are performed.

image

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
@@ -545,12 +545,15 @@ public class UtilityCommands {
     public void calc(Actor actor, @Text String input) throws CommandException {
         try {
             Expression expression = Expression.compile(input);
+            double result;
             if (actor instanceof SessionOwner) {
-                actor.print("= " + expression.evaluate(
-                        new double[]{}, WorldEdit.getInstance().getSessionManager().get((SessionOwner) actor).getTimeout()));
+                result = expression.evaluate(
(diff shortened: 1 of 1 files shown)


me4502:
What does it look like now? Would it make more sense to possibly use a hover event to keep it looking clean?

TheMolkaPL:
The screenshot I provided is taken before this change. I print user input before the expression result.
Hover event is a solution but would not work on console properly.

wizjany:
I'd rather have the input and result on one line. Preferably with padding. That way if you do multiple successive calculations they will line up, instead of trying to read two line chunks at a time.
eg.
```
10     = 5+ 5
400   = 20 * 20
```
issue is we don't know how much padding is needed if future results get big. Alternative is input first:
```
5 + 5 = 10
20 * 20 = 400
```
which also doesn't line up tho.

TheMolkaPL:
This would be unreadable with bigger expressions (with eg. loops).

wizjany:
My suggestion, or your code? if the input it long enough it'll wrap regardless of whether you're printing on one line or two.

TheMolkaPL:
Your mean. The result would be hard to read in tons of numbers.

me4502:
Can you please provide a screenshot of what it looks like with your latest changes?

TheMolkaPL:
Sure

58a47c952b3fe643af7cd1a7d49ea34b6e594a73:

image

wizjany:
following the discussion in discord, here's how it'd look with formatting (color, decimalformat) for readability:
https://i.imgur.com/Jte9QDL.png
or
https://i.imgur.com/Dplt5hP.png

TheMolkaPL:
b3053f19ce07be2b5eb7b12cf59feb51574f8ae2

PR #466, opened 2019-05-01
Insert Locale.ROOT into all case change methods

kenzierocks:
This makes all case changes locale-insensitive. I basically did a find & replace, but it doesn't look like there's any places where we have display-facing case changes where this might matter.

I also added a checkstyle rule, to prevent this issue from popping up in the future. If for some reason, we need a locale-sensitive case change, it's easy to add it to a `checkstyle-suppressions` file to exclude that file from the check, or to explicitly pass `Locale.getDefault()` (preferred solution).

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/snapshot/Snapshot.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/snapshot/Snapshot.java
@@ -35,6 +35,7 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.time.ZonedDateTime;
+import java.util.Locale;
 import java.util.zip.ZipFile;
 
 /**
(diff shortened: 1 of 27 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/snapshot/Snapshot.java
```diff
@@ -101,9 +103,9 @@ private ChunkStore internalGetChunkStore() throws IOException, DataException {
 
                 return chunkStore;
             }
-        } else if (file.getName().toLowerCase().endsWith(".tar.bz2")
-                || file.getName().toLowerCase().endsWith(".tar.gz")
-                || file.getName().toLowerCase().endsWith(".tar")) {
+        } else if (lowerCaseFileName.endsWith(".tar.bz2")
```
me4502: Will this break if they legitimately have other locale characters in their path?

PR #467, opened 2019-05-04
Piston-based suggestions

kenzierocks:
This PR implements suggestions via Piston.

TODO:
- [x] Figure out how WorldEdit can send suggestions back to all platforms
  - [x] Figure out how to map between `String args`-style and `String[] args`-style for suggestions
- [ ] Possibly explore implementing commands directly in Brigadier for Forge, to enhance speed on client-side completions. This might best be left as an improvement in a future PR, and for now we can bind ASK_SERVER.

### DIFF
--- a/worldedit-bukkit/build.gradle
+++ b/worldedit-bukkit/build.gradle
@@ -20,7 +20,7 @@ dependencies {
     compile 'com.sk89q:dummypermscompat:1.10'
     compile 'org.bukkit:bukkit:1.13.2-R0.1-SNAPSHOT' // zzz
     compile "io.papermc:paperlib:1.0.1"
-    compile 'org.slf4j:slf4j-jdk14:1.7.26'
+    compile 'org.apache.logging.log4j:log4j-slf4j-impl:2.8.1'
     compile 'org.bstats:bstats-bukkit:1.4'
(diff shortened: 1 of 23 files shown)


kenzierocks:
Note: This PR currently leaves out the sending of suggestions, due to problems I noted in the Discord server. Copied here for convenience:

>octylFractalToday at 00:53
so suggestions needs... some discussion
Piston's model is obviously based on List<String> as the input args, and that's fine
Brigadier works on a String, StringRange, and has the suggestions replace that range
which imo is much nicer, and it's pretty simple for us to map between them
but WorldEdit and Bukkit are modeled as an input String, and a List<String> to replace the "last argument"
which... won't work, since I think WE likely defines "last argument" differently from bukkit, and we don't always want to replace the "last argument" (Piston can choose arguments that are beyond the last one)
so I'm wondering if we should be re-modeling WE to be more like Brigadier, because that's likely to be compatible with all systems, and just drop suggestions in the bukkit-impl if they're not going to be valid
also, re: "pretty simple for us to map between them"

>this is another part where worldedit makes this hard™, because we have commandDetection. wiz said earlier we might just want to drop this, and I think this would be a good idea (besides maybe still allowing the craftscript shortcuts)

PR #470, opened 2019-05-12
Add AsyncCommandBuilder

wizjany:
Replaces AsyncCommandHelper, see https://github.com/EngineHub/WorldGuard/pull/408 for info


worldedit-core/src/main/java/com/sk89q/worldedit/command/util/AsyncCommandBuilder.java
```diff
@@ -0,0 +1,171 @@
...
+                        if (converted instanceof CommandExecutionException) {
+                            Throwable parentCause = converted;
+                            while ((parentCause = parentCause.getCause()) != null) {
+                                if (parentCause instanceof com.sk89q.minecraft.util.commands.CommandException) {
+                                    targetEx = ((com.sk89q.minecraft.util.commands.CommandException) parentCause);
+                                    break;
+                                }
+                            }
+                        }
+                        String message = targetEx.getMessage() != null ? targetEx.getMessage() : "An unknown error occurred.";
```
kenzierocks: You should use `getRichMessage()` or whatever I called it, to retrieve the full `Component` message, in the case of a regular `CommandException` for `targetEx`. Otherwise the message might not be very pretty.

worldedit-core/src/main/java/com/sk89q/worldedit/util/paste/EngineHubPaste.java
```diff
@@ -35,11 +34,11 @@
...
 
     @Override
-    public ListenableFuture<URL> paste(String content) {
-        return Pasters.getExecutor().submit(new PasteTask(content));
+    public Callable<URL> paste(String content) {
+        return new PasteTask(content);
     }
 
-    private final class PasteTask implements Callable<URL> {
+    public static final class PasteTask implements Callable<URL> {
```
kenzierocks: Is there a reason this was made `public`?

wizjany: No. I remember making it static but not public :thinking: fixed

worldedit-core/src/main/java/com/sk89q/worldedit/command/util/AsyncCommandBuilder.java
wizjany: done. it doesn't matter much atm since WG only throws (2011)CommandExceptions, but it will once it throws (Piston)CommandExceptions.

PR #471, opened 2019-05-12
Replaced WorldEdit with ${project.name}

MrNoel13:


wizjany:
The tokens in this file were replaced by an ant script and the file bundled in a zip which we used to distribute.
That is long gone. If we want to keep this file, it either needs to be made static or someone needs to rewrite (probably as part of the gradle process) the replacement script. Also since we don't distribute a zip anymore, there's nowhere to put it. I'd rather just work on the real documentation than worry about this relic of the past.

PR #472, opened 2019-05-16
Add more DFU.

wizjany:
Legacy mapper now uses the data fixers to upgrade blocks and item types
(e.g. signs, dyes that changed names in 1.14).
The sponge schematic reader can now attempt to use the data fixers to
upgrade blocks, block entities, biomes and entities. This has been
tested with the 1.13 -> 1.14 changes. It is yet to be seen if it will
continue to work because...
The mc edit schematic reader has code for using data fixers, but it is
currently disabled as there seem to be some issues with fixing up older
block entities.


TomyLobo:
I'd prefer adding more FUD.

PR #473, opened 2019-05-18
Added suggestions to a lot of factory-related commands

me4502:
Few talking points:

Patterns/Masks are real complex and allow multiple to be used. This does not handle that, it's just a basic level of functionality to cover simple cases of setting a block. Not sure how well Piston/MC supports this.

In vanilla commands (/fill) it appears to provide a tonne more feedback with usage/command sections/colours - is this possible to achieve?

Also - for suggesting completions on BlockState properties, idk how that'll be achieved in terms of how the system works.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/FactoryConverter.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/FactoryConverter.java
@@ -96,7 +96,7 @@ public class FactoryConverter<T> implements ArgumentConverter<T> {
 
     @Override
     public List<String> getSuggestions(String input) {
-        return limitByPrefix(factoryExtractor.apply(worldEdit).getSuggestions(), input);
+        return factoryExtractor.apply(worldEdit).getSuggestions(input);
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


wizjany:
In general, I dislike the no-arg `getSuggestions` method. If we pass in the current context, each factory can generate a lot more useful suggestions, including parsing the current block to generate states for that block (e.g., if `slab[ty` is types, we can suggest `slab[type=top]` or `slab[type=bottom]` or even `slab[type=top,`)

DarkArc:
I think that's a fair point @wizjany. Also definitely have concerns about insanely log tab complete lists if the entire registry is up for tab complete -- though maybe I'm just misunderstanding how this would work.

wizjany:
the client has a scrollable list. https://i.imgur.com/7Intbb3.png

DarkArc:
Witchcraft, so fancy compared to the olden days I'm used to lol

me4502:
Yeah massive lists isn't an issue - vanilla commands have the same issues tbh. 

As for the vanilla stuff, yeah - I'm aware we're not sending a graph to the client. Was just wondering if there were ways around some of it.

PR #474, opened 2019-05-23
Add /tracemask.

wizjany:
Allows setting a mask used for block traces. This allows brush tools to
pass through various materials, such as water (e.g. `/tracemask #solid`
or `/tracemask !air,water`) before starting to build.
By default, a null mask is equivalent to #existing (original behavior).

---

https://gfycat.com/ImmaculateFrayedCockatiel

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
@@ -51,7 +51,7 @@ public class ToolUtilCommands {
     @CommandPermissions("worldedit.superpickaxe")
     public void togglePickaxe(Player player, LocalSession session,
                               @Arg(desc = "The new super pickaxe state", def = "")
-                                  Boolean superPickaxe) throws WorldEditException {
+                                  Boolean superPickaxe) {
         boolean hasSuperPickAxe = session.hasSuperPickAxe();
         if (superPickaxe != null && superPickaxe == hasSuperPickAxe) {
             player.printError("Super pickaxe already " + (superPickaxe ? "enabled" : "disabled") + ".");
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
@@ -133,7 +133,7 @@ public class ToolUtilCommands {
     )
     @CommandPermissions("worldedit.brush.options.tracemask")
     public void traceMask(Player player, LocalSession session,
-                     @Arg(desc = "The mask to set", def = "")
+                          @Arg(desc = "The mask to set", def = "")
                              Mask mask) throws WorldEditException {
         if (mask == null) {
             session.getBrushTool(player.getItemInHand(HandSide.MAIN_HAND).getType()).setTraceMask(null);
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
@@ -22,15 +22,19 @@ package com.sk89q.worldedit.command;
 import com.sk89q.worldedit.LocalSession;
 import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.WorldEditException;
+import com.sk89q.worldedit.command.tool.Tool;
+import com.sk89q.worldedit.command.tool.TraceTool;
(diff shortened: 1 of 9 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
```diff
@@ -22,15 +22,19 @@
...
 import com.sk89q.worldedit.command.util.CommandPermissionsConditionGenerator;
 import com.sk89q.worldedit.entity.Player;
 import com.sk89q.worldedit.function.mask.Mask;
 import com.sk89q.worldedit.function.pattern.Pattern;
 import com.sk89q.worldedit.util.HandSide;
+import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import org.enginehub.piston.annotation.Command;
 import org.enginehub.piston.annotation.CommandContainer;
 import org.enginehub.piston.annotation.param.Arg;
+import org.enginehub.piston.exception.StopExecutionException;
```
kenzierocks: Looks like an unused import?

worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
```diff
@@ -122,4 +126,21 @@ public void size(Player player, LocalSession session,
...
+    )
+    @CommandPermissions("worldedit.brush.options.tracemask")
+    public void traceMask(Player player, LocalSession session,
+                     @Arg(desc = "The mask to set", def = "")
+                             Mask mask) throws WorldEditException {
+        if (mask == null) {
+            session.getBrushTool(player.getItemInHand(HandSide.MAIN_HAND).getType()).setTraceMask(null);
+            player.print("Trace mask disabled.");
+        } else {
+            session.getBrushTool(player.getItemInHand(HandSide.MAIN_HAND).getType()).setTraceMask(mask);
```
kenzierocks: Wouldn't it be a better idea to move this line outside the conditional? It's essentially the same as the one before.

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BrushTool.java
```diff
@@ -86,6 +86,24 @@ public void setMask(Mask filter) {
         this.mask = filter;
     }
 
+    /**
+     * Get the mask used for identifying where to stop traces
```
kenzierocks: Might want to add a period here.

DarkArc:
Awesome idea :)

worldedit-core/src/main/java/com/sk89q/worldedit/entity/Player.java
```diff
@@ -210,6 +211,17 @@
...
+    /**
+     * Get the point of the block being looked at. May return null.
+     * Will return the farthest away block before matching the stop mask if useLastBlock is true and no other block is found.
+     *
+     * @param range how far to checks for blocks
+     * @param useLastBlock try to return the last valid block not matching the stop mask found
+     * @param stopMask the mask used to determine when to stop tracing
+     * @return point
+     */
+    Location getBlockTrace(int range, boolean useLastBlock, Mask stopMask);
```
kenzierocks: Would it maybe make sense to make the `stopMask` nullable? Seems like there's a lot of branching around this call, when we could just pass it through.

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BrushTool.java
```diff
@@ -47,6 +46,7 @@
     protected static int MAX_RANGE = 500;
     protected int range = -1;
     private Mask mask = null;
+    private Mask traceMask = null;
```
kenzierocks: Should this default to the same setting as `TargetBlock`? Seems odd to have it differ.

worldedit-core/src/main/java/com/sk89q/worldedit/entity/Player.java
wizjany: tbh i'm used to just making overrides for everything to maintain API compat.

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BrushTool.java
wizjany: Should `Mask` also default to Masks.alwaysTrue()? I think no because then we end up with `MaskIntersection` of `(Masks.alwaysTrue(), Masks,alwaysTrue())`... which should optimize but doesn't :P

kenzierocks: This is fine.

PR #475, opened 2019-05-24
Restore Forge regen functionality

kenzierocks:
No idea if this has a chance of breaking with mods or something. It's basically what we used in 1.12.2, except the removal of `setWorld`, which probably doesn't matter since the world won't be inserted into the dim map.


me4502:
Curious as to how easy this would be to actually make async (At least the gen part). I guess we'd need to setup WorldEdit to use callbacks a lot more, which would be a fair bit of work.

wizjany:
not completely relevant to this PR, but wrt regen in general - if we can start regening regions without regenning whole chunks in the world, it may be worth signalling to editsession that it doesn't have to go save the unaffected areas within the chunks and restore them after the regen. this may show up again if we have to rewrite bukkit regen like this in @.14 since spigot looks like they will no longer support it in api.

PR #478, opened 2019-06-01
feat(i18n): add support for translatable text replacement

me4502:
TODO:

- [x] Make all text translatable
- [x] Setup CI to upload/download translations
- [x] Decide if a user-provided defaults file in config directory should prevent ALL jar-based languages loading


worldedit-core/src/main/java/com/sk89q/worldedit/util/translation/LocalisationHelpers.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Turn a translation key into a ".singular" or ".plural"
+     * depending on what the given number is.
+     *
+     * @param translationKey The base translation key
+     * @param number The number
+     * @return The key with .plural or .singular appended
+     */
+    public static String pluraliseI18n(String translationKey, float number) {
+        if (number == 1) {
```
Tobi406: Can the if statement there may be changed? To `if (number == 1  || number == -1)`. As it should be eg. "-1 block".

MattBDev: The only number that is singular is the number one. Any other number, including 0 and negatives are considered plural. I think the way it is right now is fine

MattBDev: The only number that is singular is the number one. Any other number, including 0 and negatives are considered plural. I think the way it is right now is fine

MattBDev: The only number that is singular is the number one. Any other number, including 0 and negatives are considered plural. I think the way it is right now is fine

MattBDev: The only number that is singular is the number one. Any other number, including 0 and negatives are considered plural. I think the way it is right now is fine

MattBDev: The only number that is singular is the number one. Any other number, including 0 and negatives are considered plural. I think the way it is right now is fine

MattBDev: The only number that is singular is the number one. Any other number, including 0 and negatives are considered plural. I think the way it is right now is fine.

MattBDev: The only number that is singular is the number one. Any other number, including 0 and negatives are considered plural. I think the way it is right now is fine.

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -85,6 +115,9 @@
     "worldedit.timezone.current": "The current time in that timezone is: {0}",
     "worldedit.version.version": "WorldEdit version {0}",
 
+    "worldedit.command.time-elapsed": "{0}s elapsed.",
+    "worldedit.command.time-elapsed.verbose": "{0}s elapsed (history: {0} changed; {1} blocks/sec).",
```
wizjany: 0 - 1 - 2?

worldedit-core/src/main/java/com/sk89q/worldedit/util/translation/LocalisationHelpers.java
tyggerr: I'm just a random user, but I have an issue with this : in French -1, 0, 1, 1.5 and even 1.99 are all singular. Which means this system will not work correctly. 
This also breaks in Russian and most asian languages where the plural changes based on the number.

wizjany: Yes, this has been discussed before, but without a full localization lib to handle all the cases there will always be issues like this
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals

me4502: A decision has been made to ignore all forms of pluralisation, similar to what Minecraft does. This prevents all of these corner cases.

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorldEdit.java
```diff
@@ -237,13 +238,13 @@ public void run(InputStream inputStream) {
             while (scanner.hasNextLine()) {
                 String line = scanner.nextLine();
                 if (line.equals("stop")) {
-                    commandSender.print("Stopping!");
+                    commandSender.printInfo(TextComponent.of("Stopping!"));
```
octylFractal: No localization keys for these?

worldedit-core/build.gradle.kts
```diff
@@ -84,3 +87,28 @@ sourceSets {
...
 }
+
+val crowdinApiKey = "crowdin_apikey"
+
+if (project.hasProperty(crowdinApiKey)) {
+    tasks.named<UploadSourceFileTask>("crowdinUpload") {
+        apiKey = "${project.property(crowdinApiKey)}"
+        projectId = "worldedit-core"
+        files = arrayOf(
+                object {
```
octylFractal: I feel like this would be better represented as a Map,
```gradle.kts
mapOf("name" to "string.json", "source" to "...")
```

Additionally, `source` would be shorter as `"${file("src/main/resources/lang/strings.json")}"`

worldedit-core/build.gradle.kts
```diff
@@ -84,3 +87,28 @@ sourceSets {
...
+                    var name = "strings.json"
+                    var source = "$projectDir/src/main/resources/lang/strings.json"
+                }
+        )
+    }
+
+    tasks.named<DownloadTranslationsTask>("crowdinDownload") {
+        apiKey = "${project.property(crowdinApiKey)}"
+        destination = "$projectDir/src/main/resources/lang"
```
octylFractal: Similar for resolving files. Also, I feel like this this task should output to the generated code, unless this is intended to be committed to source?

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -116,7 +118,7 @@ public void sphereBrush(Player player, LocalSession session,
             tool.setBrush(new SphereBrush(), "worldedit.brush.sphere");
         }
 
-        player.print(String.format("Sphere brush shape equipped (%.0f).", radius));
+        player.printInfo(TranslatableComponent.of("worldedit.brush.sphere.equip", TextComponent.of((int) radius)));
```
octylFractal: Did you intend to lose precision here? Old code could format a one-digit decimal as well, but this is just integers.

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -85,14 +88,17 @@ public void limit(Actor actor, LocalSession session,
...
                 return;
             }
         }
 
         session.setBlockChangeLimit(limit);
-        actor.print("Block change limit set to " + limit + "."
-                + (limit == config.defaultChangeLimit ? "" : " (Use //limit to go back to the default.)"));
+        Component component = TextComponent.empty().append(TranslatableComponent.of("worldedit.limit.set", TextComponent.of(limit)));
+        if (limit != config.defaultChangeLimit) {
+            component.append(TranslatableComponent.of("worldedit.limit.return-to-default", TextColor.GRAY));
```
octylFractal: Do you need to add a space here?

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -171,17 +180,17 @@ public void drawSelection(Player player, LocalSession session,
         }
         boolean useServerCui = session.shouldUseServerCUI();
         if (drawSelection != null && drawSelection == useServerCui) {
-            player.printError("Server CUI already " + (useServerCui ? "enabled" : "disabled") + ".");
+            player.printError(TranslatableComponent.of(useServerCui ? "worldedit.drawsel.enabled.already" : "worldedit.drawsel.disabled.already"));
```
octylFractal:
```suggestion
            player.printError(TranslatableComponent.of("worldedit.drawsel.already." + (useServerCui ? "enabled" : "disabled")));
```

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -271,16 +280,16 @@ public void searchItem(Actor actor,
                                List<String> query) {
         String search = String.join(" ", query);
         if (search.length() <= 2) {
-            actor.printError("Enter a longer search string (len > 2).");
+            actor.printError(TranslatableComponent.of("worldedit.searchitem.too-short"));
             return;
         }
         if (blocksOnly && itemsOnly) {
-            actor.printError("You cannot use both the 'b' and 'i' flags simultaneously.");
+            actor.printError(TranslatableComponent.of("worldedit.searchitem.b-and-i"));
```
octylFractal: I feel like the translation key would be better as `.no-block-and-item-flags`, or at least something more verbose.

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/mixin/AccessorServerPlayerEntity.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.sk89q.worldedit.fabric.mixin;
+
+import net.minecraft.server.network.ServerPlayerEntity;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+@Mixin(ServerPlayerEntity.class)
+public interface AccessorServerPlayerEntity {
```
octylFractal: Why is this separate from the regular Mixin? It seems like accessors should work fine there?

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/mixin/MixinServerPlayerEntity.java
```diff
@@ -27,6 +27,7 @@
 import net.minecraft.util.Hand;
 import net.minecraft.world.World;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
```
octylFractal: Extra import.

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/mixin/AccessorServerPlayerEntity.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Accessor("clientLanguage")
```
octylFractal: You don't need the value explicitly set here, to quote the docs:

>Normally the target field name is inflected by examining the annotated method name. If the annotated method starts with "get", "set" or "is" followed by a capital letter, then the prefix is stripped and the remainder of the method name is used as the target field name.

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricPlayer.java
```diff
@@ -128,6 +130,11 @@ public void dispatchCUIEvent(CUIEvent event) {
         this.player.networkHandler.sendPacket(packet);
     }
 
+    @Override
+    public Locale getLocale() {
+        return Locale.forLanguageTag(((AccessorServerPlayerEntity) this.player).getClientLanguage().replace("_", "-"));
```
octylFractal: IIRC this doesn't work, Mixin interfaces don't exist at runtime, only their implemented interfaces. So you would want to create a `ServerPlayerEntityAccess` interface, list the `String getClientLanguage()` method there (NO mixin annotations), then implement that in `MixinServerPlayerEntity` using `@Accessor`. Then here, you cast to `ServerPlayerEntityAccess` and can call the method.

worldedit-core/src/main/java/com/sk89q/worldedit/util/translation/TranslationManager.java
```diff
@@ -0,0 +1,160 @@
...
+    public void setDefaultLocale(Locale defaultLocale) {
+        this.defaultLocale = defaultLocale;
+    }
+
+    private Map<String, String> filterTranslations(Map<String, String> translations) {
+        translations.entrySet().removeIf(entry -> entry.getValue().isEmpty());
+        return translations;
+    }
+
+    private Map<String, String> parseTranslationFile(File file) throws IOException {
```
octylFractal: Probably better to merge these into one `parseTranslationFile(InputStream)`, use `InputStreamReader` and pass that to GSON.

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/ResourceLoader.java
```diff
@@ -40,4 +40,17 @@ public static URL getResource(Class clazz, String name) throws IOException {
         }
         return url;
     }
+
+    public static URL getResourceRoot(String name) throws IOException {
```
octylFractal: Can you make the other method delegate to this one, or vice versa if you think that's easier? Would like to not repeat the modjar hack.

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/mixin/AccessorServerPlayerEntity.java
me4502: From memory, it's because you can't just blindly cast things to a regular mixin, but you can with an Accessor

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricPlayer.java
me4502: I think this is the reason I used the separate accessor mixin - as they work in this situation

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -205,7 +207,12 @@ public void smoothBrush(Player player, LocalSession session,
         tool.setSize(radius);
         tool.setBrush(new SmoothBrush(iterations, mask), "worldedit.brush.smooth");
 
-        player.print(String.format("Smooth brush equipped (%.0f x %dx, using %s).", radius, iterations, mask == null ? "any block" : "filter"));
+        player.printInfo(TranslatableComponent.of(
+                "worldedit.brush.smooth.equip",
+                TextComponent.of((int) radius),
+                TextComponent.of(iterations),
+                TextComponent.of(mask == null ? "any block" : "filter")
```
octylFractal: Shouldn't these be translated too? Either as two separate top-level keys (`.equip.anyblock`/`.equip.filter`) or passed as translated args (`worldedit.mask.none`/`worldedit.mask.applied`).

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -187,7 +188,7 @@ public void save(Actor actor, LocalSession session,
         SchematicSaveTask task = new SchematicSaveTask(actor, f, format, holder, overwrite);
         AsyncCommandBuilder.wrap(task, actor)
                 .registerWithSupervisor(worldEdit.getSupervisor(), "Saving schematic " + filename)
```
octylFractal: Don't we need some translations for these other messages as well? Or are we going to do a second pass on them?

worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
```diff
@@ -54,15 +55,15 @@ public void togglePickaxe(Player player, LocalSession session,
                                   Boolean superPickaxe) {
         boolean hasSuperPickAxe = session.hasSuperPickAxe();
         if (superPickaxe != null && superPickaxe == hasSuperPickAxe) {
-            player.printError("Super pickaxe already " + (superPickaxe ? "enabled" : "disabled") + ".");
+            player.printError(TranslatableComponent.of(superPickaxe ? "worldedit.tool.superpickaxe.enabled.already" : "worldedit.tool.superpickaxe.disabled.already"));
```
octylFractal: I feel like we could benefit from having a top-level `worldedit.feature.already.enabled`/`.disabled`, that takes an argument like `worldedit.tool.superpickaxe`, with English TR like `%s already enabled`. That way we don't have a separate key for each individual tool's already enabled/disabled. Could be applicable to the plain `enabled`/`disabled` messages as well.

worldedit-core/src/main/java/com/sk89q/worldedit/command/WorldEditCommands.java
```diff
@@ -70,25 +73,31 @@ public WorldEditCommands(WorldEdit we) {
...
-                Platform platform = pm.queryCapability(capability);
-                actor.printDebug(String.format("%s: %s", capability.name(), platform != null ? platform.getPlatformName() : "NONE"));
-            } catch (NoCapablePlatformException e) {
-                actor.printDebug(String.format("%s: %s", capability.name(), "NONE"));
-            }
+            Platform platform = pm.queryCapability(capability);
+            producer.append(
+                    TextComponent.of(capability.name(), TextColor.GRAY)
+                    .append(TextComponent.of(": ")
+                    .append(TextComponent.of(platform != null ? platform.getPlatformName() : "NONE")))
```
octylFractal: Do we need a key for `NONE`?

worldedit-core/src/main/java/com/sk89q/worldedit/command/WorldEditCommands.java
```diff
@@ -70,25 +73,31 @@ public WorldEditCommands(WorldEdit we) {
         desc = "Get WorldEdit version"
     )
     public void version(Actor actor) {
-        actor.print("WorldEdit version " + WorldEdit.getVersion());
-        actor.print("https://github.com/EngineHub/worldedit/");
+        actor.printInfo(TranslatableComponent.of("worldedit.version.version", TextComponent.of(WorldEdit.getVersion())));
+        actor.printInfo(TextComponent.of("https://github.com/EngineHub/worldedit/"));
```
octylFractal: Mind fixing capitalization on the repository name?

worldedit-core/src/main/java/com/sk89q/worldedit/command/WorldEditCommands.java
```diff
@@ -70,25 +73,31 @@ public WorldEditCommands(WorldEdit we) {
...
-                actor.printDebug(String.format("%s: %s", capability.name(), "NONE"));
-            }
+            Platform platform = pm.queryCapability(capability);
+            producer.append(
+                    TextComponent.of(capability.name(), TextColor.GRAY)
+                    .append(TextComponent.of(": ")
+                    .append(TextComponent.of(platform != null ? platform.getPlatformName() : "NONE")))
+            ).newline();
         }
+        actor.print(new MessageBox("Capabilities", producer, TextColor.GRAY).create());
```
octylFractal: Should this message box title get a TR key as well?

worldedit-core/src/main/java/com/sk89q/worldedit/function/factory/Deform.java
```diff
@@ -186,9 +192,11 @@ public Operation resume(RunContext run) throws WorldEditException {
         public void cancel() {
         }
 
+
         @Override
-        public void addStatusMessages(List<String> messages) {
-            messages.add("deformed using " + expression);
+        public Iterable<Component> getStatusMessages() {
+            return Lists.newArrayList(TranslatableComponent.of("worldedit.operation.deform.expression",
```
octylFractal: Would prefer `ImmutableList.of` here, and in other `getStatusMessages` as well.

worldedit-core/src/main/java/com/sk89q/worldedit/function/operation/Operation.java
```diff
@@ -56,7 +60,22 @@
...
+    }
 
+    /**
+     * Gets an iterable of messages that describe the current status of the
+     * operation.
+     *
+     * @return The status messages
+     */
+    default Iterable<Component> getStatusMessages() {
+        // TODO Remove legacy code WorldEdit 8.0.0
```
octylFractal: Can we add a one-time-deprecation-warning here? Just something to show it in the server logs of users a little bit too.

e.g. "An operation is using the old status message API: \<class name>", and just add the class to a set to skip the message on next calls.

worldedit-core/src/main/java/com/sk89q/worldedit/util/PropertiesConfiguration.java
```diff
@@ -119,6 +120,12 @@ public void load() {
         allowSymlinks = getBool("allow-symbolic-links", allowSymlinks);
         serverSideCUI = getBool("server-side-cui", serverSideCUI);
         extendedYLimit = getBool("extended-y-limit", extendedYLimit);
+        defaultLocaleName = getString("default-locale", defaultLocaleName);
+        if (defaultLocaleName.equals("default")) {
```
octylFractal: Shouldn't we extract this if-default-else-tag logic to a function in LocalConfiguration?

Can just do something like `setDefaultLocaleFromName(defaultLocaleName)` then.

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/ResourceLoader.java
```diff
@@ -27,16 +27,31 @@
...
+    }
+
+    public static URL getResourceRoot(String name) throws IOException {
+        URL url = ResourceLoader.class.getResource("/" + name);
+        if (url == null) {
+            try {
+                return getResourceForgeHack(name);
             } catch (Exception e) {
-                // Not forge.
+                throw new IOException("Could not find " + name);
```
octylFractal: Move the try-catch into the `getResourceForgeHack`?

worldedit-core/src/main/resources/lang/.gitignore
```diff
@@ -0,0 +1 @@
+*/
```
octylFractal: This seems out of place. What's the issue with putting this rule in the base gitignore?

worldedit-core/src/main/java/com/sk89q/worldedit/util/translation/TranslationManager.java
```diff
@@ -0,0 +1,159 @@
...
+ * component = The part of the plugin, eg expand
+ * message = A descriptor for which message, eg, expanded
+ * meta = Any extra information such as plural/singular (Can have none to infinite)
+ */
+public class TranslationManager {
+
+    private static final Gson gson = new GsonBuilder().create();
+    private static final Type STRING_MAP_TYPE = new TypeToken<Map<String, String>>() {}.getType();
+
+    private final Map<Locale, Map<String, String>> translationMap = new HashMap<>();
```
octylFractal: Does this need to be thread-safe? I don't recall if we have any off-thread chat senders, but I feel like the async paste stuff is.

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgePlayer.java
```diff
@@ -203,6 +204,11 @@ public void setFlying(boolean flying) {
         }
     }
 
+    @Override
+    public Locale getLocale() {
+        return Locale.forLanguageTag(player.language.replace('_', '-'));
```
octylFractal: Just thinking of this down here, but can we add, maybe to TextUtils or something, this language mapping function? I have a hunch it will need fixing/tuning in the future, and I'd rather not replace the ~4 instances of it :)

When I say this I mean the whole mapping from `player.language` to a `Locale`, so it would include `forLanguageTag` in the function body.

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
me4502: So IMO at this point, a second pass makes the most sense, to try and get the base translations infrastructure in as soon as possible.

MattBDev: I completely agree with Matt on this one.

worldedit-core/src/main/java/com/sk89q/worldedit/function/operation/Operation.java
```diff
@@ -56,7 +63,29 @@
...
      *
      * @param messages The list to add messages to
+     * @deprecated Will be removed in WorldEdit 8.0 - use the Component variant
      */
-    void addStatusMessages(List<String> messages);
+    @Deprecated
+    default void addStatusMessages(List<String> messages) {
+    }
 
+    Set<String> warnedDeprecatedClasses = new HashSet<>();
```
octylFractal: Should we add javadoc to mark this as internal? I know people aren't likely to use it, but it is currently public.

worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
me4502: So I had a look at this, and tbh I don't think there are enough instances to really warrant it

PR #481, opened 2019-06-07
Re-add delchunks

wizjany:
Todo:
* [x] Make less hack :hammer_and_wrench: 
* [x] Make code pretty (◕ᴗ◕✿)
* [x] Implement on other platforms
  * [x] Handle DIM-worlds (nested folders) - may need additional methods in World
  * [x] Add startup call

Wishlist:
* [x] Mode to delete by modified time (stored in second 4k of headers)
* ~~[ ] Mode to delete by inhabited time/other params - probably not yet, since it requires reading NBT from within the chunk sections of the region file (possible but a bit further out)~~
* ~~[ ] Full MCRegion abstraction - probably not needed yet, currently just overwriting offsets in the MCR header section~~


worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ChunkDeleter.java
```diff
@@ -0,0 +1,258 @@
...
+        boolean doBackup = args.length > 1 && args[1].equalsIgnoreCase("true") || promptDoBackup();
+        String chunkFile = args[0];
+
+        ChunkDeleter cd = new ChunkDeleter(doBackup);
+        if (!cd.loadChunkFile(chunkFile)) {
+            System.exit(1);
+        }
+        cd.readChunksFromFile();
+
+        System.out.print("Press [enter] to begin deletion...");
```
kenzierocks: Perhaps "any key" would be better? Currently that's the actual criteria.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ChunkDeleter.java
```diff
@@ -0,0 +1,258 @@
...
+            logger.warn("Couldn't load chunk deletion list. Skipping chunk deletion this time.");
+        }
+    }
+
+    private static boolean promptDoBackup() {
+        Scanner scanner = new Scanner(System.in);
+        Boolean backup = null;
+        System.out.print("Would you like to create backups of the region files before deleting chunks? [Y/n] ");
+        while (backup == null) {
+            String input = scanner.next();
```
kenzierocks: I think you want `nextLine()`.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ChunkDeleter.java
```diff
@@ -0,0 +1,258 @@
...
+        return backup;
+    }
+
+    public ChunkDeleter(boolean doBackup) {
+        this.backup = doBackup;
+    }
+
+    private boolean backup;
+    private List<String> chunks;
+    private Map<String, Map<String, Set<BlockVector2>>> regionMap;
```
kenzierocks: Perhaps a `Table<String, String, Set<>>` would be nicer?

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ChunkDeleter.java
```diff
@@ -0,0 +1,258 @@
...
+            Map<String, Set<BlockVector2>> regions = entry.getValue();
+            backupRegions(world, regions.keySet());
+        }
+    }
+
+    private static void backupRegions(String world, Set<String> regions) throws IOException {
+        Path worldDir = Paths.get(world, "region");
+        for (String region : regions) {
+            Path mca = worldDir.resolve(region);
+            if (mca.toFile().exists()) {
```
kenzierocks: What's wrong with `Files.exists`?

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ChunkDeleter.java
```diff
@@ -0,0 +1,258 @@
...
+            return false;
+        }
+    }
+
+    private static void deleteChunks(String world, Map<String, Set<BlockVector2>> regions) {
+        Path worldDir = Paths.get(world, "region");
+        for (Map.Entry<String, Set<BlockVector2>> entry : regions.entrySet()) {
+            String region = entry.getKey();
+            Path mca = worldDir.resolve(region);
+            if (mca.toFile().exists()) {
```
kenzierocks: Same here.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ChunkDeleter.java
wizjany: I was thinking about this but I think we depend on the server classpath in all cases for guava to exist (meaning in "offline" mode we'd need to add the server jar to the sh/bat's classpath arg)

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ChunkDeleter.java
wizjany: Need to test this probably since the last time I used Scanner (like 6 years ago?) I think I remember something about it tokenizing input, so it wouldn't work until the user types a full word. /shrug maybe tho

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
```diff
@@ -109,6 +111,11 @@ public void onLoad() {
         // Setup platform
         server = new BukkitServerInterface(this, getServer());
         worldEdit.getPlatformManager().register(server);
+
+        final File delChunks = Paths.get(getDataFolder().getPath(), "startup_delete_chunks.txt").toFile();
```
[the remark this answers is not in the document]
wizjany: Yup, need to get used to the "new" API. A path is *so* abstract it doesn't have `#exists`, looks like that's in `Files` though.

wizjany:
wrt sys.out/err/logger, the idea is that user interaction in "offline" mode is done through sys in/out/err. That should just be in the static main and promptbackup methods though. I'll make sure that's the case. everything else should use logger. (tho atm I am annoyed by the fact that the logger isn't prefixing messages. need to figure out how this works)

wizjany:
Seems to work now. Might re-add the offline script as an option.
Other thing I was thinking about is the ability to cleanup the backup files. Obviously don't want to do it automatically as the user might want to revert changes, but atm they will clutter for a while.
On the other hand, doesn't hurt to have backups :^)

worldedit-core/src/main/java/com/sk89q/worldedit/internal/anvil/ChunkDeleter.java
```diff
@@ -0,0 +1,327 @@
...
+            .create();
+
+    public static ChunkDeletionInfo readInfo(Path chunkFile) throws IOException, JsonSyntaxException {
+        String json = new String(Files.readAllBytes(chunkFile), StandardCharsets.UTF_8);
+        return chunkDeleterGson.fromJson(json, ChunkDeletionInfo.class);
+    }
+
+    public static void writeInfo(ChunkDeletionInfo info, Path chunkFile) throws IOException, JsonIOException {
+        String json = chunkDeleterGson.toJson(info, new TypeToken<ChunkDeletionInfo>() {}.getType());
+        try (BufferedWriter writer = Files.newBufferedWriter(chunkFile, StandardOpenOption.CREATE)) {
```
kenzierocks: Currently this will make the command fail with "File already exists", which doesn't really mean anything to the user. Perhaps we can check if the file exists or catch the appropriate exception, and inform the user that "There is already a scheduled chunk deletion"? Or merge the two deletions?

kenzierocks:
I'm unsure if we should leave the commented script code in -- I didn't review it again, so uncommenting it in the future would potentially lead to problems. I would prefer we simply add it as a separate PR if we want, and remove it from the current PR.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/anvil/ChunkDeleter.java
wizjany: It should (and from my testing, does) overwrite the current contents of the file with the new contents (which is any previously existing batches + the new batch).

PR #482, opened 2019-06-09
Update to Forge 1.14.

wizjany:
And more misc fixes.

~~And expanded the scope of the item brush.~~


worldedit-core/src/main/java/com/sk89q/worldedit/command/ApplyBrushCommands.java
```diff
@@ -111,11 +116,19 @@ public void forest(CommandParameters parameters,
...
     public void item(CommandParameters parameters,
                      Player player, LocalSession localSession,
                      @Arg(desc = "The type of item to use")
-                         BaseItem item) throws WorldEditException {
-        setApplyBrush(parameters, player, localSession, new ItemUseFactory(item));
+                         BaseItem item,
+                     @Arg(desc = "The direction to face applying the item", def = "up")
+                     @Direction(includeDiagonals = true)
+                         BlockVector3 direction) throws WorldEditException {
+        player.print(TextComponent.of("Warning: ", TextColor.RED)
```
kenzierocks: Any reason this isn't all one message? All the text is going to be red regardless, right?

worldedit-core/src/main/java/com/sk89q/worldedit/command/ApplyBrushCommands.java
```diff
@@ -111,11 +116,19 @@ public void forest(CommandParameters parameters,
...
+                .append(TextComponent.of("This brush simulates item usages. Its effects may not work on all platforms, " +
+                        "may not be undoable, and may cause strange interactions with other mods/plugins. Use at your own risk.")));
+        setApplyBrush(parameters, player, localSession, new ItemUseFactory(item,
+                com.sk89q.worldedit.util.Direction.findClosest(direction.toVector3(), com.sk89q.worldedit.util.Direction.Flag.ALL)));
```
kenzierocks: Probably should just add a new converter in `DirectionConverter` that can handle `@Direction Direction`. Though that may not look great as a parameter. Maybe we should rename the `@Direction` annotation?

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeItemRegistry.java
```diff
@@ -29,6 +34,14 @@
     @Nullable
     @Override
     public String getName(ItemType itemType) {
-        return super.getName(itemType); // TODO
+        if (FMLLoader.getDist().isClient()) {
+            final Item item = RegistryManager.ACTIVE.getRegistry(Item.class)
+                    .getValue(ResourceLocation.tryCreate(itemType.getId()));
+            if (item == null) {
```
kenzierocks: Perhaps swap the branches here, so you can drop the repeated return?

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorldEdit.java
```diff
@@ -135,20 +134,10 @@ private void load(FMLLoadCompleteEvent event) {
         if (FMLLoader.getDist() == Dist.CLIENT) {
             // we want to setup platform before we hit the main menu
             // but this event is async -- so we must delay until the first game loop:
-            Minecraft.getInstance().addScheduledTask(this::setupPlatform);
+            //Minecraft.getInstance().enqueue(this::setupPlatform);
```
kenzierocks: If you're going to drop this for sure, I would remove the surrounding method as well. If it's working without it, might as well remove it all.

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/net/handler/WECUIPacketHandler.java
```diff
@@ -70,11 +71,17 @@ public static void onPacketData(ServerCustomPayloadEvent event) {
...
+            new SCustomPayloadPlayPacket(
                     new ResourceLocation(ForgeWorldEdit.MOD_ID, ForgeWorldEdit.CUI_PLUGIN_CHANNEL),
                     event.getPayload()
-            ).processPacket(Minecraft.getInstance().player.connection);
+            ).processPacket(Minecraft.getInstance().player.field_71174_a);
         } catch (ThreadQuickExitException ignored) {
         }
     }
+
+    private static boolean validate(String p) {
```
kenzierocks: Can we share these validate methods?

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgePlayer.java
```diff
@@ -183,18 +188,34 @@ public boolean hasPermission(String perm) {
...
+            } catch (IOException e) {
+                return;
+            }
+            player.connection.sendPacket(packetOut);
+            if (block instanceof BaseBlock) {
+                final BaseBlock baseBlock = (BaseBlock) block;
+                final CompoundTag nbtData = baseBlock.getNbtData();
+                if (nbtData != null) {
+                    player.connection.sendPacket(new SUpdateTileEntityPacket(
+                            new BlockPos(pos.getBlockX(), pos.getBlockY(), pos.getBlockZ()), 7,
```
kenzierocks: Can we put that `7` in a constant field, named `STRUCTURE_BLOCK_TILE_ENTITY_ID` or something? Honestly, we should be careful with this, because `7` won't work if we start doing non-structure blocks. Perhaps we should have a map of IDs?

worldedit-forge/build.gradle
```diff
@@ -57,7 +57,7 @@ minecraft {
         }
     }
 
-    accessTransformer = file('worldedit_at.cfg')
+    //accessTransformer = file('worldedit_at.cfg')
```
kenzierocks: Can we just drop the AT configuration for now? If we end up needing in in 1.15, etc. we can add it back. It doesn't seem worth of keeping it tracked while commented.

PR #486, opened 2019-06-10
[1.12.2] Implement regenerate for Sponge

clienthax:


worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorld.java
```diff
@@ -142,6 +152,43 @@ public boolean setBlock(Vector position, BaseBlock block, boolean notifyAndLight
...
+            try {
+                tempWorldProperties = server.createWorldProperties("worldedittemp", this.regenerationArchetype);
+            } catch (IOException e) {
+                e.printStackTrace();
+                return false;
+            }
+
+            Optional<World> tempWorldOpt = server.loadWorld(tempWorldProperties);
+            if (!tempWorldOpt.isPresent()) {
+                return false;
```
kenzierocks: Might want to include a log line here, just so it's easy to know why regen isn't working.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorld.java
```diff
@@ -142,6 +152,43 @@ public boolean setBlock(Vector position, BaseBlock block, boolean notifyAndLight
...
+                freshWorld.getChunk(chunk.getBlockX(), 0, chunk.getBlockZ());
+            }
+
+            SpongeWorld from = SpongeWorldEdit.inst().getWorld(freshWorld);
+            for (BlockVector vec : region) {
+                editSession.setBlock(vec, from.getBlock(vec));
+            }
+
+            // Remove temp world
+            server.unloadWorld(freshWorld);
```
kenzierocks: Might want to put this (and following line) in a `finally` clause, in case the `MaxChangedBlocksException` occurs.

clienthax:
Changes done

wizjany:
Hm, is it possible to prevent sponge from loading spawn? failing that entirely, maybe can we change the world spawn to be inside the area we want to regen? This is noticeably slower than the forge "equivalent" version atm. Seems to work though.

wizjany:
Also, can we move the archetype building down to the regen method when it's needed? This class is meant to be a one-time use wrapper around a world. we don't have one object for each world that just hangs around, so we're calling that archetype potentially hundreds or thousands of times needlessly.

clienthax:
> Hm, is it possible to prevent sponge from loading spawn? failing that entirely, maybe can we change the world spawn to be inside the area we want to regen? This is noticeably slower than the forge "equivalent" version atm. Seems to work though.

I have attempted this with the last push, however it doesn't seem that sponge is obeying the world properties setGenerateSpawnOnLoad value currently, taking a closer look at it,
It also seems that the SerializationBehaviors.NONE is not honored either as it still writes files to the world folder.

clienthax:
> Also, can we move the archetype building down to the regen method when it's needed? This class is meant to be a one-time use wrapper around a world. we don't have one object for each world that just hangs around, so we're calling that archetype potentially hundreds or thousands of times needlessly.

Aha, i couldn't quite figure out if this was the case or not, I have now moved it into the regen method.

wizjany:
LGTM now. Realistically this is probably ok to pull now since you do manually cleanup to workaround the sponge issues, so while those would be nice to have, this works ok. (deleteWorld won't throw if there's nothing to delete right?)

clienthax:
> LGTM now. Realistically this is probably ok to pull now since you do manually cleanup to workaround the sponge issues, so while those would be nice to have, this works ok. (deleteWorld won't throw if there's nothing to delete right?)

I think it will throw, but sponge should handle it in the form of a nasty log message
https://github.com/SpongePowered/SpongeCommon/blob/73bd5a690228331902862e20115ce9c841e68b70/src/main/java/org/spongepowered/common/world/WorldManager.java#L1106

wizjany:
I mean when `SerializationBehaviors.NONE` is respected, there won't be files on disk to delete right?
WorldProperties shouldn't be null.

clienthax:
> I mean when `SerializationBehaviors.NONE` is respected, there won't be files on disk to delete right?
> WorldProperties shouldn't be null.

I wouldn't imagine so, as the world will still technically exist in memory, delete should just clear it from memory / being available

wizjany:
merged (I changed return to true on success, as the docs specify)

PR #487, opened 2019-06-11
//naturalize improvements

DarkArc:
The naturalizer for as long as I can remember has reported every block being changed, this corrects the behavior to report the correct number of block changes.

Additionally, I've changed naturalize to work more like a replace, which prevents excessive buildup of history records, and allows naturalize to be used on much larger areas for terraforming.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/block/Naturalizer.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/block/Naturalizer.java
@@ -27,6 +27,7 @@ import com.sk89q.worldedit.function.LayerFunction;
 import com.sk89q.worldedit.function.mask.BlockTypeMask;
 import com.sk89q.worldedit.function.mask.Mask;
 import com.sk89q.worldedit.math.BlockVector3;
+import com.sk89q.worldedit.world.block.BlockState;
 import com.sk89q.worldedit.world.block.BlockTypes;
 
 /**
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/block/Naturalizer.java
```diff
@@ -65,21 +66,35 @@ public boolean isGround(BlockVector3 position) {
...
+            case 1:
+            case 2:
+            case 3:
+                return BlockTypes.DIRT.getDefaultState();
+            default:
+                return BlockTypes.STONE.getDefaultState();
+        }
+    }
+
+    private boolean makeChangeTo(BlockVector3 position, int depth) throws WorldEditException {
```
kenzierocks: Maybe call this `naturalize`?

worldedit-core/src/main/java/com/sk89q/worldedit/function/block/Naturalizer.java
```diff
@@ -65,21 +66,35 @@ public boolean isGround(BlockVector3 position) {
...
-                    editSession.setBlock(position, BlockTypes.GRASS_BLOCK.getDefaultState());
-                    break;
-                case 1:
-                case 2:
-                case 3:
-                    editSession.setBlock(position, BlockTypes.DIRT.getDefaultState());
-                    break;
-                default:
-                    editSession.setBlock(position, BlockTypes.STONE.getDefaultState());
+            if (makeChangeTo(position, depth)) {
```
kenzierocks: Combine with previous if statement? Feels a little more appropriate to me.

PR #492, opened 2019-06-27
Fix painting offset and broken rotation on copy/paste operations

Brokkonaut:
Fixes WORLDEDIT-3429

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
@@ -35,6 +35,7 @@ import com.sk89q.worldedit.math.transform.Transform;
 import com.sk89q.worldedit.util.Direction;
 import com.sk89q.worldedit.util.Direction.Flag;
 import com.sk89q.worldedit.util.Location;
+import com.sk89q.worldedit.world.entity.EntityTypes;
 
 /**
  * Copies entities provided to the function to the provided destination
@@ -91,9 +92,18 @@ public class ExtentEntityCopy implements EntityFunction {
         if (state != null) {
             Location newLocation;
             Location location = entity.getLocation();
+            // If the entity has stored the location in the NBT data, we use that location
+            CompoundTag tag = state.getNbtData();
+            boolean hasTilePosition = tag != null && tag.containsKey("TileX") && tag.containsKey("TileY") && tag.containsKey("TileZ");
+            if (hasTilePosition) {
+                location = location.setPosition(Vector3.at(tag.asInt("TileX"), tag.asInt("TileY"), tag.asInt("TileZ")).add(0.5, 0.5, 0.5));
+            }
 
             Vector3 pivot = from.round().add(0.5, 0.5, 0.5);
             Vector3 newPosition = transform.apply(location.toVector().subtract(pivot));
+            if (hasTilePosition) {
+                newPosition = newPosition.subtract(0.5, 0.5, 0.5);
+            }
             Vector3 newDirection;
 
             newDirection = transform.isIdentity() ?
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
```diff
@@ -142,14 +155,15 @@ private BaseEntity transformNbtData(BaseEntity state) {
                         .putInt("TileZ", newTilePosition.getBlockZ());
 
                 if (hasFacing) {
-                    Direction direction = MCDirections.fromHanging(tag.asInt("Facing"));
+                    boolean isPainting = state.getType() == EntityTypes.PAINTING; // Paintings are very special
+                    Direction direction = isPainting ? MCDirections.fromPre13Hanging(tag.asInt("Facing")) : MCDirections.fromHanging(tag.asInt("Facing"));
```
kenzierocks: We need to fix this for item frames as well. They use the same horizontal-hanging value.

kenzierocks: Also, it would probably be a good idea to rename it from `Pre13Hanging` to `HorizontalHanging`, since that's what it actually is.

worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
```diff
@@ -104,6 +106,17 @@ public boolean apply(Entity entity) throws WorldEditException {
             // Some entities store their position data in NBT
             state = transformNbtData(state);
 
+            // Paintings are very special
+            if (state.getType() == EntityTypes.PAINTING) {
```
kenzierocks: Why do you need this? It doesn't make a ton of sense to me, since the Tile position was just updated to where the location should be. If this is inconsistent, we should probably change how `newLocation` is calculated, not read it out of NBT data we just wrote.

worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
Brokkonaut: No item frames do not use those values. Just paintings still use the old values,

worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
Brokkonaut: We need this, because getLocation for (vanilla) paintings is inconsistent. The calculation is perfectly fine, just Mojang adds some strange offset. When setting the location of a painting, it updates TileXYZ to the values set and then updates the location to something different.

This is a special thing that just exists for paintings.

kenzierocks: Alright, then:

>we should probably change how newLocation is calculated, not read it out of NBT data we just wrote.

Brokkonaut: I have no idea how I should do that. Only the nbt data contains the correct values.

kenzierocks: Just re-use the existing code that makes the NBT values in `transformNbtData`? If that's not correct, then this won't be either. We haven't written back to MC yet, so it can't have made changes.

worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
kenzierocks: Hmm, it looks like item frames override it. But Leash Knots do not, don't know if we want to handle those @me4502 @wizjany

wizjany: There's a chance mods add entities which extend these (paintings/item frames) and have the same caveats. Unfortunately beyond having some entity property that tells us whether a Hanging entity uses a 4- or 6- value Facing, we probably can't handle all those cases.
That said, we *do* have isPainting and isItemFrame, so perhaps we should prefer those over `== EntityTypes.XYZ` in case mods do extend. (If they change behavior, we will be wrong either way - I think checking extends is "safer").

worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
Brokkonaut: The code uses the old NBT values to transform them. I could use the old NBT values and transform them again the same way as in `transformNbtData`, but would that be any better than this change?

Brokkonaut: Ok, I changed that. Is it ok now?

worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
Brokkonaut: There is no way instead of checking the type, because ClipboardEntities are not faceted.

For leash knots I don't know - currently they cannot be correctly pasted at all, so I cannot test that. And I do not think it is in the scope of this PR to fix that.

PR #495, opened 2019-06-29
Pass through wall improvements

DarkArc:
This change simplifies the algorithm greatly.

Additionally, this fixes a bug where if standing in a non-solid block
i.e. a glass pane, //thru, and the nav wand would not work.


wizjany:
Hm, this seems to have a lot of issues: it puts me on top of walls, I can no longer pass through the ground...

https://gfycat.com/HatefulMadLabradorretriever

without PR:

https://gfycat.com/ScrawnyPlushChinesecrocodilelizard

DarkArc:
That's really strange I'll look into this more. Might be a small issue from updating it to WE 7. I don't experience this when I use it.

PR #498, opened 2019-07-04
Buffered extent improvements

kenzierocks:
This PR fixes problems with extents that buffer `setBlock` calls for later. I'm unsure if changing the EditSession to return blocks via the extents, rather than the world, breaks anything.


kenzierocks:
That issue should be fixed now.

PR #499, opened 2019-07-05
First attempt at fixing quoted string oddities.

wizjany:
This fixes single word quoted arguments vanishing (e.g. ``//set "stone"`` would just be parsed as ``//set ``).

This partially fixes suggestions for quoted strings:
``//gmask "#sel #`` will suggest
 (note: missing the leading quotation mark now)
but then
https://github.com/EngineHub/WorldEdit/blob/f2f9c266029123869dfd04b1fc9220dbcadccd77/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformCommandManager.java#L589-L601
converts it to
 (is this right?)
which, when it gets to here: https://github.com/EngineHub/WorldEdit/blob/3173e2610939f5331d1b035019b0ea4d975680dc/worldedit-core/src/main/java/com/sk89q/worldedit/internal/command/CommandUtil.java#L80-L93
returns empty


kenzierocks:
All suggestions should be in working order now.

PR #501, opened 2019-07-10
Update to Gradle 5

kenzierocks:
This ensures we keep up-to-date on Gradle releases for faster & smarter builds. Additionally, Gradle 5+ has better Kotlin DSL support, which is easier to write and clearer to read.

Note: previously ForgeGradle played rough with Gradle 5. I didn't encounter problems, so I presume they have fixed the issues.


gradle.properties
```diff
@@ -1,4 +1 @@
-# Sets default memory used for gradle commands. Can be overridden by user or command line properties.
-# This is required to provide enough memory for the Minecraft decompilation process.
-#org.gradle.jvmargs=-Xmx3G
-org.gradle.daemon=false
\ No newline at end of file
+org.gradle.jvmargs=-Xmx1G
```
MattBDev: Will this be enough memory for Forge to decompile MC?

MattBDev:
I believe all issues with Gradle 5 and ForgeGradle were resolved with this commit MinecraftForge/ForgeGradle@dcc31a519ea652259648c8338311d5091b7b9257
I hope this is merged soon.

gradle.properties
wizjany: as you can see from the previously-commented-out 3G arg, it became unnecessary ages ago when FG re-did their process. The issue is now with loom.

kenzierocks: Yes, this is more a consequence of the Gradle 5 update -- the daemon now uses `-Xmx64m` by default, rather than letting the VM pick max memory (which is usually ~1g). This value was determined by running the build until Gradle stopped printing memory errors.

PR #503, opened 2019-07-11
Kotlin DSL Gradle Scripts

kenzierocks:
Changes all DSL scripts to Kotlin, instead of Groovy. This is easier to write + read because of the static typing provided by the DSL, as well as improved autocomplete.

Dependent on #501.

Tested platforms:
- [x] Bukkit
- [x] Fabric
- [x] Forge
- Sponge will have to be tested when we update it.


kenzierocks:
Note: I'm currently unsure if the produced jars are even valid. Testing will be needed before this can be merged.

PR #504, opened 2019-07-15
Switch to JUnit 5

kenzierocks:
Final build system changes, I promise 🙂.

Biggest difference is `assertX(message, params)` is now `assertX(params, message)`, and the use of `assertThrows` instead of `try { ... fail(); } catch (X x) {}`.

### DIFF
--- a/worldedit-core/src/test/java/com/sk89q/worldedit/internal/expression/ExpressionTest.java
+++ b/worldedit-core/src/test/java/com/sk89q/worldedit/internal/expression/ExpressionTest.java
@@ -19,16 +19,6 @@
 
 package com.sk89q.worldedit.internal.expression;
 
-import static java.lang.Math.atan2;
-import static java.lang.Math.sin;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-
-import org.junit.Before;
-import org.junit.Test;
-import org.mockito.Mockito;
-
 import com.sk89q.worldedit.LocalConfiguration;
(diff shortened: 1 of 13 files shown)


worldedit-core/src/test/java/com/sk89q/worldedit/internal/expression/ExpressionTest.java
```diff
@@ -181,13 +171,11 @@ public void testQuery() throws Exception {
...
-            simpleEval("for(i=0;i<256;i++){for(j=0;j<256;j++){for(k=0;k<256;k++){for(l=0;l<256;l++){ln(pi)}}}}");
-            fail("Loop was not stopped.");
-        } catch (EvaluationException e) {
-            assertTrue(e.getMessage().contains("Calculations exceeded time limit"));
-        }
+    public void testTimeout() {
+        EvaluationException e = assertThrows(EvaluationException.class,
+            () -> simpleEval("for(i=0;i<256;i++){for(j=0;j<256;j++){for(k=0;k<256;k++){for(l=0;l<256;l++){ln(pi)}}}}"),
+            "Loop was not stopped.");
+        assertTrue(e.getMessage().contains("Calculations exceeded time limit"));
```
wizjany: Not your change, but just noticed this should probably be changed to ExpressionTimeoutException instead of depending on the message.

kenzierocks: Narrowed it, but I still like asserting on the message, to ensure we don't get a different timeout of some sort.

PR #505, opened 2019-07-16
Memory optimizations

kenzierocks:
Long-running PR to collect as many memory optimizations as we can reasonably perform to WorldEdit.

Current features:
- `LocatedBlock` overhead in `LocatedBlockList` removed, just stores the corresponding fields
- `BlockMap` -- new Map implementation designed to provide low overhead when faced with large amounts of block entries, rather than using power of 2 tables.
- `PositionList` interface with support for full Y range if requested, and a slightly more optimized long variant if not.


worldedit-core/src/main/java/com/sk89q/worldedit/util/collection/LocatedBlockList.java
```diff
@@ -19,74 +19,101 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Wrapper around a list of blocks located in the world.
  */
 public class LocatedBlockList implements Iterable<LocatedBlock> {
 
-    private final Map<BlockVector3, LocatedBlock> map = new LinkedHashMap<>();
+    private final BlockMap blocks = BlockMap.create();
+    private final LongArrayList order = new LongArrayList();
```
Barteks2x: While some parts of the new code (`BlockMap`) appear to take possibility of taller worlds into account, this one doesn't. Is this something that will continue to work and this will be changed later, or is this going to limit world height to +-2048 blocks?

kenzierocks: There will be a config option to transform this from a `0-4095` range to normal integer limits, at the price of more memory. 

Also, if you think that storing the Y as signed would be better than unsigned, I can do that instead. I just assumed unsigned to begin with.

Barteks2x: Vanilla `BlockPos.toLong` is signed so I assumed this is the same way here. But it doesn't matter too much.

worldedit-core/src/main/java/com/sk89q/worldedit/util/collection/VectorPositionList.java
```diff
@@ -0,0 +1,52 @@
...
+import com.sk89q.worldedit.math.BlockVector3;
+import it.unimi.dsi.fastutil.objects.ObjectArrayList;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+
+class VectorPositionList implements PositionList {
+
+    private final List<BlockVector3> delegate = new ObjectArrayList<>();
```
Barteks2x: Maybe an IntList and just keep triples of ints? That should be only 50% more memory than LongList, instead of over twice as much.

kenzierocks: Done.

PR #506, opened 2019-07-16
don't buffer move unnecessarily (untested)

boy0001:
something like this


me4502:
This needs rebasing on master, and also needs to be tested (wiz said the implementation is incorrect).

kenzierocks:
Closing unless @boy0001 requests it be re-opened for conflict resolution + a fix to the algorithm.

PR #508, opened 2019-07-21
feature(cli): Added a CLI version of WorldEdit, and allowed most commands to be run from console

me4502:
TODO:

* [x] Add a //world command to override the world
* [x] General cleanup and UX improvements to the CLI worldedit
* [x] CommandBlock support, they have an inherent position that we can use

This is gonna be a painful merge with translations - so imma put that on hold until this is in.


TomyLobo:
Would ClipboardWorlds be useful for in-game editing, too?

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
```diff
@@ -885,25 +903,38 @@ public Calendar detectDate(String input) {
...
-    public EditSession createEditSession(Player player) {
-        checkNotNull(player);
+    public EditSession createEditSession(Actor actor) {
+        checkNotNull(actor);
 
-        BlockBag blockBag = getBlockBag(player);
+        World world = null;
+        if (hasWorldOverride()) {
+            world = getWorldOverride();
+        } else if (actor.isPlayer() && actor instanceof Player) {
```
kenzierocks: This seems redundant? Is there a case where we have "Player" but not "isPlayer"?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformCommandManager.java
```diff
@@ -222,30 +223,58 @@ private void registerArgumentConverters() {
...
-                        }
-                    });
-            });
+                context -> {
+                    LocalSession localSession = context.injectedValue(Key.of(LocalSession.class))
+                            .orElseThrow(() -> new IllegalStateException("No LocalSession"));
+                    return context.injectedValue(Key.of(Actor.class))
+                            .map(actor -> {
+                                try {
+                                    World world;
```
kenzierocks: This should re-use the world resolved below, by doing `context.injectedValue(Key.of(World.class))`. That way we don't duplicate code.

kenzierocks: And, you can probably remove the Actor injected value too.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformCommandManager.java
```diff
@@ -222,30 +223,58 @@ private void registerArgumentConverters() {
...
+                                    if (localSession.hasWorldOverride()) {
+                                        return localSession.getWorldOverride();
+                                    } else if (actor.isPlayer() && actor instanceof Player) {
+                                        return ((Player) actor).getWorld();
+                                    } else {
+                                        throw new MissingWorldException();
+                                    }
+                                } catch (MissingWorldException e) {
+                                    exceptionConverter.convert(e);
+                                    throw new AssertionError("Should have thrown a new exception.");
```
kenzierocks: Maybe include `e` as the cause here?

me4502:
> Would ClipboardWorlds be useful for in-game editing, too?

/Possibly/, depends how we plan on doing things. If we allow //world to act on the clipboard, technically it could be useful.

worldedit-cli/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="WARN" packages="net.minecraft,com.mojang">
```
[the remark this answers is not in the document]
me4502: This is the log4j2.xml from Minecraft itself, I'll change the packages

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+package com.sk89q.worldedit.cli;
+
+import com.sk89q.worldedit.util.PropertiesConfiguration;
+
+import java.io.File;
+
+public class CLIConfiguration extends PropertiesConfiguration {
+
+    public CLIConfiguration(CLIWorldEdit mod) {
+        super(new File(mod.getWorkingDir() + File.separator + "worldedit.properties"));
```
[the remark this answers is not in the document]
me4502: Where I copy pasted it from didn't use that :P

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
me4502: Honestly I should just make a LocatableActor interface that Player implements - as CommandBlocks are locatable as well.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
```diff
@@ -0,0 +1,158 @@
...
+    public void printRaw(String msg) {
+        for (String part : msg.split("\n")) {
+            sender.sendMessage(part);
+        }
+    }
+
+    @Override
+    public void print(String msg) {
+        for (String part : msg.split("\n")) {
+            sender.sendMessage("\u00A7d" + part);
```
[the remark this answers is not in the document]
kenzierocks: Or probably even better, wrap them in the appropriate component, and delegate to `print(Component)`.

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLICommandSender.java
```diff
@@ -0,0 +1,164 @@
...
+
+import java.io.File;
+import java.util.UUID;
+
+public class CLICommandSender implements Actor {
+
+    /**
+     * One time generated ID.
+     */
+    private static final UUID DEFAULT_ID = UUID.fromString("a233eb4b-4cab-42cd-9fd9-7e7b9a3f74be");
```
kenzierocks: Might be a good idea to generate this as a [Version 5](https://en.wikipedia.org/wiki/Universally_unique_identifier#Versions_3_and_5_(namespace_name-based)) ID, just to completely avoid an unlikely random collision. It makes it truly unique among MC's UUID set. Of course, this probably doesn't matter overall, I just like the idea of embedding the name into the ID.

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLICommandSender.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Override
+    public void printRaw(String msg) {
+        for (String part : msg.split("\n")) {
+            sender.info(part);
+        }
+    }
+
+    public static final String ANSI_PURPLE = "\u001B[35m";
```
kenzierocks: Any reason for these being public?

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorldEdit.java
```diff
@@ -0,0 +1,349 @@
...
+        for (String name : fileRegistries.getDataFile().itemtags.keySet()) {
+            if (ItemCategory.REGISTRY.get(name) == null) {
+                ItemCategory.REGISTRY.register(name, new ItemCategory(name));
+            }
+        }
+    }
+
+    public void onInitialized() {
+        // Setup working directory
+        workingDir = new File("worldedit").toPath();
```
kenzierocks: Would prefer `Paths.get("worldedit")`.

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorldEdit.java
```diff
@@ -0,0 +1,349 @@
...
+    public Platform getPlatform() {
+        return this.platform;
+    }
+
+    /**
+     * Get the working directory where WorldEdit's files are stored.
+     *
+     * @return the working directory
+     */
+    public File getWorkingDir() {
```
kenzierocks: Any reason to not make this return `Path`? It's new API.

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorldEdit.java
```diff
@@ -0,0 +1,349 @@
...
+    public void run(InputStream inputStream) {
+        Scanner scanner = new Scanner(inputStream);
+        while (scanner.hasNextLine()) {
+            String line = scanner.nextLine();
+            if (line.equalsIgnoreCase("stop")) {
+                commandSender.print("Stopping!");
+                break;
+            }
+            if (line.startsWith("save")) {
+                String[] bits = line.split(" ");
```
kenzierocks: I would prefer it if we allowed quoting or escaping here, just for user-friendliness.

worldedit-cli/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,24 @@
...
+            <PatternLayout pattern="[%d{HH:mm:ss}] [%t/%level]: %msg%n" />
+            <Policies>
+                <TimeBasedTriggeringPolicy />
+                <OnStartupTriggeringPolicy />
+            </Policies>
+        </RollingRandomAccessFile>
+    </Appenders>
+    <Loggers>
+        <Root level="info">
+            <filters>
```
kenzierocks: I don't think we need this filter.

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/WorldConverter.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Override
+    public Component describeAcceptableArguments() {
+        return this.choices;
+    }
+
+    @Override
+    public List<String> getSuggestions(String input) {
+        return WorldEdit.getInstance().getPlatformManager()
```
kenzierocks: You should probably extract the world-stream extraction to its own method.

worldedit-core/src/main/java/com/sk89q/worldedit/entity/Entity.java
```diff
@@ -33,7 +33,7 @@
  * instance of an entity, but a {@link BaseEntity} can be created from
  * this entity by calling {@link #getState()}.</p>
  */
-public interface Entity extends Faceted {
+public interface Entity extends Faceted, Locatable {
```
kenzierocks: I'm unsure if this change is binary-compatible. Usually calling e.g. `Entity#getLocation()` will result in:
```asm
invokevirtual com/sk89q/worldedit/entity/Entity.getLocation:()
```
Which would imply to me that it looks for exactly that method at runtime, and if it's in a superclass it won't be found. But perhaps I am wrong.

worldedit-core/src/main/java/com/sk89q/worldedit/entity/Player.java
```diff
@@ -281,13 +285,6 @@
      */
     void setPosition(Vector3 pos, float pitch, float yaw);
 
-    /**
-     * Move the player.
-     *
-     * @param pos where to move them
-     */
-    void setPosition(Vector3 pos);
```
kenzierocks: Same issue as Entity perhaps.

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLICommandSender.java
me4502: Nope

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLICommandSender.java
me4502: This is the same ID we're using for the console in Bukkit & Sponge

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorld.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package com.sk89q.worldedit.cli;
+
+public interface CLIWorld {
+
+    /**
+     * Saves this world back to file.
+     *
+     * @param force Force a save
```
kenzierocks: Under what circumstances does this not save if this is false?

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorldEdit.java
```diff
@@ -0,0 +1,330 @@
...
+                file = app.commandSender.openFileOpenDialog(new String[]{"schem", "dat"});
+            } else {
+                file = new File(fileArg);
+            }
+            if (file == null) {
+                throw new IllegalArgumentException("A file must be provided!");
+            }
+            if (file.getName().endsWith("level.dat")) {
+                throw new IllegalArgumentException("level.dat file support is unfinished.");
+            } else if (file.getName().endsWith(".schem")) {
```
kenzierocks: Maybe handle this generically with format detection?

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorld.java
me4502: When the world isn't dirty

kenzierocks: So, document that? :P

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorldEdit.java
```diff
@@ -0,0 +1,332 @@
...
+
+        InputStream inputStream = System.in;
+
+        try {
+            CommandLine cmd = new DefaultParser().parse(options, args);
+
+            String fileArg = cmd.getOptionValue('f');
+            File file;
+            if (fileArg == null) {
+                file = app.commandSender.openFileOpenDialog(new String[]{"schem", "dat"});
```
kenzierocks: Should be dynamically built from formats now.

me4502: Oh, true

wizjany:
LoggingCommandHandler may need to be updated for these changes too

PR #510, opened 2019-07-28
Make BlockStateHolder extend Pattern.

wizjany:
Removes the need for wrapping them in BlockPattern.

~~todo: mark editsession methods as deprecated, probably?~~

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BaseBlock.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BaseBlock.java
@@ -25,7 +25,6 @@ import com.sk89q.jnbt.CompoundTag;
 import com.sk89q.jnbt.StringTag;
 import com.sk89q.jnbt.Tag;
 import com.sk89q.worldedit.blocks.TileEntityBlock;
-import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.registry.state.Property;
 
 import java.util.Map;
@@ -164,11 +163,6 @@ public class BaseBlock implements BlockStateHolder<BaseBlock>, TileEntityBlock {
 
     @Override
     public BaseBlock toBaseBlock() {
-        return this;
-    }
-
-    @Override
-    public BaseBlock apply(BlockVector3 position) {
         return this;
     }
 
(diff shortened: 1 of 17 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockState.java
```diff
@@ -207,6 +208,11 @@ public BaseBlock toBaseBlock() {
         return this.emptyBaseBlock;
     }
 
+    @Override
+    public BaseBlock apply(BlockVector3 position) {
```
kenzierocks: This override isn't necessary. The overhead of `toBaseBlock()` is negligible.

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BaseBlock.java
```diff
@@ -166,6 +167,11 @@ public BaseBlock toBaseBlock() {
         return this;
     }
 
+    @Override
+    public BaseBlock apply(BlockVector3 position) {
```
kenzierocks: Same here, the default in BSH is fine.

MattBDev:
I know FAWE already does this but what prompted this change in WE?

me4502:
I wonder if it's worth making BlockType a pattern and getting rid of BlockTypePattern too :thinking:

wizjany:
Was actually considering making type return its default state as a pattern. FuzzyBlockState could be a RandomPattern by itself, though I'm not sure that's really useful. BlockTypePattern isn't a thing tho :P

me4502:
Oh I must've deleted BlockTypePattern then oof... That's fair :joy:

But yeah it may make sense for it to return its default state as a pattern, just to make the code neater

wizjany:
Yea there's a lot of getDefaultState going on. One of the issues with BlockType is that if people grab BlockType from BlockTypes, there's a potential for it to be null, which means we need to start null-checking pattern inputs. This hopefully shouldn't be too common tho, but for example `BlockTypes.SIGN` was null before the 1.14 block types update.

me4502:
We could just do a checkNotNull, realistically that's an error on their part if they've grabbed a block that doesn't exist from BlockTypes (Or it's being used on the wrong version)

PR #511, opened 2019-07-30
Better cubic chunks compatibility (negative heights, general height-related fixes)

Barteks2x:
Changes so far:
 * A `default-vertical-size` config option that controls effects of functionality that previously worked over the entire world height
 * `//thaw`, `//snow` and `//green` now take an optional vertical size parameter, defaults to `default-vertical-size`
 * `//chunk` command selects the current `16x256x16` section instead of the whole world height. Optionally can take `x,y,z` instead of `x,z` as coordinates. `y` defaults to 0.
 * `//expand vert` - uses `default-vertical-size` instead of full world height by default, optionally can take size as parameter (I'm not sure if that's the right way to do it, it appears to go a bit against the intention of that command)
 * `//removeabove` and `//removebelow` use `default-vertical-size` instead of world height limit. Optionally, user can provide a value for it.
 * gravity brush will use `default-vertical-size` when using full height requested
 * Removed previous hacky code to get minY. Cubic chunks will transform the `getMinY` method instead.

To be done:
 * More testing, especially without cubic chunks installed
 * Verify that command's help output is still actually helpful
 * Verify that there are no breaking api changes
 * Better name for the config option?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: java
 notifications:
   email: false
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -766,7 +766,7 @@ public class EditSession implements Extent {
         MaskIntersection mask = new MaskIntersection(
                 new RegionMask(new EllipsoidRegion(null, origin, new Vector(radius, radius, radius))),
                 new BoundedHeightMask(
-                        Math.max(origin.getBlockY() - depth + 1, 0),
+                        Math.max(origin.getBlockY() - depth + 1, getWorld().getMinY()),
                         Math.min(getWorld().getMaxY(), origin.getBlockY())),
                 Masks.negate(new ExistingBlockMask(this)));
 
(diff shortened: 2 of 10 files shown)


kenzierocks:
Despite #521, I'll be leaving this open for now as an unofficial 1.12.2 support for CC. Since it targets wiz's CC branch, I'll let them handle merging it if it's desired.

wizjany:
@Barteks2x is there a status update on this? I see that you committed https://github.com/OpenCubicChunks/CubicChunks/commit/d7c0c4ed3bf30dbc7d06329bde97b9b280d1670a a while back, but it looks like the last release on CurseForge doesn't include that change.

If I merge this (into the CC branch at least, not necessarily the 1.12 branch) so people can test it out, is there a way to get CC with the transformer?

Barteks2x:
CC with the transformer is definitely on curse now, but I need to look over the code again and make sure that new changes don't introduce something unexpected.

wizjany:
Yea, though we just did a final(?) re-release of the forge 1.12 WE since there were a few key commits on that branch that weren't on CF. I don't mind having this stuff as a continuation on the 1.12 branch, though we're not really going to maintain it, especially since octy ported the changed to WE 7 already.

Barteks2x:
This build of WorldEdit is already being used anyway (quite a lot), just distributed as a jar file instead of downloaded from curse. So I don't particularly mind it being unmaintained. So is the currently used build.

wizjany:
"This build" being the current EngineHub branch that uses reflection to call getMinY, or being your branch that lets CC do the transforming?
I'm ok merging this PR now, so that the builds will be available on our dev build site, and then we can later discuss if we want to pull it into the 1.12 branch directly or leave it separate.

Barteks2x:
The build I've been talking about is a build from this PR. I can also make it up to date with the latest 1.12.2 branch changes.

Right now I have a separate branch (mostly for internal use and testing, in no way intended for PR) with working `//regen` but that's very specific to cubic chunks and I don't see a nice way to integrate that into WorldEdit itself (see: https://github.com/Barteks2x/WorldEdit/commit/047994c50f214cd9ebee3b0c371f794ccd8a8fdc)

octylFractal:
Closing this as it will likely never be merged.

PR #514, opened 2019-08-03
Add -e/-b/-m flags to //stack and //move, to match copy and paste.

wizjany:
Fixes WORLDEDIT-3935.


wizjany:
er, seems i pushed this from local instead of via github, so e25a6d21cd49ef9afabadc786ce794d40120b767 for reference

PR #515, opened 2019-08-07
Fix source masks using EditSession instead of clipboard.

wizjany:
Eventually this entire system (Request) needs to be ripped out.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/FactoryConverter.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/FactoryConverter.java
@@ -29,10 +29,13 @@ import com.sk89q.worldedit.extension.input.InputParseException;
 import com.sk89q.worldedit.extension.input.ParserContext;
 import com.sk89q.worldedit.extension.platform.Actor;
 import com.sk89q.worldedit.extent.Extent;
+import com.sk89q.worldedit.extent.transform.BlockTransformExtent;
 import com.sk89q.worldedit.function.mask.Mask;
 import com.sk89q.worldedit.function.pattern.Pattern;
 import com.sk89q.worldedit.internal.annotation.ClipboardMask;
 import com.sk89q.worldedit.internal.registry.AbstractFactory;
+import com.sk89q.worldedit.math.transform.Transform;
+import com.sk89q.worldedit.session.ClipboardHolder;
 import com.sk89q.worldedit.session.request.RequestExtent;
 import com.sk89q.worldedit.util.formatting.text.Component;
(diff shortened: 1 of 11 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/FactoryConverter.java
```diff
@@ -42,29 +46,48 @@
...
+            new FactoryConverter<>(worldEdit, WorldEdit::getMaskFactory, "mask", c -> {}));
         commandManager.registerConverter(Key.of(BaseItem.class),
-            new FactoryConverter<>(worldEdit, WorldEdit::getItemFactory, "item"));
+            new FactoryConverter<>(worldEdit, WorldEdit::getItemFactory, "item", c -> {}));
+
+        commandManager.registerConverter(Key.of(Mask.class, clipboardMask()),
+                new FactoryConverter<>(worldEdit, WorldEdit::getMaskFactory, "mask",
+                        context -> {
+                            try {
+                                context.setExtent(context.getSession().getClipboard().getClipboard());
```
wizjany: note to self: get holder transform and wrap clipboard region in transform extent

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/FactoryConverter.java
```diff
@@ -42,29 +49,56 @@
 import org.enginehub.piston.inject.Key;
 
 import java.util.List;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 public class FactoryConverter<T> implements ArgumentConverter<T> {
 
+    @AutoAnnotation
+    private static ClipboardMask clipboardMask() {
```
kenzierocks: You don't need this part here. You can use `ClipboardMask.class` in `Key.of`.

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/FactoryConverter.java
```diff
@@ -42,29 +49,56 @@
...
     private final WorldEdit worldEdit;
     private final Function<WorldEdit, AbstractFactory<T>> factoryExtractor;
     private final String description;
+    private final Consumer<ParserContext> contextTweaker;
 
     private FactoryConverter(WorldEdit worldEdit,
                              Function<WorldEdit, AbstractFactory<T>> factoryExtractor,
-                             String description) {
+                             String description,
+                             Consumer<ParserContext> contextTweaker) {
```
kenzierocks: Might be nicer to allow `null` for this, and just pass in `null` instead of `c -> {}`.

PR #208, opened 2012-08-16
Add support for snapshot directories in the form of date/world

nvx:
Closes redmine issue http://redmine.sk89q.com/issues/1460

PR #518, opened 2019-09-08
Support date/world snapshot directories

kenzierocks:
This supersedes #208, building upon the original commit to strictly verify that there is a date in the snapshot path, rather than accepting any name in the above folder (would mis-match with a world named "region", for example).

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/snapshot/SnapshotRepository.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/snapshot/SnapshotRepository.java
@@ -92,11 +92,13 @@ public class SnapshotRepository {
                     detectDate(snapshot);
                     list.add(snapshot);
                 }
-            } else if (file.isDirectory() && file.getName().equalsIgnoreCase(worldName)) {
+            } else if (file.isDirectory()) {
                 for (String name : file.list(filter)) {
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/snapshot/SnapshotRepository.java
```diff
@@ -92,11 +95,28 @@ public SnapshotRepository(String dir) {
...
+                    }
+                } else {
+                    Calendar date = fileNameDateParser.detectDate(file);
+                    if (date != null) {
+                        // <date>/<world>
+                        ZonedDateTime zdt = ZonedDateTime.ofInstant(
+                            date.toInstant(),
+                            ZoneOffset.UTC
+                        );
+                        for (String name : file.list((dir, name) -> name.equalsIgnoreCase(worldName))) {
```
wizjany:
```suggestion
                        for (String name : file.list((dir, name) -> Files.getNameWithoutExtension(name).equalsIgnoreCase(worldName))) {
```
or else `world.zip` won't match `world`

wizjany: Also `filter` is ignored here so potentially non-snapshot files will be listed if they match the world name.

kenzierocks:
This PR is deprecated in favor of #524, which resolves this & other issues with the current snapshot API.

PR #519, opened 2019-09-08
/command tool

kenzierocks:
Binds a CraftScript to a tool. Supersedes #328.

Semi-related: I'm starting to think that shoving all WorldEdit tools into global name space might be a _bad_ idea, and perhaps we should start considering using `/tool` as a prefix. Especially with "/command", which I can see being a decent name for other things.


kenzierocks:
Following discussion on Discord, this doesn't appear to be very useful (just use the last-script command instead, or type the command again, or use command history), and if you want a command-based brush, the craftscript command can already equip a custom brush iirc.

PR #521, opened 2019-09-09
Cubic Chunks compatibility (version 7)

octylFractal:
This is a port of #511 for WE 7. I don't know if we added more spots that might need to be ported as well.

@Barteks2x I would appreciate your review on this as well, just to ensure it's compatible with CC 1.14.4 (when it's out).

P.S. Despite the commit name, this is encompassing the relevant diff of that PR vs. `forge-archive/1.12.2`. I thought that commit name + description would be best for attributing authorship.


worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1681,9 +1681,26 @@ public int makePyramid(BlockVector3 position, Pattern block, int size, boolean f
...
+    /**
+     * Thaw blocks in a cylinder.
+     *
+     * @param position the position
+     * @param radius the radius
+     * @param height the height (upwards and downwards)
+     * @return number of blocks affected
+     * @throws MaxChangedBlocksException thrown if too many blocks are changed
+     */
+    public int thaw(BlockVector3 position, double radius, int height)
```
me4502: Wouldn't the height bar part of the radius?

kenzierocks: No, we're making these into cylinders. The height will usually be wildly different (i.e. 255) for most ops.

worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
```diff
@@ -208,7 +210,8 @@ public void hpos2(Player player, LocalSession session) throws WorldEditException
     @CommandPermissions("worldedit.selection.chunk")
     public void chunk(Player player, LocalSession session,
                       @Arg(desc = "The chunk to select", def = "")
-                          BlockVector2 coordinates,
+                      @Chunk3d
+                          BlockVector3 coordinates,
```
Barteks2x: This looks like a breaking change to me, not sure if important.

kenzierocks: It's not considered API, so not dev-breaking, and due to the Chunk3d annotation, it also parses old syntax, so not user-breaking.

PR #526, opened 2019-10-24
Fix flipping of chests and stairs

Brokkonaut:


wizjany:
will close https://dev.enginehub.org/youtrack/issue/WORLDEDIT-4004

me4502:
Is it worthwhile ensuring that these "type" and "shape" properties have the values that we expect them to? I can see these being rather common property names amongst mods and future MC blocks.

Brokkonaut:
Ok, I added a check for that

Tsoccerguy3:
I just compiled the plugin for 7.01 and 7.1 . The fix is working for //flip North , South , East and West . 
//flip up and flip down are still broken .

Tsoccerguy3:
> 
> 
> Ok, I added a check for that

better test //flip up and //flip down

Brokkonaut:
> better test //flip up and //flip down

Should be fixed now

me4502:
Thanks

PR #527, opened 2019-10-26
Expressions Part 1: ANTLR

octylFractal:
This PR re-writes expressions in ANTLR instead of using a custom parser. This allows us to expand the parser with new features easily, as well as provide an official grammar that can be used to validate expressions in other languages.

As expressions are not considered API, all old classes have been discarded.

To be continued in:
Expressions Part 2: New types & methods (strings, arrays?)
Expressions Part 3: Native execution (MethodHandle/bytecode gen)


TomyLobo:
Does this actually solve any problem?

octylFractal:
1) It does not fix any problems by changing to ANTLR, but there are a few other bug fixes in the latest commits. These can be cherry picked if we decide against this PR.
2) Moving to JS would make validating existing expressions extremely hard, and I doubt every expression will continue to work, considering we have `^` and `**` as power expressions. I don't think it's a good choice, since we'd have to have some flag to indicate "I want JS expressions".
3) Matthew discussed migrating away from JS for CraftScripts, and towards having a universal language for expressions and CraftScripts. I personally prefer this direction, as CraftScripts can be migrated easily.

My issue with using pretty much any other language is that it is significantly larger than whatever we add. Currently Rhino is optional for those who don't need CraftScripts, and I don't want to start requiring it for running simple expressions. The alternatives are Groovy, Scala, Kotlin, and maybe Jython, but all of these come with large standard libraries. Would we be downloading these? Bundle expressions as another plugin? Optional feature?

octylFractal:
Re: checking existing expressions, I really don't have a lot to work with, so if anyone can _provide_ expressions, I can add them to the test suite. Otherwise, I will assume they do work, because proving the negative is impossible.

I will add the expressions we feature on the docs right now.

octylFractal:
To clear up any misconceptions, I only intend to add strings and arrays of double/string. Currently the latter is implemented using functions, and the former is quite necessary for handling 1.13+ blocks. I don't intend to add a huge type system. Personally I don't see it as enough to push for switching to JS, as it's a minor change to the existing ANTLR code.

TomyLobo:
Oracle Java 8 comes with a JS implementation called Nashorn, so Rhino isn't really necessary.

As for 1.13+ blocks, have they done away with numeric block IDs completely now?
I wasn't aware of that.
I thought they'd have to have some kind of numeric ID for efficient storage at least.

Honestly, I thought you were just going to add types as "neat thing".
My apologies, this seems necessary under those premises.

Btw, we'll probably want some kind of lookup table type in order to translate between a block id string and something that can be computed with.

wizjany:
Nashorn is officially deprecated as of java 11. While I have no faith in mojang upping the requirement anytime soon, it seems like a bad idea to migrate to something that is slated for removal.

Block IDs are internally ints, but this mapping is generated at runtime, so there is no consistency and it's terrible user UX to make them use the info tool constantly to check IDs. Additionally, the id is per state, meaning one block type generally encompasses an entire range of ids, and there's no "data value" per se, so you can't do `type, -1` for "all states of given type".

And yes, we have legacy mappings. They are already used atm but the implementation is rather inefficient. They're also used for converting legacy schematics.

TomyLobo:
What I meant is that if there will be arrays in the expression language, we should also add hashtables.
Seems like a minor step with the infrastructure for arrays already in place.

wizjany:
Ah ok. Does the conversion really need to be done in the expression though or is that unrelated?
All the mechanisms for converting strings to blocks to ids etc is done under the hood in other places (e.g BlockStateIdAccess).

I'm not sure if you just think a table/dict structure should be available for expressions anyway, or if you think it's necessary for the conversions. I thought that the plan was just to let users input strings for blocks/masks and convert things transparently so the user doesn't even have to know they are internally ints.

PR #528, opened 2019-10-28
Fix stack command diagonal shifting with -s flag

Pieter12345:
Fix `//stack -s [num]` shifting the selection to unexpected places. This fix causes `//stack -s [num]` behavior in diagonal directions to be consistent with the behavior in non-diagonal directions, which means that the last stack will be selected.


octylFractal:
Verified this does fix the issue. Thanks!

PR #529, opened 2019-10-31
Ensure we own commands before completing them.

wizjany:
Dunno if there's a better way to do this?


wizjany:
note to self: check where `/worldedit:xyz` stuff goes, since we don't register those (done by bukkit itself)

wizjany:
So, implementation note: I decided to strip out `/worldedit:` and just pass `xyz` to the command manager (and ultimately to piston) because I believe the plugin-prefixed command name is a platform detail (though I believe sponge has this system too?) and there's no reason another platform can't have valid commands with `:` in them, so I didn't think this should be handled in the command manager itself but rather on the platform-specific stuff.

octylFractal:
Forge also has this mechanism iirc, so perhaps we should add some common handling. Not sure if completions are broken when it is used though.

PR #530, opened 2019-11-08
Expressions Part 2: Compilation

octylFractal:
This PR replaces the current `EvaluatingVisitor` with a compiler using `MethodHandle`s. This allows for much faster execution (up to ~12x), at the cost of slightly complicating implementation.

Still TODO:
- [x] Remove old visitor
- [x] Potentially clean-up code. Please comment with any readability issues, and I'll try to smooth them out. MethodHandles aren't much more complicated than normal code, but the way they combine can be hard to understand.


worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/invoke/ExpressionCompiler.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.lang.invoke.LambdaConversionException;
+import java.lang.invoke.LambdaMetafactory;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+
+import static com.sk89q.worldedit.internal.expression.invoke.ExpressionHandles.COMPILED_EXPRESSION_SIG;
+import static com.sk89q.worldedit.internal.expression.invoke.ExpressionHandles.safeInvoke;
+import static java.lang.invoke.MethodType.methodType;
```
TomyLobo: These static imports don't help readability.
I think resolving those would help one to visually spot that, for example, COMPILED_EXPRESSION_SIG is from another class.

octylFractal: I guess the question is how many of them to remove? I can understand removing `COMPILED_EXPRESSION_SIG`, but `safeInvoke` gets a little verbose without the import.

`MethodType.methodType` is just plain redundant, and I don't agree with removing that import.

TomyLobo: safeInvoke seems to be only used once in this class

octylFractal: Yea, I inlined it for this class.

TomyLobo:
Seriously, we need to talk about import static.
I think it removes context from the place where the imported function is used and moves it up into the import area where no one looks.
Two examples of this:
1. :
     * `methodType`? where is that function? is it a wrapper to create `MethodType`s here? or is it elsewhere? Is it ours? is it API?
     * `MethodType.methodType`? where is that function? oh it is in the class `MethodType`, the same class of which I am creating an instance right now, so it's not a wrapper and it's not ours but in the API.
2. :
     * `ID`? where is that constant? at the top of the current class? or is it elswehere? is it ours? is it API?
     * `ExpressionLexer.ID`? where is that constant? oh it is in the class `ExpressionLexer`, which is our ANTLR-generated lexer.

With every "import static" you add, ask yourself, am I obfuscating the code here or is the context enough to explain every single use of this import.
A clear-cut example of good use of static imports are test cases:
I'd say in that case, the context explains what assertTrue does, for example.

octylFractal:
Going off of "context explains what `assertTrue` does", isn't that also the case for `methodType`? How are the two different? `assertTrue` could be ours, could be third-party. `methodType` is pretty clearly creating an object to represent the type of method we want from the `Lookup` object.

2 is kinda valid, but at the same time `ExpressionLexer.GREATER_THAN_OR_EQUAL`, etc. is very large, and to me, this:
```java
            switch (ctx.op.getType()) {
                case LESS_THAN:
                    return (l, r) -> boolToDouble(l < r);
                case LESS_THAN_OR_EQUAL:
                    return (l, r) -> boolToDouble(l <= r);
                case GREATER_THAN:
                    return (l, r) -> boolToDouble(l > r);
                case GREATER_THAN_OR_EQUAL:
                    return (l, r) -> boolToDouble(l >= r);
            }
```
reads cleaner than this:
```java
            switch (ctx.op.getType()) {
                case ExpressionLexer.LESS_THAN:
                    return (l, r) -> boolToDouble(l < r);
                case ExpressionLexer.LESS_THAN_OR_EQUAL:
                    return (l, r) -> boolToDouble(l <= r);
                case ExpressionLexer.GREATER_THAN:
                    return (l, r) -> boolToDouble(l > r);
                case ExpressionLexer.GREATER_THAN_OR_EQUAL:
                    return (l, r) -> boolToDouble(l >= r);
            }
```

It's so much extra text that my brain needs to skip over while reading.

My counter-argument to "you don't know where it is" is that the name of the class doesn't tell you either. ExpressionLexer could easily be third-party (maybe I added an expression parsing library), and to find out you would still need to go up to the imports if you're not in an IDE. If you _are_ in an IDE, then the point is totally invalid because you can navigate to definition in a single click, and then you immediately find out where it is.

TomyLobo:
`assertTrue` only appears in test cases and in test cases you know that `assertTrue` is from a library (which one in particular isn't really all that important in order to understand what the code does)
`MethodType.methodType` screams "factory method". `methodType` screams "helper method in this class"
I don't know about your brain, but my brain skips over the parts that stay the same in those lines:
```java
                case ExpressionLexer....:
                    return (l, r) -> boolToDouble(l ... r);
                case ExpressionLexer....:
                    return (l, r) -> boolToDouble(l ... r);
                case ExpressionLexer....:
                    return (l, r) -> boolToDouble(l ... r);
                case ExpressionLexer....:
                    return (l, r) -> boolToDouble(l ... r);
```
As you can see, not even half of that is the class names :)
As long as you're not editing code on your phone, horizontal space shouldn't be an issue, even if you operate your monitor in portrait mode.

When reviewing code on GitHub, one is typically not in an IDE and the imports are not easily accessible unless there was a change in them.

PR #531, opened 2019-11-09
Misc fixes

wizjany:
Only potential concern is defining a `-S` flag where `-s` is already in use.
Using lower/upper case flags in unix commands like this is pretty common imo, dunno how well that translates to minecraft users.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/generator/GardenPatchGenerator.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/generator/GardenPatchGenerator.java
@@ -26,7 +26,6 @@ import com.sk89q.worldedit.function.RegionFunction;
 import com.sk89q.worldedit.function.pattern.Pattern;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.world.block.BlockState;
-import com.sk89q.worldedit.world.block.BlockStateHolder;
 import com.sk89q.worldedit.world.block.BlockTypes;
 
 import java.util.Random;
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/generator/GardenPatchGenerator.java
```diff
@@ -193,11 +192,11 @@ public static Pattern getPumpkinPattern() {
      * Set a block only if there's no block already there.
      *
      * @param position the position
-     * @param block the block to set
+     * @param block the pattern to set
      * @return if block was changed
      * @throws MaxChangedBlocksException thrown if too many blocks are changed
      */
-    private static <B extends BlockStateHolder<B>> boolean setBlockIfAir(EditSession session, BlockVector3 position, B block) throws MaxChangedBlocksException {
+    private static boolean setBlockIfAir(EditSession session, BlockVector3 position, Pattern block) throws MaxChangedBlocksException {
```
octylFractal: Can we also rename the parameter?

TomyLobo:
I'm having trouble getting *IT people* to remember uppercase vs. lowercase flags, let alone non-technical people.
Better try to come up with another letter. there are 26 of them, one ought to be free :)

TomyLobo:
Maybe `-n`? That's a common short-form flag for `--dry-run` or `--not-really`.
Can be documented as "-n no actual pasting"

MattBDev:
I agree that - n would be a better flag for the onlySelect parameter.

PR #532, opened 2019-11-10
Bypass Bukkit in more situations

me4502:
Due to performance regressions in Bukkit, it makes sense to change how these keys operate.


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
```diff
@@ -360,7 +367,13 @@ public static BlockType asBlockType(Material material) {
         if (!material.isBlock()) {
```
wizjany: might be able to skip this (and the `isItem` below) as the EnumMap stores a `null` object for us and can just skip `Material`'s switches.

me4502: The javadoc for Map's computeIfAbsent says `If the specified key is not already associated with a value (or is mapped to null)`, so this may cause it to try to re-compute every time

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
```diff
@@ -387,21 +390,22 @@ public static ItemType asItemType(Material material) {
...
-                    return WorldEdit.getInstance().getBlockFactory().parseFromInput(input, TO_BLOCK_CONTEXT).toImmutableState();
-                } catch (InputParseException e) {
-                    e.printStackTrace();
-                    return null;
-                }
+        int cacheKey =
+                WorldEditPlugin.getInstance().getBukkitImplAdapter() != null
+                        ? WorldEditPlugin.getInstance().getBukkitImplAdapter().getInternalBlockStateId(blockData)
+                            .orElseGet(() -> blockData.getAsString().hashCode())
+                        : blockData.getAsString().hashCode();
```
octylFractal: Is there a test we could perform to validate that this doesn't collide? I would prefer to use an incrementing Atomic Integer, but if this can be proven to work its also okay.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
```diff
@@ -411,13 +415,8 @@ public BlockState apply(@Nullable String input) {
...
     public static <B extends BlockStateHolder<B>> BlockData adapt(B block) {
         checkNotNull(block);
-        return blockDataCache.computeIfAbsent(block.getAsString(), new Function<String, BlockData>() {
-            @Nullable
-            @Override
-            public BlockData apply(@Nullable String input) {
-                return Bukkit.createBlockData(block.getAsString());
-            }
-        }).clone();
+        int cacheKey = BlockStateIdAccess.getBlockStateId(block.toImmutableState()).orElse(block.hashCode());
```
octylFractal: Similar question here, though as it's not a string it may be different.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
me4502: An incrementing atomic integer wouldn't directly work here, as it needs to be reliably reproducible. As for hash collisions, it theoretically could collide yes. However only when the ImplAdapter is missing

PR #534, opened 2019-11-28
Fix legacy import

MicleBrick:
At least one old block which was renamed (stone slab -> smooth stone slab) shared names with a new block with the same name (stone slab). Since it previously only used the data fixer to set up the legacy map if there was no matching one for the old name in the new names, it would use the new block in the map, even though blocks on that version with that id were referring to the old block. This PR changes the order it tries it in from fixer if available, to default if that didn't work, to fail.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/LegacyMapper.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/LegacyMapper.java
@@ -96,19 +96,32 @@ public final class LegacyMapper {
             final String value = blockEntry.getValue();
             blockEntries.put(id, value);
 
-            try {
-                if (fixer != null) {
+            BlockState state = null;
+
+            // if fixer is available, try using that first, as some old blocks that were renamed share names with new blocks
+            if (fixer != null) {
+                try {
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/LegacyMapper.java
```diff
@@ -95,26 +95,34 @@ private void loadFromResource() throws IOException {
...
+                try {
                     String newEntry = fixer.fixUp(DataFixer.FixTypes.BLOCK_STATE, value, 1631);
-                    try {
-                        BlockState state = WorldEdit.getInstance().getBlockFactory().parseFromInput(newEntry, parserContext).toImmutableState();
-                        blockToStringMap.put(state, id);
-                        stringToBlockMap.put(id, state);
-                        fixed = true;
-                    } catch (InputParseException ignored) {
-                    }
+                    state = WorldEdit.getInstance().getBlockFactory().parseFromInput(newEntry, parserContext).toImmutableState();
```
octylFractal: Can you pull `WorldEdit.getInstance().getBlockFactory()` out to a variable? Seems appropriate.

octylFractal:
Thanks!

PR #535, opened 2019-12-01
Use commandblock name as session key

me4502:
We also need to decide whether these persist or not. It may make sense to persist them, as otherwise admins may be confused as to why fast mode has turned off etc

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
@@ -41,8 +41,6 @@ import javax.annotation.Nullable;
 
 public class BukkitBlockCommandSender extends AbstractNonPlayerActor implements Locatable {
 
-    private static final String SESSION_PREFIX = "CMDBLK";
-
     private final BlockCommandSender sender;
     private final WorldEditPlugin plugin;
     private final Location location;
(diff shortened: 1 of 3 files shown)


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
```diff
@@ -53,7 +53,7 @@ public BukkitBlockCommandSender(WorldEditPlugin plugin, BlockCommandSender sende
         this.plugin = plugin;
         this.sender = sender;
         this.location = BukkitAdapter.adapt(sender.getBlock().getLocation());
-        this.uuid = new UUID(location.toVector().toBlockPoint().hashCode(), location.getExtent().hashCode());
```
octylFractal: I know we were doing it before, but I would prefer if we used a standard UUID generation scheme, such as UUIDv4 (Random) seeded with the name, or UUIDv5 (namespaced).

octylFractal: Or perhaps, since it's built-in to Java, UUIDv3 (also namespaced): https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html#nameUUIDFromBytes-byte:A-

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
```diff
@@ -53,7 +53,7 @@ public BukkitBlockCommandSender(WorldEditPlugin plugin, BlockCommandSender sende
...
+        this.uuid = UUID.nameUUIDFromBytes(sender.getName().getBytes());
```
octylFractal: We should use the Charset-specifying version of `getBytes()` so that session persistence is OS-independent.

wizjany: random thought: is this the same method MC uses to gen offline uuids? if so, is it possible on offline servers to make a commandblock named `OfflinePlayer:xyz` and have them share a session with a player named `xyz`? if so, do we care?

me4502: I've resolved both of these issues

PR #536, opened 2019-12-06
Add //line support for convex polyhedral selections

enveeed:
This PR adds support for convex polyhedral selections to the `//line` command.

It allows connecting the vertices of a convex polyhedral selection with lines, making it work similar to `//curve` but instead of a spline all vertices are connected with lines. The old functionality with connecting the two corner positions of a cuboid selection is preserved.

`EditSession.drawLine(...)` was modified to take a `BlockVector3` list, just like `EditSession.drawSpline(...)`

screenshot displaying generated lines

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -19,7 +19,7 @@
 
 package com.sk89q.worldedit;
 
-import com.google.common.collect.Lists;
+import com.google.common.collect.ImmutableList;
 import com.sk89q.worldedit.entity.BaseEntity;
 import com.sk89q.worldedit.entity.Entity;
 import com.sk89q.worldedit.event.extent.EditSessionEvent;
@@ -2187,12 +2187,12 @@ public class EditSession implements Extent, AutoCloseable {
      *
      * @return number of blocks affected
      * @throws MaxChangedBlocksException thrown if too many blocks are changed
-     * 
-     * @see #drawLine(Pattern, List, double, boolean) 
+     *
+     * @see #drawLine(Pattern, List, double, boolean)
      */
     public int drawLine(Pattern pattern, BlockVector3 pos1, BlockVector3 pos2, double radius, boolean filled)
             throws MaxChangedBlocksException {
-        return drawLine(pattern, Lists.newArrayList(pos1,pos2), radius, filled);
+        return drawLine(pattern, ImmutableList.of(pos1, pos2), radius, filled);
     }
 
     /**
@@ -2212,9 +2212,6 @@ public class EditSession implements Extent, AutoCloseable {
         Set<BlockVector3> vset = new HashSet<>();
 
         for (int i = 0; vectors.size() != 0 && i < vectors.size() - 1; i++) {
-
-            boolean notdrawn = true;
-
             BlockVector3 pos1 = vectors.get(i);
             BlockVector3 pos2 = vectors.get(i + 1);
 
@@ -2225,10 +2222,11 @@ public class EditSession implements Extent, AutoCloseable {
 
             if (dx + dy + dz == 0) {
                 vset.add(BlockVector3.at(tipx, tipy, tipz));
-                notdrawn = false;
+                continue;
             }
 
-            if (Math.max(Math.max(dx, dy), dz) == dx && notdrawn) {
+            int dMax = Math.max(Math.max(dx, dy), dz);
+            if (dMax == dx) {
                 for (int domstep = 0; domstep <= dx; domstep++) {
                     tipx = x1 + domstep * (x2 - x1 > 0 ? 1 : -1);
                     tipy = (int) Math.round(y1 + domstep * ((double) dy) / ((double) dx) * (y2 - y1 > 0 ? 1 : -1));
@@ -2236,10 +2234,7 @@ public class EditSession implements Extent, AutoCloseable {
 
                     vset.add(BlockVector3.at(tipx, tipy, tipz));
                 }
-                notdrawn = false;
-            }
-
-            if (Math.max(Math.max(dx, dy), dz) == dy && notdrawn) {
+            } else if (dMax == dy) {
                 for (int domstep = 0; domstep <= dy; domstep++) {
                     tipy = y1 + domstep * (y2 - y1 > 0 ? 1 : -1);
                     tipx = (int) Math.round(x1 + domstep * ((double) dx) / ((double) dy) * (x2 - x1 > 0 ? 1 : -1));
@@ -2247,10 +2242,7 @@ public class EditSession implements Extent, AutoCloseable {
 
                     vset.add(BlockVector3.at(tipx, tipy, tipz));
                 }
-                notdrawn = false;
-            }
-
-            if (Math.max(Math.max(dx, dy), dz) == dz && notdrawn) {
+            } else /* if (dMax == dz) */ {
                 for (int domstep = 0; domstep <= dz; domstep++) {
                     tipz = z1 + domstep * (z2 - z1 > 0 ? 1 : -1);
                     tipy = (int) Math.round(y1 + domstep * ((double) dy) / ((double) dz) * (y2-y1>0 ? 1 : -1));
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
```diff
@@ -125,16 +125,25 @@ public int line(Actor actor, EditSession editSession,
...
         checkCommandArgument(thickness >= 0, "Thickness must be >= 0");
 
-        CuboidRegion cuboidregion = (CuboidRegion) region;
-        BlockVector3 pos1 = cuboidregion.getPos1();
-        BlockVector3 pos2 = cuboidregion.getPos2();
-        int blocksChanged = editSession.drawLine(pattern, pos1, pos2, thickness, !shell);
+
+        List<BlockVector3> vectors;
+
+        if(region instanceof CuboidRegion) {
```
TheMolkaPL: There should be a space after `if`.

```java
if (statement)
```

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
```diff
@@ -125,16 +125,25 @@ public int line(Actor actor, EditSession editSession,
...
+            CuboidRegion cuboidregion = (CuboidRegion) region;
+            vectors = Lists.newArrayList(cuboidregion.getPos1(), cuboidregion.getPos2());
+        }
+        else {
```
TheMolkaPL:
```java
} else {
```

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -2176,64 +2176,70 @@ public int hollowOutRegion(Region region, int thickness, Pattern pattern) throws
...
-     * @param pos2 The other point that defines the line.
+     * @param vectors the list of vectors to draw the line between
      * @param radius The radius (thickness) of the line.
      * @param filled If false, only a shell will be generated.
      *
      * @return number of blocks affected
      * @throws MaxChangedBlocksException thrown if too many blocks are changed
      */
-    public int drawLine(Pattern pattern, BlockVector3 pos1, BlockVector3 pos2, double radius, boolean filled)
+    public int drawLine(Pattern pattern, List<BlockVector3> vectors, double radius, boolean filled)
```
wizjany: this is a breaking signature change, please leave an overload

enveeed: Oh yeah, not sure why that didn't cross my mind. Will fix.

wizjany:
haven't gotten time to test yet, but what's the behavior when making hollow lines? i think the spline/curve leaves the connections hollow as well, right? does this line work the same, or are ends capped?

enveeed:
I believe it works the same, due to the usage of `getBallooned(...)` and `getHollowed(...)`, I did not change anything in that regard. Here is an example:

top-down view of cut-in-half example generation result

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
enveeed: Do you think a
```
@see #drawLine(Pattern, List, double, boolean) 
```
doc addition to the overload for the old signature would be good or shall I rather leave it out?

wizjany: I think I personally tend to copy the whole thing (and change the individual params), you can probably check some other overloaded methods here in EditSession. `@see` is probably ok too, definitely wouldn't leave it blank.

octylFractal: IMO it would be a good idea to deprecate the old version too, just to keep the API small in the future. Not sure how wiz feels about that.

NJDaeger:
Couldn't this also work for Polygonial2DRegions?

wizjany:
> 
> 
> Couldn't this also work for Polygonial2DRegions?

No, Poly2D are 2d. You'd have to (arbitrarily) either use the min or max of the region as the y-coord.

NJDaeger:
Could it just gradually move from the min point to the max point as the line is drawn then?

wizjany:
starting from where? is pos1 miny and posN maxY? why not the other way? you're introducing a lot of behavioral assumptions that, if implemented, would need their own set of options.

NJDaeger:
Hmm. Well instead of doing that, maybe it could just fill the entire region with a wall from the min point to the max point?

wizjany:
like say maybe, a //walls command?

NJDaeger:
Kind of? I feel like with the walls command, you'd want to connect the origin point with the last point added to the selection to make a closed wall. The line command could be made to work the same way except it wouldn't make it a closed wall, the line would just stop at the endpoints of the selection, if that makes sense.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
wizjany: while I am not against deprecation here, we do have a lot of convenience overloads already (eg point+apothem vs region args). would we want to deprecate those too?

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
```diff
@@ -112,8 +112,8 @@ public int set(Actor actor, EditSession editSession,
 
     @Command(
         name = "/line",
-        desc = "Draws a line segment between cuboid selection corners",
```
wizjany: `segment` is correct here, not sure why it was removed? Probably should be "Draws line segment*s*" now though.

enveeed: You're right, should've put more care into editing the strings. Coming right up.

wizjany:
> Kind of? I feel like with the walls command, you'd want to connect the origin point with the last point added to the selection to make a closed wall. The line command could be made to work the same way except it wouldn't make it a closed wall, the line would just stop at the endpoints of the selection, if that makes sense.

That's still far beyond what I would expect a `//line` command to do. Sounds more like you want to add options to `//walls` instead, which is straying farther and farther from the scope of this PR.

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
enveeed: Do you believe this would be good?

> "Draws line segments between cuboid selection corners or convex polyhedral selection vertices"

It kinda irks me that it talks about segments in the plural but the cuboid selection will only generate one segment, i'm not a native English speaker though.

NJDaeger:
I'm not saying implement walls, I'm just trying to justify that there would be differences with what they do. I don't think it's beyond what //line would do personally, as it's essentially what it does already with the set command.

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
wizjany: That sounds fine to me...if you really want to be explicit about the separation, `Draws a line segment between cuboid corners, or line segments between convex poly vertices.` but that's getting verbose. Language-wise, saying `segments` is fine though.

wizjany:
> I'm not saying implement walls, I'm just trying to justify that there would be differences with what they do. I don't think it's beyond what //line would do personally, as it's essentially what it does already with the set command.

what? what does the set command have to do with anything? You are literally just asking for the walls command but without the plane formed by joining the first and last vertices in a poly2d region. This has nothing to do with drawing lines between two points any more.

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
enveeed: I see, thanks for your input. I will use what I came up with then.

octylFractal:
I added some simplifications to the logic now that a loop has been introduced, and used ImmutableList instead.

octylFractal:
Tested and this works with latest changes still. Thanks!

PR #537, opened 2019-12-07
Bypass bukkit api for regeneration.

wizjany:
Done in impl adapter now.

Todo:
- [ ] Test on 1.13 I guess
- [x] Maybe change the failure message? ("unsupported on this platform" or "check console for errors"?)


worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
```diff
@@ -400,22 +400,25 @@ public int stack(Actor actor, World world, EditSession editSession, LocalSession
...
-            world.regenerate(region, editSession);
+            success = world.regenerate(region, editSession);
         } finally {
             session.setMask(mask);
         }
-        actor.print("Region regenerated.");
+        if (success) {
+            actor.print("Region regenerated.");
+        } else {
+            actor.print("Unable to regenerate chunks.");
```
octylFractal: Shouldn't this be `printError`?

As to the message, I think it's fine, maybe add a "See console for details." sentence.

wizjany: the issue is that there isn't a console message on unsupported platforms (eg no adapter). could add one i guess?

octylFractal: Yes, I would add one. I think that's the cleanest solution at the moment, I would prefer an exception but it would be weird to return a bool status _and_ throw exceptions.

PR #539, opened 2019-12-11
BukkitAdapter: Add Bukkit CommandSender --> WorldEdit Actor

JOO200:
I want to use patterns in my plugin for the console.
There is no nice way to adapt a Bukkit CommandSender to a WorldEdit Actor.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
@@ -142,6 +142,16 @@ public class BukkitAdapter {
     }
 
     /**
+     * Create a Bukkit CommandSender from a WorldEdit Actor.
+     *
+     * @param actor The WorldEdit actor
+     * @return The Bukkit command sender
+     */
+    public static CommandSender adapt(Actor actor) {
+        return ((BukkitCommandSender) actor).getSender();
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


octylFractal:
Technically speaking, this is not necessary because the method is already public. You're free to call it on the plugin yourself.

JOO200:
Yes and no: The method is public but WorldEdit.getInstance() is package private.

wizjany:
notably, WG always has its wrap method on the plugin instead of some adapter class, and it's in the "officially documented" api, so I would probably treat WE the same way.
and yea, you're supposed to get the plugin via plugin manager, as "plugin" methods are dependent on plugin availability (vs just platform availability)

me4502:
IMO if we do this, it also makes sense to include one specifically for players

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
```diff
@@ -119,6 +121,16 @@ public static World adapt(org.bukkit.World world) {
...
+    /**
+     * Create a WorldEdit Actor from a Bukkit CommandSender
+     *
+     * @param sender The Bukkit CommandSender
+     * @return The WorldEdit Actor
+     */
+    public static Actor adapt(CommandSender sender) {
+        return WorldEditPlugin.getInstance().wrapCommandSender(sender);
+    }
+
```
DarkArc: Should probably allow the other way as well `Actor` -> `CommandSender`.

PR #543, opened 2020-01-06
Make missing Bukkit impl adapter more user-friendly.

wizjany:
* 100s of "Unknown property" warnings in log aren't useful, only drown
  out the actual adapter warning
* Most people aren't looking at their console anyway, so hijack the old
  Actor#tellVersion to warn of missing adapters.

example when first running a command (per session):


octylFractal:
Perhaps we can name Actor#tellVersion differently? Like "greet" or something? I know it won't align with LS, but I think it would be worthwhile.

PR #545, opened 2020-01-25
Update maze.js

Gimlao:
Now works with the current selected region, using the region width+length+height to generate the maze.
Also removed the vertical maze feature.

### DIFF
--- a/contrib/craftscripts/maze.js
+++ b/contrib/craftscripts/maze.js
@@ -21,363 +21,314 @@ importPackage(Packages.com.sk89q.worldedit);
 importPackage(Packages.com.sk89q.worldedit.blocks);
 importPackage(Packages.com.sk89q.worldedit.math);
 
-usage = "<block> [width] [length] [height] [size] [thickness] flags\n";
-usage += "\n";
-usage += "Maze Flags:\n";
-usage += "• i adds an entry and an exit\n";
-usage += "• y places the entry and the exit randomly\n";
-usage += "• f adds a floor\n";
-usage += "• c adds a ceiling\n";
-usage += "• e places air blocks\n";
-usage += "• a places air blocks only\n";
-usage += "• v creates a vertical maze\n";
-usage += "\n";
-usage += "Solver Flags:\n";
-usage += "• s enables the maze solver\n";
-usage += "• g places glass if wrong or unvisited\n";
-usage += "• r places red wool if wrong\n";
-usage += "• b places blue wool if unvisited";
+usage = "Usage :\n";
+usage += "§6/.s <block> [size] [thickness] flags\n";
+usage += "§7• §6[size] §r: Changes corridors' width\n";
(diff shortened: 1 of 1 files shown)


wizjany:
Seems a bit like this should be its own thing, given all the removed functionality.

Gimlao:
Actually it's now really easier to generate a maze like this.

You now only need to select a region to determine the maze width, length and height, instead of always trying multiple commands, fly in the air to see how if looks like, /undo, try another command...

It's still doing the exact same things.
The only thing I removed is the vertical feature, since it was only me trying stupid things in 2013. x)

wizjany:
point is that `//sphere` wasn't removed just because `//sel sphere` was added.

Gimlao:
Maybe I should create an alternate "region_maze.js" file ?

wizjany:
that was the idea yea

Gimlao:
So here is the old "maze.js" script with the color update on the usage message + "/.s ?" to show help about flags.

I'm going to create another "region_maze.js" file, instead.

contrib/craftscripts/maze.js
```diff
@@ -21,363 +21,369 @@ importPackage(Packages.com.sk89q.worldedit);
...
-usage += "\n";
-usage += "Solver Flags:\n";
-usage += "• s enables the maze solver\n";
-usage += "• g places glass if wrong or unvisited\n";
-usage += "• r places red wool if wrong\n";
-usage += "• b places blue wool if unvisited";
-
-context.checkArgs(1, -1, usage);
+usage = "Usage :\n";
+usage += "§6/.s <block> [width] [length] [height] [size] [thickness] flags\n";
```
octylFractal: If we're going to add color I would rather not do it with the legacy color codes. Craftscripts should be able to use the text libraries just fine.

contrib/craftscripts/maze.js
```diff
@@ -21,363 +21,369 @@ importPackage(Packages.com.sk89q.worldedit);
...
-        }
-
-        if (c) {
-            if (!v) sess.setBlock(origin.add(x, z - 1 + h, y), block);
-            else sess.setBlock(origin.add(x, y, -z + 1 - h), block);
-        }
+if (!argv[1]) context.print(usage);
+else if (argv[1] == "?") context.print(help);
+else {
+    block = context.getBlock(argv[1]);
```
octylFractal: Rather than dump the entire rest of the script into an `else`, would it be possible to throw an exception / error after the `context.print`? I feel like it may not be super user-friendly, and we should probably add some sort of `exit(Component)`, but having literally the entire rest of the script indented is a big 👎 from me.

Gimlao:
@octylFractal Feel free to help me there and modify things as you want, since I actually don't know how to do that. =/

Same thing with the "region_maze.js" file.

me4502:
I'm closing this as you haven't resolved the noted issues in almost a year. If you want to fix the issues and re-submit, feel free to reopen the PR or make a new one.

PR #547, opened 2020-01-27
Replace fast mode with a side effect system, with fine-grained control

me4502:
Just trying out a few performance gains. 

This PR also includes this worldedit-adapters branch, https://github.com/EngineHub/worldedit-adapters/compare/feature/performance-tweaks.

I'm seeing smallish to moderate improvements with the removal of streams and optionsals, and a large improvement with the changes to the adapter. The adapter changes remove physics events, which get worse the more plugins use it.


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitServerInterface.java
```diff
@@ -218,6 +220,18 @@ public String getPlatformVersion() {
         return capabilities;
     }
 
+    private static final Collection<SideEffect> SUPPORTED_SIDE_EFFECTS = ImmutableSet.copyOf(EnumSet.of(
```
kashike: You can use `Sets.immutableEnumSet` instead of `ImmutableSet.copyOf(EnumSet.of` - applies to all cases throughout this PR.

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIPlatform.java
```diff
@@ -153,6 +156,11 @@ public String getPlatformVersion() {
         return capabilities;
     }
 
+    @Override
+    public Collection<SideEffect> getSupportedSideEffects() {
+        return Lists.newArrayList();
```
octylFractal: Probably doesn't matter, but this can use `ImmutableList.of()`, `Collections.emptyList()`, or any other constant empty collection to avoid a tiny allocation. It's generally more performant and compact as well.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/adapter/BukkitImplAdapter.java
```diff
@@ -178,6 +182,13 @@ default boolean simulateItemUse(World world, BlockVector3 position, BaseItem ite
      */
     BaseItemStack adapt(ItemStack itemStack);
 
+    /**
+     * Get the {@link SideEffect}s that this adapter supports.
+     *
+     * @return The side effects that are supported
+     */
+    Collection<SideEffect> getSupportedSideEffects();
```
octylFractal: I would prefer if we enforced this to be a `Set`. I don't think having multiple of the same `SideEffect` has any meaning.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/adapter/BukkitImplAdapter.java
```diff
@@ -91,19 +94,20 @@ default void tickWatchdog() {
...
     /**
      * Notifies the simulation that the block at the given location has
      * been changed and it must be re-lighted (and issue other events).
      *
      * @param position position of the block
      * @param previousType the type of the previous block that was there
+     * @param sideEffectApplier side effects to apply
      */
-    void notifyAndLightBlock(Location position, BlockState previousType);
+    void notifyAndLightBlock(Location position, BlockState previousType, SideEffectApplier sideEffectApplier);
```
octylFractal: If we're breaking binary compat, might as well rename this to something better, probably the same as the `World` name.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -465,9 +466,24 @@ public SurvivalModeExtent getSurvivalExtent() {
...
-            fastModeExtent.setEnabled(enabled);
+        if (worldApplyingExtent != null) {
+            worldApplyingExtent.setSideEffectApplier(enabled ? SideEffectApplier.defaults() : SideEffectApplier.none());
+        }
+    }
+
+    /**
+     * Set which block updates should occur.
+     *
+     * <p>Fast mode may skip lighting checks or adjacent block
```
octylFractal: This javadoc text should be removed / changed.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/Platform.java
```diff
@@ -174,4 +177,5 @@
      */
     Map<Capability, Preference> getCapabilities();
 
+    Collection<SideEffect> getSupportedSideEffects();
```
octylFractal: Can we also limit this to a `Set`?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
```diff
@@ -289,6 +291,10 @@ public LocalConfiguration getConfiguration() {
         return queryCapability(Capability.CONFIGURATION).getConfiguration();
     }
 
+    public Collection<SideEffect> getSupportedSideEffects() {
```
octylFractal: Also return `Set` here if above changes apply.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/world/WorldApplyingExtent.java
```diff
@@ -27,64 +27,37 @@
...
-import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 /**
- * Implements "fast mode" which may skip physics, lighting, etc.
+ * An extent that sets blocks in the world, with BlockUpdateOptions.
  */
-public class FastModeExtent extends AbstractDelegateExtent {
+public class WorldApplyingExtent extends AbstractDelegateExtent {
```
octylFractal: I don't know if I really like this name. Does `SideEffectExtent` sound good?

worldedit-core/src/main/java/com/sk89q/worldedit/util/SideEffect.java
```diff
@@ -0,0 +1,74 @@
...
+    public boolean requiresCleanup() {
+        return this.dirty;
+    }
+
+    public enum State {
+        OFF,
+        ON,
+        DELAYED;
+
+        private String displayName;
```
octylFractal: Also make this `final`.

worldedit-core/src/main/java/com/sk89q/worldedit/util/SideEffect.java
```diff
@@ -0,0 +1,74 @@
...
+    LIGHTING(State.ON, false),
+    NEIGHBORS(State.ON, false),
+    CONNECTIONS(State.ON, false),
+    ENTITY_AI(State.OFF, false),
+    PLUGIN_EVENTS(State.OFF, false);
+
+    private String displayName;
+    private String description;
+    private State defaultValue;
+    private boolean dirty;
```
octylFractal: I would prefer if we made these fields `final`.

Also, I think it makes more sense if `dirty` and `requireCleanup()` shared the same name (I like the latter).

worldedit-core/src/main/java/com/sk89q/worldedit/util/SideEffectApplier.java
```diff
@@ -0,0 +1,93 @@
...
+        this.sideEffects = ImmutableMap.of();
+        updateSideEffects();
+    }
+
+    public SideEffectApplier(Map<SideEffect, SideEffect.State> sideEffects) {
+        this.sideEffects = Maps.immutableEnumMap(sideEffects);
+        updateSideEffects();
+    }
+
+    private void updateSideEffects() {
```
octylFractal: I find it weird that you have a separate method for this, rather than dumping it in the `(Map)` constructor, and using `this(ImmutableMap.of())` for the no-arg constructor (`Maps.immutableEnumMap` can handle empty maps fine)

worldedit-core/src/main/java/com/sk89q/worldedit/util/SideEffectApplier.java
```diff
@@ -0,0 +1,93 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+
+import java.util.Arrays;
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class SideEffectApplier {
```
octylFractal: This name seems odd for something that doesn't actually handle _applying_ the side effect. It's not an _applier_, it's just a collection of side effects. I don't have a better name than `SideEffectSet`, and that doesn't seem quite right here. Maybe someone else has a good name?

worldedit-core/src/main/java/com/sk89q/worldedit/util/formatting/component/SideEffectBox.java
```diff
@@ -0,0 +1,80 @@
...
+public class SideEffectBox extends PaginationBox {
+
+    private static List<SideEffect> sideEffects;
+
+    private SideEffectApplier sideEffectApplier;
+
+    public SideEffectBox(SideEffectApplier sideEffectApplier) {
+        super("Side Effects");
+
+        if (sideEffects == null) {
```
octylFractal: This would be better as a static `getSideEffects`, rather than randomly in the constructor.

worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
```diff
@@ -107,17 +108,54 @@
...
      * Notifies the simulation that the block at the given location has
      * been changed and it must be re-lighted (and issue other events).
      *
      * @param position position of the block
      * @param previousType the type of the previous block that was there
+     * @param sideEffectApplier which side effects to perform
      * @return true if the block was successfully notified
      */
-    boolean notifyAndLightBlock(BlockVector3 position, BlockState previousType) throws WorldEditException;
+    boolean notifyBlock(BlockVector3 position, BlockState previousType, SideEffectApplier sideEffectApplier) throws WorldEditException;
```
octylFractal: I don't like the name `notifyBlock` -- if it is really notifying the simulation, we should name it `sendBlockChange` or something. Personally I prefer the name I had, `applySideEffects`.

octylFractal: To be clear, my issue with the current name is that it sounds like it does something with the block itself, which isn't really true, it's about the change at the location.

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorld.java
```diff
@@ -166,8 +168,44 @@ public Path getStoragePath() {
         return null;
     }
 
+    public void markAndNotifyBlock(World world, BlockPos pos, @Nullable WorldChunk worldChunk, net.minecraft.block.BlockState blockState,
```
octylFractal: I wonder if mixins has some trick we can use to call this dynamically, rather than needing to copy it out.

worldedit-core/src/main/java/com/sk89q/worldedit/util/SideEffectApplier.java
kashike: `SideEffectSet` sounds good to me.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/world/WorldApplyingExtent.java
me4502: Possibly, yeah - my main reason for naming it this, is it's the extent that actually writes the changes to the world

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorld.java
me4502: Copying it out makes it easier to understand IMO, also IIRC this didn't actually exist within Fabric, it's embedded within a method. Forge & Bukkit both split it out to its own method.

worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
```diff
@@ -117,7 +141,20 @@
...
+        return applySideEffects(position, previousType, SideEffectSet.defaults());
+    }
+
+    /**
+     * Applies a set of side effects on the given block.
+     *
+     * @param position position of the block
+     * @param previousType the type of the previous block that was there
+     * @param sideEffectSet which side effects to perform
+     * @return true if the block was successfully notified
```
octylFractal: This still says `notified` rather than ... `@return true if the side effects were successfully applied`.

Thinking about API design here, would it make more sense to (1) use a normal `Set<SideEffect>` here, as the specific state of each effect has no bearing on the action taken by `applySideEffects`; and (2) return a `Set<SideEffect>` of the effects that were actually applied OR not able to be applied (not sure which is better). It just seems like a `boolean` is not expressive enough when a platform could implement _some_ effects but not all.

worldedit-core/src/main/java/com/sk89q/worldedit/util/SideEffectSet.java
```diff
@@ -0,0 +1,94 @@
...
+    private boolean appliesAny;
+
+    private SideEffectSet() {
+        this(ImmutableMap.of());
+    }
+
+    public SideEffectSet(Map<SideEffect, SideEffect.State> sideEffects) {
+        this.sideEffects = Maps.immutableEnumMap(sideEffects);
+
+        appliesAny = sideEffects.values().stream().anyMatch(state -> state != SideEffect.State.OFF);
```
wizjany: can't this line just be moved below the next and read `appliesAny = !appliedSideEffects.isEmpty()`?

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -129,25 +133,67 @@ public void timeout(Actor actor, LocalSession session,
...
-            actor.printInfo(TranslatableComponent.of("worldedit.fast.disabled"));
-        } else {
-            session.setFastMode(true);
-            actor.printInfo(TranslatableComponent.of("worldedit.fast.enabled"));
+        if (sideEffect == null || showInfoBox) {
+            SideEffectBox sideEffectBox = new SideEffectBox(session.getSideEffectSet());
+            try {
+                actor.print(sideEffectBox.create(1));
+            } catch (InvalidComponentException e) {
+                e.printStackTrace();
```
wizjany: re-throw wrapped as CommandException? shouldn't happen but if it does the user gets no feedback in chat

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/EnumConverter.java
```diff
@@ -21,14 +21,17 @@
 
 import com.google.common.collect.ImmutableSet;
 import com.sk89q.worldedit.EditSession;
+import com.sk89q.worldedit.WorldEdit;
```
wizjany: random import?

worldedit-core/src/main/java/com/sk89q/worldedit/util/SideEffectSet.java
```diff
@@ -0,0 +1,94 @@
...
+
+public class SideEffectSet {
+    private static final SideEffectSet DEFAULT = new SideEffectSet(
+            Arrays.stream(SideEffect.values()).collect(Collectors.toMap(Function.identity(), SideEffect::getDefaultValue))
+    );
+    private static final SideEffectSet NONE = new SideEffectSet();
+
+    private final Map<SideEffect, SideEffect.State> sideEffects;
+    private final Set<SideEffect> appliedSideEffects;
+    private boolean appliesAny;
```
wizjany: can be final if the above are as well

worldedit-core/src/main/java/com/sk89q/worldedit/util/formatting/component/SideEffectBox.java
```diff
@@ -0,0 +1,84 @@
...
+
+        TextComponent.Builder builder = TextComponent.builder();
+        builder = builder.append(TranslatableComponent.of(effect.getDisplayName(), TextColor.YELLOW)
+                .hoverEvent(HoverEvent.of(HoverEvent.Action.SHOW_TEXT, TranslatableComponent.of(effect.getDescription()))));
+        for (SideEffect.State uiState : SideEffect.State.values()) {
+            builder = builder.append(TextComponent.space());
+            builder = builder.append(TranslatableComponent.of(uiState.getDisplayName(), uiState == state ? TextColor.WHITE : TextColor.GRAY)
+                    .clickEvent(ClickEvent.runCommand("//fast -h " + effect.name().toLowerCase(Locale.US) + " " + uiState.name().toLowerCase(Locale.US)))
+                    .hoverEvent(HoverEvent.showText(uiState == state ?
+                     TextComponent.of("Current") : TextComponent.of("Click to set to ").append(TranslatableComponent.of(uiState.getDisplayName()))))
```
wizjany: i18n?

worldedit-core/src/main/java/com/sk89q/worldedit/world/NullWorld.java
```diff
@@ -66,14 +70,14 @@ public String getId() {
         return "null";
     }
 
-    @Override
-    public <B extends BlockStateHolder<B>> boolean setBlock(BlockVector3 position, B block, boolean notifyAndLight) throws WorldEditException {
+    @Override public <B extends BlockStateHolder<B>> boolean setBlock(BlockVector3 position, B block, SideEffectSet sideEffectSet) throws WorldEditException {
```
wizjany: override lost its line

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
me4502: I copied this from WorldGuard code :P

wizjany: ouch, got me.

worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
```diff
@@ -107,7 +110,30 @@
...
      */
-    <B extends BlockStateHolder<B>> boolean setBlock(BlockVector3 position, B block, boolean notifyAndLight) throws WorldEditException;
+    @Deprecated
+    default <B extends BlockStateHolder<B>> boolean setBlock(BlockVector3 position, B block, boolean notifyAndLight) throws WorldEditException {
+        return setBlock(position, block, notifyAndLight ? SideEffectSet.defaults() : SideEffectSet.none());
+    }
+
+    /**
+     * Similar to {@link Extent#setBlock(BlockVector3, BlockStateHolder)} but a
+     * {@code notifyAndLight} parameter indicates whether adjacent blocks
```
octylFractal:
```suggestion
     * {@code sideEffectSet} parameter indicates whether adjacent blocks
```

octylFractal: Honestly this whole javadoc needs a rewording I think.

worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
```diff
@@ -107,7 +110,30 @@
...
+     *
+     * <p>On implementations where the world is not simulated, the
+     * {@code notifyAndLight} parameter has no effect either way.</p>
+     *
+     * @param position position of the block
+     * @param block block to set
+     * @param sideEffectSet which side effects to perform
+     * @return true if the block was successfully set (return value may not be accurate)
+     */
+    <B extends BlockStateHolder<B>> boolean setBlock(BlockVector3 position, B block, SideEffectSet sideEffectSet) throws WorldEditException;
```
octylFractal: Do we want to report the same information as `applySideEffects` return value somehow?

me4502: I personally think for this one it's not as necessary. The point of this method is to set the block, so the return type should reflect whether that was done correctly. The side effects are just side effects, after all.

PR #548, opened 2020-01-30
Improve speed of BSIA, clean up Int2BaseBlockMap

octylFractal:
This drops the BiMap in `BlockStateIdAccess` for a `Int2ObjectOpenHashMap<BlockState>` & a field in `BlockState`. The hash computation is relatively slow vs. a field read, so this improves the speed of putting blocks into a `BlockMap` significantly (~20s over 50 million puts).

I also cleaned up Int2BaseBlockMap to have no repeated reads to BSIA, but according to timing data, it's likely that the JVM was optimizing those out already! This just ensures that it happens before JIT kicks in.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/Capability.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/Capability.java
@@ -84,7 +84,9 @@ public enum Capability {
             BlockRegistry blockRegistry = platform.getRegistries().getBlockRegistry();
             for (BlockType type : BlockType.REGISTRY) {
                 for (BlockState state : type.getAllStates()) {
-                    BlockStateIdAccess.register(state, blockRegistry.getInternalBlockStateId(state));
+                    BlockStateIdAccess.register(state,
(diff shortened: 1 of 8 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/Capability.java
```diff
@@ -84,7 +84,9 @@ void initialize(PlatformManager platformManager, Platform platform) {
             BlockRegistry blockRegistry = platform.getRegistries().getBlockRegistry();
             for (BlockType type : BlockType.REGISTRY) {
                 for (BlockState state : type.getAllStates()) {
-                    BlockStateIdAccess.register(state, blockRegistry.getInternalBlockStateId(state));
+                    BlockStateIdAccess.register(state,
+                        blockRegistry.getInternalBlockStateId(state)
+                            .orElse(BlockStateIdAccess.invalidId()));
```
me4502: Doesn't this return an int? So you can't orElse?

me4502: Wait nvm, that's on the registry

PR #549, opened 2020-01-31
Update draw.js

Gimlao:
Fixed the non working orientation and palette feature.

### DIFF
--- a/contrib/craftscripts/draw.js
+++ b/contrib/craftscripts/draw.js
@@ -133,12 +133,12 @@ context.checkArgs(1, 3, "<image> <orientation> <palette>");
 
 var f = context.getSafeOpenFile("drawings", argv[1], "png", ["png", "jpg", "jpeg", "bmp"]);
 var sess = context.remember();
-var upright = argv[2] === "v";
+var upright = argv[2] == "v";
 var colors = clothColors;
-if(argv[3] === "opt") {
+if(argv[3] == "opt") {
 	colors = clothColorsOpt;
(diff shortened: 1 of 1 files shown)


contrib/craftscripts/draw.js
```diff
@@ -133,12 +133,12 @@ context.checkArgs(1, 3, "<image> <orientation> <palette>");
 
 var f = context.getSafeOpenFile("drawings", argv[1], "png", ["png", "jpg", "jpeg", "bmp"]);
 var sess = context.remember();
-var upright = argv[2] === "v";
+var upright = argv[2] == "v";
```
me4502: All of these changes shouldn't be doing anything at all?

argv is a string array, and it's being compared to a string. In fact, these are all strings, so there is no case when it should fail on valid input.

Gimlao: I don't know !
Nothing happens when I use "v" "opt" and "optHD"
Then I replaced all "===" by "==", and now it works.

wizjany: While I agree that this shouldn't matter, I can reproduce sooo I guess?
for reference, `typeof(argv[3])` returns `object` (as does `typeof(argv)`) despite it being passed in as `String[]` from java (and `Context.javaToJS` correctly returns a NativeJavaArray for it).

unsure if this is something that changed at some point in rhino, since this definitely used to work? (using 1.7.11 now since octy added that for forge, while we used to use 1.7R2)

PR #556, opened 2020-02-21
Fluid masks for forge/fabric.

wizjany:
Entirely untested atm, just throwing ideas around.


me4502:
I wonder if this would make sense to use the concept of "Fluid" that sits aside from blocks, also it could be implemented on Bukkit with the adapter :P

wizjany:
yea so in theory MC has a fluid api now, but the backing implementation just checks the waterlogged state. but in bukkit we don't care here because the only things the liquid mask should return are water and lava. the waterlogged states are checked separately because //drain isn't allowed to just delete chests and fences and so on that are waterlogged, it has a separate thing that de-waterlogs.

so yes, if we wanted we could rip out the waterlogged stuff and the fluid mask (deprecate them i guess idk) and add fluid methods to our api, that way the backing impl takes care of it for us.

wizjany:
Pushed the second version in another commit, however the `adapter.adapt(BlockState)` methods currently re-apply properties each time instead of using internal ID like bukkit does. should potentially allow the adapters to do that (while still allowing them to calculate properties for first-time use).

PR #561, opened 2020-03-01
Add compatibility handlers for beds and banners for MCEdit Schematics

chickeneer:
These handlers are tested to work for a schematic saved in 1.12.2. Then loaded in 1.15.2.
Overall class layout used from SkullBlockCompatibilityHandler.

Prior to this PR. All schematic beds load as red beds. And all banners load as white banners. It is worth noting that for some crazy reason, mojang inverted the color integer values. That is handled as well.

I am not aware of any TileEntity blockstate conversion issues past these. Feel free to ping me on Discord if you want me to add more of these Compatibility handlers; I am willing.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java
@@ -109,10 +109,7 @@ public class BannerBlockCompatibilityHandler implements NBTCompatibilityHandler 
         return block;
     }
 
-    private static String convertBannerType(Integer oldType, boolean isWall) {
-        if (oldType == null) {
-            return null;
-        }
+    private static String convertBannerType(int oldType, boolean isWall) {
         String color;
         switch (oldType) {
             case 0:
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java
@@ -1,0 +1,171 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.extent.clipboard.io.legacycompat;
+
+import com.sk89q.jnbt.CompoundTag;
+import com.sk89q.jnbt.CompoundTagBuilder;
+import com.sk89q.jnbt.IntTag;
+import com.sk89q.jnbt.ListTag;
+import com.sk89q.jnbt.Tag;
+import com.sk89q.worldedit.registry.state.DirectionalProperty;
+import com.sk89q.worldedit.registry.state.IntegerProperty;
+import com.sk89q.worldedit.registry.state.Property;
+import com.sk89q.worldedit.world.block.BlockState;
+import com.sk89q.worldedit.world.block.BlockStateHolder;
+import com.sk89q.worldedit.world.block.BlockType;
+import com.sk89q.worldedit.world.block.BlockTypes;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class BannerBlockCompatibilityHandler implements NBTCompatibilityHandler {
+
+    private static final DirectionalProperty FacingProperty;
+    private static final IntegerProperty RotationProperty;
+
+    static {
+        DirectionalProperty tempFacing;
+        IntegerProperty tempRotation;
+        try {
+            tempFacing = (DirectionalProperty) (Property<?>) BlockTypes.WHITE_WALL_BANNER.getProperty("facing");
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java
```diff
@@ -0,0 +1,150 @@
...
+                    }
+                    return (B) state;
+                }
+            }
+        }
+        return block;
+    }
+
+    private String convertBannerType(Integer oldType, boolean isWall) {
+        switch (oldType) {
```
octylFractal: Can't you just get the color on the switch (or use a map/list), and append "wall_banner"/"banner" afterwards? Would be much simpler.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BedBlockCompatibilityHandler.java
```diff
@@ -0,0 +1,123 @@
...
+                    values.remove("color");
+                    return (B) state;
+                }
+            }
+        }
+        return block;
+    }
+
+    private String convertBedType(Integer oldType) {
+        switch (oldType) {
```
octylFractal: Similar here.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java
chickeneer: Could we. Yes. 
I chose this implementation strictly for consistency with existing code. Not that I prefer this over other implementations.
https://github.com/EngineHub/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/SkullBlockCompatibilityHandler.java#L82

Functionally it would end up doing the same thing. I would have no problem altering the PR to reflect that. Although, "Simpler" seems like an opinion in this case. The simplest is to accept the PR as presented since it works. But the Best may be to request an alteration for better code quality. Not to get caught up on word semantics.

chickeneer: Sorry, I didn't leave anything there to reply to. Is this definitely a change we want? It is an easy change, but I do not desire to bother with it unless it is actually wanted.

wizjany: Skull one is like that because 2 are skulls and 4 are heads. All of these are banners. Not a huge deal either way though imo.

chickeneer:
I have refactored the switch statements to cleaner.
Admittedly, also probably fixed an issue where passed in a null to switches said switch statements.

Let me know if anything else is needed.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java
```diff
@@ -0,0 +1,155 @@
...
+            case 10:
+                color = "lime";
+            case 11:
+                color = "yellow";
+            case 12:
+                color = "light_blue";
+            case 13:
+                color = "magenta";
+            case 14:
+                return "orange";
```
octylFractal: bug: shouldn't return here

chickeneer:
I was an idiot. Pushed up the wrong fix. The version which didn't even compile...
Re-pushed.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java
```diff
@@ -0,0 +1,171 @@
...
+                        values.put("Patterns", new ListTag(((ListTag) patternsTag).getType(), tempList));
+                    }
+                    return (B) state;
+                }
+            }
+        }
+        return block;
+    }
+
+    private static String convertBannerType(Integer oldType, boolean isWall) {
```
octylFractal: This doesn't have to be `Integer` -- the `IntTag` can never return `null`, if you look at it.

Same for beds :)

chickeneer: I am updating the PR to reflect this. Is there a reason that `getValue()` doesn't return type `int`. Not that it makes a practical difference

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BedBlockCompatibilityHandler.java
```diff
@@ -0,0 +1,141 @@
...
+    public <B extends BlockStateHolder<B>> B updateNBT(B block, Map<String, Tag> values) {
+        Tag typeTag = values.get("color");
+        if (typeTag instanceof IntTag) {
+            String bedType = convertBedType(((IntTag) typeTag).getValue());
+            if (bedType != null) {
+                BlockType type = BlockTypes.get("minecraft:" + bedType);
+                if (type != null) {
+                    BlockState state = type.getDefaultState();
+
+                    Property facingProp = type.getProperty("facing");
```
chickeneer: It is worth noting my ide flags these as `Raw use of parameterized class 'Property'`. Where in this case it should be `Property<Direction>`.
I matched the usage from https://github.com/EngineHub/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/SkullBlockCompatibilityHandler.java#L64
Would it be preferred for me to  be explicit about the parameterized type?

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java
octylFractal: Yes, inheritance. Super type returns Object.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BedBlockCompatibilityHandler.java
octylFractal: I think it's fine to leave as is. Personally, I would prefer fixing it, but that can be painful.

octylFractal:
Finally merging this. Thanks!

PR #1263, opened 2020-03-18
Optionally validate placed blocks

octylFractal:
This allows re-connection of fences, chests, panes, etc.; removal of
blocks in invalid states; and more!

The Connections side-effect already did this for neighbors, so this is
now renamed to Validation for accuracy, and left on by default.

This also fixes some inconsistencies between the Fabric & Forge
markAndNotifyBlock function.

Adapter PR: https://github.com/EngineHub/worldedit-adapters/pull/15


worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorld.java
```diff
@@ -170,7 +170,13 @@ public Path getStoragePath() {
...
     }
 
-    public void markAndNotifyBlock(World world, BlockPos pos, @Nullable WorldChunk worldChunk, net.minecraft.block.BlockState blockState,
+    /**
+     * This is a heavily modified function stripped from MC to apply worldedit-modifications.
+     *
+     * It is taken from Forge's World class, not Fabric's. This code should be consistent with the
+     * ForgeWorld code.
+     */
+    private void markAndNotifyBlock(World world, BlockPos pos, WorldChunk worldChunk, net.minecraft.block.BlockState blockState,
```
wizjany: can we rename either `blockstate` or `state`? or is this just supposed to mirror Forge's naming as closely as possible for copy-ability.
If that's the case, maybe just a comment like `// blockState = existing, state = new, blockState2 = ... hold up that's the existing one`

octylFractal: I've actually already corrected this while refactoring to allow Spigot to be more easily implemented.

PR #1276, opened 2020-03-30
Add OP-only blocks to "disallowed-blocks".

Mickey42302:
Hi,

I would like to suggest adding Command Blocks, Structure Blocks, Structure Voids, and Jigsaw Blocks to "disallowed-blocks" by default. On most servers, only players with Operator status can place/remove Command Blocks, Structure Blocks, and Jigsaw Blocks in Creative mode. However, they can still be placed by any player with access to WorldEdit using commands (for example, you could use "//set jigsaw"). Adding these to the default configuration would ensure that only players with permission to use disallowed blocks (or those with Operator status) can place them with WorldEdit.


octylFractal:
This does not prevent copying and pasting these blocks due to how disallowed blocks works. Additionally, it is not WorldEdit's goal to be a non-admin tool, it should only be given to people you trust. If you want to block these in a limited way, then feel free to add this configuration to your own server.

PR #1285, opened 2020-04-07
Delay command block active update if needed

### LINKED ISSUE #1284: Command block sessions access the world async
**Versions**
WorldEdit version: 7.1.0
Platform version: 1.15.2

**Describe the bug**
Command block session access the world async when checking for active status.

**To Reproduce**
1. Have command block sessions
2. Try to use WorldEdit until the server freezes

**Expected behavior**
WorldEdit functions normally with command block sessions

**Additional context**
https://paste.enginehub.org/nXe2zbOZ1

octylFractal:
If we are checking if a command block session is active, and we are not
on the main thread, schedule a task instead to update the active status
eventually, rather than waiting on the main thread.

Fixes #1284.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
@@ -19,8 +19,6 @@
 
 package com.sk89q.worldedit.bukkit;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import com.sk89q.worldedit.WorldEdit;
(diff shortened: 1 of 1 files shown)


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
```diff
@@ -146,16 +147,34 @@ public BlockCommandSender getSender() {
     @Override
     public SessionKey getSessionKey() {
         return new SessionKey() {
+
+            private volatile boolean active = true;
+
+            private void updateActive() {
```
wizjany, line 153: this at least needs to check if the chunk is loaded. unloaded block = inactive block.

PR #1295, opened 2020-04-13
Add image brush

KennyTV:
As per Wiz' request, copied over goBrush's image brush functionality.

Since goBrush has insultingly bad code, doesn't publish 1.15 builds publicly, and since I thought WorldEdit might benefit from such a functionality, this pr exists. :>

Using the ImageBrush, a bufferedimage is resized and put at the highest point of the clicked location. The intensity influences the brushes height, fill sets block above or removes block from below, flat flattens surrounding blocks at max to the the origin's height.
https://streamable.com/x0r400

Todo:
- tabcompletion for the imagenames (not familiar with your command lib, how can this be achieved?)
- ~~a manager for loading and getting the images (where to put the manager in structurally?)~~
- ~~brush manager holding (just threw it somewhere, also couldn't find _one_ center point to call the load from)~~
- ship default image brushes
- ~~(approach to image loading and resizing in general?)~~
- Rebase and squash when done

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/WorldEdit.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/WorldEdit.java
@@ -53,6 +53,7 @@ import com.sk89q.worldedit.scripting.RhinoCraftScriptEngine;
 import com.sk89q.worldedit.session.SessionManager;
 import com.sk89q.worldedit.util.Direction;
 import com.sk89q.worldedit.util.Location;
+import com.sk89q.worldedit.util.assets.ImageManager;
 import com.sk89q.worldedit.util.concurrency.EvenMoreExecutors;
 import com.sk89q.worldedit.util.concurrency.LazyReference;
 import com.sk89q.worldedit.util.eventbus.EventBus;
@@ -126,6 +127,7 @@ public final class WorldEdit {
             LazyReference.from(() -> new TranslationManager(
                     WorldEdit.getInstance().getPlatformManager().queryCapability(Capability.CONFIGURATION).getResourceLoader()
             ));
+    private final ImageManager imageManager = new ImageManager(this);
 
     private final BlockFactory blockFactory = new BlockFactory(this);
     private final ItemFactory itemFactory = new ItemFactory(this);
@@ -249,6 +251,15 @@ public final class WorldEdit {
      */
     public TranslationManager getTranslationManager() {
         return translationManager.getValue();
+    }
+
+    /**
+     * Return the image manager.
+     *
+     * @return the image manager
+     */
+    public ImageManager getImageManager() {
+        return imageManager;
     }
 
     /**
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
@@ -33,9 +33,11 @@ import com.sk89q.worldedit.command.tool.brush.CylinderBrush;
 import com.sk89q.worldedit.command.tool.brush.GravityBrush;
 import com.sk89q.worldedit.command.tool.brush.HollowCylinderBrush;
 import com.sk89q.worldedit.command.tool.brush.HollowSphereBrush;
+import com.sk89q.worldedit.command.tool.brush.ImageBrush;
 import com.sk89q.worldedit.command.tool.brush.OperationFactoryBrush;
 import com.sk89q.worldedit.command.tool.brush.SmoothBrush;
 import com.sk89q.worldedit.command.tool.brush.SphereBrush;
+import com.sk89q.worldedit.command.util.AsyncCommandBuilder;
 import com.sk89q.worldedit.command.util.CommandPermissions;
 import com.sk89q.worldedit.command.util.CommandPermissionsConditionGenerator;
 import com.sk89q.worldedit.command.util.CreatureButcher;
@@ -61,8 +63,10 @@ import com.sk89q.worldedit.session.ClipboardHolder;
 import com.sk89q.worldedit.session.request.RequestExtent;
 import com.sk89q.worldedit.util.HandSide;
 import com.sk89q.worldedit.util.TreeGenerator;
+import com.sk89q.worldedit.util.formatting.text.Component;
 import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import com.sk89q.worldedit.util.formatting.text.TranslatableComponent;
+import com.sk89q.worldedit.util.formatting.text.format.TextColor;
 import com.sk89q.worldedit.world.biome.BiomeType;
 import com.sk89q.worldedit.world.block.BlockTypes;
 import org.enginehub.piston.annotation.Command;
@@ -70,6 +74,10 @@ import org.enginehub.piston.annotation.CommandContainer;
 import org.enginehub.piston.annotation.param.Arg;
 import org.enginehub.piston.annotation.param.ArgFlag;
 import org.enginehub.piston.annotation.param.Switch;
+
+import java.awt.image.BufferedImage;
+import java.util.Set;
+import java.util.concurrent.Callable;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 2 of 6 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/ImageBrush.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    private double getHeightAt(BufferedImage image, int x, int y) {
+        int rgb = image.getRGB(x, y);
+
+        int red = rgb >>> 16 & 0xFF;
+        int green = rgb >>> 8 & 0xFF;
+        int blue = rgb & 0xFF;
+        if (red == 0 && green == 0 && blue == 0) {
```
me4502, line 102: Is there a reason for this short circuit?

I know I'm probably nit-picking here, but I feel that this would be "slower" than not having it, due to extra complexity preventing JVM optimisations.

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -320,6 +323,37 @@ public void butcherBrush(Player player, LocalSession session,
         player.printInfo(TranslatableComponent.of("worldedit.brush.butcher.equip", TextComponent.of((int) radius)));
     }
 
+    @Command(
+        name = "image",
+        desc = "Image brush, uses images as input"
+    )
+    @CommandPermissions("worldedit.brush.image")
+    public void imageBrush(Player player, LocalSession session,
```
octylFractal, line 331: It would be preferred to leave these methods as package-private, I know most of the others are `public` (I forgot to change them at the time) but new ones should be consistent.

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -320,6 +323,37 @@ public void butcherBrush(Player player, LocalSession session,
...
+                           @Arg(desc = "The name of the image")
+                               String imageName,
+                           @Arg(desc = "The size of the brush", def = "5")
+                               double radius,
+                           @Arg(desc = "The intensity of the brush", def = "5")
+                               double intensity,
+                           @Arg(desc = "Whether to fill or clear blocks", def = "true")
+                               boolean fillMode,
+                           @Arg(desc = "Whether to flatten blocks to the given height", def = "false")
+                               boolean flatMode) throws WorldEditException {
```
octylFractal, line 341: These arguments should be `@Switch`es, since they are just boolean values.

I also suggest making `radius` and `intensity` `@ArgFlag`s, but that's a more controversial choice IMO. We do have some other commands that require default arguments to be re-provided right now.

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -320,6 +323,37 @@ public void butcherBrush(Player player, LocalSession session,
...
+        BufferedImage image = worldEdit.getImageBrushManager().getImage(imageName);
+        if (image == null) {
+            player.printInfo(TranslatableComponent.of("worldedit.brush.image.imageList", TextComponent.of(imageName))); //TODO add message
```
octylFractal, line 344: What does this line even mean? There's no "listing" going on here that I can see.

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/ImageBrush.java
```diff
@@ -0,0 +1,111 @@
...
+        BufferedImage resizedPattern = new BufferedImage(diameter, diameter, 1);
+        Graphics2D graphic = null;
+        try {
+            graphic = resizedPattern.createGraphics();
+            graphic.drawImage(this.image, 0, 0, diameter, diameter, null);
+        } finally {
+            graphic.dispose();
+        }
+
+        double random = RANDOM.nextDouble();
```
octylFractal, line 62: This random feature might be a good candidate for another `@Switch`, I can see it being annoying for making exact replicas.

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/ImageBrush.java
```diff
@@ -0,0 +1,111 @@
...
+
+                BaseBlock baseBlock = fill ? editSession.getBlock(block).toBaseBlock() : null;
+                for (int y = 0; y < height; y++) {
+                    if (fill) {
+                        // Only go up to the origin's level if flat mode is enabled
+                        if (!flat || block.getY() + y <= position.getY()) {
+                            editSession.setBlock(block.withY(block.getY() + y), baseBlock);
+                        }
+                    } else if ((!flat || block.getY() - y > position.getY()) && block.getY() - y >= 0
+                            && (editSession.getMask() == null || editSession.getMask().test(block.withY(block.getY() - y)))) {
```
octylFractal, line 87: I don't quite get the reasoning behind doing the mask test here -- it's going to happen in just a few calls in `setBlock` anyways.

worldedit-core/src/main/java/com/sk89q/worldedit/util/imagebrush/ImageBrushManager.java
```diff
@@ -0,0 +1,99 @@
...
+        File brushDir = new File(worldEdit.getPlatformManager().getConfiguration().getWorkingDirectory(), "brushes");
+        // Only load brushes if the directory was manually created
+        if (!brushDir.exists() || !brushDir.isDirectory()) {
+            return;
+        }
+
+        for (File file : brushDir.listFiles()) {
+            String fileName = file.getName();
+            if (fileName.indexOf(' ') != -1) {
+                logger.warn("Brush file '{}' contains spaces.", fileName);
```
octylFractal, line 64: This... isn't an issue? Arguments can contain spaces by using `""`, so I don't see a reason to ban spaces.

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
KennyTV: Thought of sending a list of brushes here, similar to how WG sends a list of flags on failure (https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java#L530)
... it is now implemented as I'd have invisioned it (with the same format/alternating colors as the WG flag list), not sure if the callable should be moved to its extra class.

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
KennyTV: As far as I can see, none of the other types use argflags for the radius or similar, so that would hit the consistency a bit with just this one

worldedit-core/src/main/java/com/sk89q/worldedit/WorldEdit.java
```diff
@@ -240,6 +242,15 @@ public TranslationManager getTranslationManager() {
         return translationManager;
     }
 
+    /**
+     * Return the image brush manager.
+     *
+     * @return the image brush manager
+     */
+    public ImageBrushManager getImageBrushManager() {
```
me4502, line 250: I'm unsure whether this should be specifically an ImageBrushManager

A more generic ImageManager, or maybe even a manager for generic assets that could be used in WE stuff (like models etc) could make more sense.

worldedit-core/src/main/java/com/sk89q/worldedit/util/imagebrush/ImageBrushManager.java
```diff
@@ -0,0 +1,105 @@
...
+        this.worldEdit = worldEdit;
+    }
+
+    /**
+     * Loads all files from the brushes directory into the images cache.
+     */
+    public void load() {
+        images.clear();
+
+        File brushDir = new File(worldEdit.getPlatformManager().getConfiguration().getWorkingDirectory(), "brushes");
```
me4502, line 56: I don't feel brushes is the correct directory for this. What if we wanted a model brush or a script brush in the future?

I'd personally prefer for this to be more generic

worldedit-core/src/main/java/com/sk89q/worldedit/util/imagebrush/ImageBrushManager.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Loads and caches image files from WorldEdit's brushes directory.
+ */
+public class ImageBrushManager {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    private final Map<String, BufferedImage> images = new HashMap<>();
```
me4502, line 43: I'm not sure we want to have these loaded for the entire duration of the server being loaded?

It may be worth at least lazy-loading these, and potentially cleaning them up if they haven't been used for X amount of time

worldedit-core/src/main/java/com/sk89q/worldedit/WorldEdit.java
KennyTV: Did _something_, but sure if the changes are close to what you imagined :p

MattBDev:
One of my biggest issues with this implementation right now is that it only loads images locally. Many users in FAWE like the image brush because they can load an image from a URL and use it with their brush.

me4502:
I'm unsure how I feel about that from a security perspective

worldedit-core/src/main/java/com/sk89q/worldedit/util/assets/AssetManager.java
```diff
@@ -0,0 +1,118 @@
...
+        int extensionIndex = path.lastIndexOf('.');
+        String assetName = extensionIndex != -1 ? path.substring(0, extensionIndex) : path;
+        assetName = assetName.toLowerCase(Locale.ROOT);
+
+        T cached = assets.getIfPresent(assetName);
+        if (cached != null) {
+            return cached;
+        }
+
+        File assetsDir = new File(worldEdit.getPlatformManager().getConfiguration().getWorkingDirectory(), "assets");
```
me4502, line 74: It's probably worthwhile having the "assets" directory name something that can be passed into the constructor

me4502:
Hey could you please squash this? I'm interested in extending it to be a complete asset system, so I'd probably merge it into another branch and would rather keep the author on the commits without making it too spammy in history

KennyTV:
Squashed and rebased 😛

me4502:
Superseded by https://github.com/EngineHub/WorldEdit/commit/ec306ccaf130170c5347a966f76fe2f28d18281a

PR #1301, opened 2020-04-16
Add tracing feature to EditSessions

octylFractal:
This helps track down what plugin (or more specifically, extent) is blocking actions from happening.

Suggestions are welcome on every part of this, as it was just the simplest way I could see today. Maybe there's something smarter we can be doing.

I did wrap every extent, just in case we have a bug in one of our own extents. It's better to be safe than sorry.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSessionBuilder.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSessionBuilder.java
@@ -22,7 +22,6 @@ package com.sk89q.worldedit;
 import javax.annotation.Nullable;
 
 import com.google.common.base.Preconditions;
-import com.sk89q.worldedit.event.extent.EditSessionEvent;
 import com.sk89q.worldedit.extension.platform.Actor;
 import com.sk89q.worldedit.extension.platform.Locatable;
 import com.sk89q.worldedit.extent.Extent;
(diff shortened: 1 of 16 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/EditSessionBuilder.java
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * A builder-style factory for {@link EditSession EditSessions}.
+ */
+public final class EditSessionBuilder {
+
+    private final EventBus eventBus;
+    @Nullable
+    private World world;
+    private int maxBlocks = -1;
+    private Actor actor;
```
wizjany, line 42: `@Nullable` actor?

worldedit-core/src/main/java/com/sk89q/worldedit/command/WorldEditCommands.java
```diff
@@ -138,6 +143,45 @@ public void report(Actor actor,
...
+        try (EditSession ignored = session) {
+            if (action.test.test(session, localSession.getPlacementPosition(actor))) {
+                actor.printInfo(TranslatableComponent.builder("worldedit.trace.success")
+                    .args(TextComponent.of(action.toString()))
+                    .build());
+                return;
+            }
+        }
+        List<TracingExtent> tracingExtents = session.getTracingExtents();
+        assert tracingExtents != null;
```
wizjany, line 166: leftover debug?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/AbstractPlayerActor.java
```diff
@@ -334,7 +334,8 @@ public void floatAt(int x, int y, int z, boolean alwaysGlass) {
             BlockVector3 spot = BlockVector3.at(x, y - 1, z);
             final World world = getWorld();
             if (!world.getBlock(spot).getBlockType().getMaterial().isMovementBlocker()) {
-                try (EditSession session = WorldEdit.getInstance().getEditSessionFactory().getEditSession(world, 1, this)) {
+                try (EditSession session =
```
wizjany, line 337: is this break necessary? looks like you tried newline/indenting the original line and then converted to the shortcut method.

worldedit-core/src/main/java/com/sk89q/worldedit/command/WorldEditCommands.java
octylFractal: No, just convincing IntelliJ that it's not `null`. I can leave it out, I just don't like leaving the warnings in IntelliJ :P

PR #1327, opened 2020-05-03
feat(i18n): use a translation loader to allow CB/WG to use WorldEdit's TranslationManager

me4502:


worldedit-core/src/main/java/com/sk89q/worldedit/util/io/WorldEditResourceLoader.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public URL getResource(Class<?> clazz, String pathname) throws IOException {
+        return clazz.getResource(pathname);
+    }
+
+    @Override
+    public URL getRootResource(String pathname) throws IOException {
+        return ResourceLoader.class.getResource("/" + pathname);
```
me4502, line 43:
```suggestion
        return WorldEditResourceLoader.class.getResource("/" + pathname);
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/ResourceLoader.java
```diff
@@ -19,36 +19,44 @@
...
-        return url;
-    }
+    /**
+     * Gets the {@link File} reference to this
+     * local resource. The file may not exist.
+     *
+     * @param pathname The pathname
+     * @return The file reference
+     */
+    File getLocalResource(String pathname);
```
octylFractal, line 61: This should probably return `Path` instead, no need to add the old IO apis to new methods.

PR #1350, opened 2020-05-29
Use a long to refer to the volume of a region to prevent overflow

### LINKED ISSUE #1344: Area calculations should not overflow
**Versions**
WorldEdit version: 7.2.0-SNAPSHOT, `#5003`
Platform version: N/A

**Describe the bug**
Selecting a large enough area results in integer overflow in the area calculation, and a negative number being printed/returned for the area.

**To Reproduce**
1. `//pos1 0,0,0`
2. `//pos2 10000,255,10000`

**Expected behavior**
The area is listed next to the new position.

Note that adding this will require an API break, perhaps something we can mitigate by using a different name for the new function.

me4502:
I've also renamed it to getVolume, as that's more accurate.

One potential concern here is with double precision on the non-cuboid selection types when it comes to the larger sizes. Doubles are less precise above integer bounds and therefore may produce incorrect results for super large selections, not sure we should care tbh.

Fixes #1344

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/SelectionPoint2DEvent.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/SelectionPoint2DEvent.java
@@ -27,8 +27,9 @@ public class SelectionPoint2DEvent implements CUIEvent {
     protected final int id;
     protected final int blockX;
     protected final int blockZ;
-    protected final int area;
+    protected final long area;
 
+    @Deprecated
     public SelectionPoint2DEvent(int id, BlockVector2 pos, int area) {
         this.id = id;
         this.blockX = pos.getX();
@@ -36,7 +37,22 @@ public class SelectionPoint2DEvent implements CUIEvent {
         this.area = area;
     }
 
+    @Deprecated
     public SelectionPoint2DEvent(int id, BlockVector3 pos, int area) {
+        this.id = id;
+        this.blockX = pos.getX();
+        this.blockZ = pos.getZ();
+        this.area = area;
+    }
+
+    public SelectionPoint2DEvent(int id, BlockVector2 pos, long area) {
+        this.id = id;
+        this.blockX = pos.getX();
+        this.blockZ = pos.getZ();
+        this.area = area;
+    }
+
+    public SelectionPoint2DEvent(int id, BlockVector3 pos, long area) {
         this.id = id;
         this.blockX = pos.getX();
         this.blockZ = pos.getZ();
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
@@ -82,7 +82,7 @@ public class EllipsoidRegion extends AbstractRegion {
         return center.toVector3().add(getRadius()).toBlockPoint();
     }
 
-    private final BigDecimal ELLIPSOID_BASE_MULTIPLIER = BigDecimal.valueOf((4.0 / 3.0) * Math.PI);
+    private static final BigDecimal ELLIPSOID_BASE_MULTIPLIER = BigDecimal.valueOf((4.0 / 3.0) * Math.PI);
 
     @Override
     public long getVolume() {
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
@@ -82,10 +82,11 @@ public class EllipsoidRegion extends AbstractRegion {
         return center.toVector3().add(getRadius()).toBlockPoint();
     }
 
+    private final BigDecimal ELLIPSOID_BASE_MULTIPLIER = BigDecimal.valueOf((4.0 / 3.0) * Math.PI);
 
     @Override
     public long getVolume() {
-        return BigDecimal.valueOf((4.0 / 3.0) * Math.PI)
+        return ELLIPSOID_BASE_MULTIPLIER
                 .multiply(BigDecimal.valueOf(radius.getX()))
                 .multiply(BigDecimal.valueOf(radius.getY()))
                 .multiply(BigDecimal.valueOf(radius.getZ()))
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
@@ -81,8 +81,8 @@ public class EllipsoidRegion extends AbstractRegion {
     }
 
     @Override
-    public int getArea() {
-        return (int) Math.floor((4.0 / 3.0) * Math.PI * radius.getX() * radius.getY() * radius.getZ());
+    public long getVolume() {
+        return (long) Math.floor((4.0 / 3.0) * Math.PI * radius.getX() * radius.getY() * radius.getZ());
     }
 
     @Override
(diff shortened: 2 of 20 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/SelectionPoint2DEvent.java
```diff
@@ -27,22 +27,38 @@
     protected final int id;
     protected final int blockX;
     protected final int blockZ;
-    protected final int area;
+    protected final long area;
 
+    @Deprecated
     public SelectionPoint2DEvent(int id, BlockVector2 pos, int area) {
```
me4502, line 33: The old code still works fine with these gone, but I fear that it'll break binary compatibility if I do. Am I correct in assuming this?

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Polygonal2DRegion.java
```diff
@@ -207,7 +207,7 @@ public int getArea() {
             j = i;
         }
 
-        return (int) Math.floor(Math.abs(area * 0.5)
+        return (long) Math.floor(Math.abs(area * 0.5)
```
octylFractal, line 210: This should also be changed to `BigDecimal`, and I think the area computation above also needs to be at least promoted to `long` to ensure we don't lose the edge cases:
```
>>> (3_000_000 * 3_000_000) > 2**32
True
```

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
```diff
@@ -61,8 +61,22 @@
      * Get the number of blocks in the region.
      *
      * @return number of blocks
+     * @deprecated use {@link Region#getVolume()} to prevent overflows
      */
-    int getArea();
+    @Deprecated
+    default int getArea() {
+        return (int) getVolume();
```
octylFractal, line 68: This is a highly dangerous setup -- I suggest we add some sort of check on `getVolume()` to indicate the problem instead of letting it crash with a `StackOverflowError` if someone overrides neither `getArea()` nor `getVolume()`.

We can reflect onto the current class and check if the `clazz.getMethod("getArea")` points to this method, defined in `Region` -- this will catch it being overriden at any point in the hierarchy.

Having this check in `getVolume()` shouldn't be a huge deal, because for all the core classes, we'll have overriden `getVolume()` and the check will never happen. It's just a compatibility for older mods that might be implementing `Region`.

worldedit-core/src/main/java/com/sk89q/worldedit/regions/RegionSelector.java
```diff
@@ -134,8 +134,22 @@
...
+        return (int) getVolume();
+    }
+
+    /**
+     * Get the number of blocks inside the region.
+     *
+     * @return number of blocks, or -1 if undefined
+     */
+    default long getVolume() {
+        // TODO Remove default once getArea is removed
```
octylFractal, line 150: Same deal here as with `Region`.

worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
```diff
@@ -80,9 +82,15 @@ public BlockVector3 getMaximumPoint() {
         return center.toVector3().add(getRadius()).toBlockPoint();
     }
 
+
     @Override
-    public int getArea() {
-        return (int) Math.floor((4.0 / 3.0) * Math.PI * radius.getX() * radius.getY() * radius.getZ());
+    public long getVolume() {
+        return BigDecimal.valueOf((4.0 / 3.0) * Math.PI)
```
octylFractal, line 88: Curious as to why you made `PI` a constant in the other class, but not this computation.

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Polygonal2DRegion.java
```diff
@@ -197,18 +199,24 @@ public BlockVector3 getMaximumPoint() {
...
+        BigDecimal area = BigDecimal.ZERO;
         int i, j = points.size() - 1;
 
         for (i = 0; i < points.size(); ++i) {
-            area += (points.get(j).getBlockX() + points.get(i).getBlockX())
-                    * (points.get(j).getBlockZ() - points.get(i).getBlockZ());
+            area = area.add(
+                    BigDecimal.valueOf(points.get(j).getBlockX() + points.get(i).getBlockX())
+                            .multiply(BigDecimal.valueOf(points.get(j).getBlockZ() - points.get(i).getBlockZ()))
+            );
```
octylFractal, line 210: This looks a little ugly -- perhaps pull out the X and Z parts to variables with names?

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Polygonal2DRegion.java
```diff
@@ -197,18 +199,24 @@ public BlockVector3 getMaximumPoint() {
...
             j = i;
         }
 
-        return (int) Math.floor(Math.abs(area * 0.5)
-                * (maxY - minY + 1));
+        return area
+                .multiply(BigDecimal.valueOf(0.5))
+                .abs()
+                .setScale(0, RoundingMode.FLOOR)
+                .multiply(BigDecimal.valueOf(maxY - minY + 1))
```
octylFractal, line 218: Any reason you did this post-floor, but still in the `BigDecimal`? you might as well do this multiplication after you get the `longValue()` instead.

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
```diff
@@ -61,8 +61,29 @@
...
+     * @return number of blocks
+     */
+    default long getVolume() {
+        // TODO Remove default status when getArea is removed.
+        try {
+            if (getClass().getMethod("getArea").getDeclaringClass().equals(this.getClass())) {
+                throw new IllegalStateException("Class " + getClass().getName() + " must override either getVolume or getArea.");
+            }
+        } catch (NoSuchMethodException e) {
+            e.printStackTrace();
```
octylFractal, line 83: IMO we should just throw here (wrapped in `AssertionException`), or completely suppress. This should never occur anyways.

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
```diff
@@ -61,8 +61,29 @@
...
+
+    /**
+     * Get the number of blocks in the region.
+     *
+     * @return number of blocks
+     */
+    default long getVolume() {
+        // TODO Remove default status when getArea is removed.
+        try {
+            if (getClass().getMethod("getArea").getDeclaringClass().equals(this.getClass())) {
```
octylFractal, line 79: This check is wrong:
```suggestion
            if (getClass().getMethod("getArea").getDeclaringClass().equals(Region.getClass())) {
```

worldedit-core/src/main/java/com/sk89q/worldedit/regions/RegionSelector.java
```diff
@@ -134,8 +134,29 @@
...
+        try {
+            if (getClass().getMethod("getArea").getDeclaringClass().equals(this.getClass())) {
+                throw new IllegalStateException("Class " + getClass().getName() + " must override either getVolume or getArea.");
+            }
+        } catch (NoSuchMethodException e) {
+            e.printStackTrace();
+        }
```
octylFractal, line 157: Same issues as the Region class.

worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
me4502: I made it a constant because it was its own BigDecimal, this one I could merge it with the other two. Although this is still not going to change so I could pull this one out too tbh.

octylFractal: Yea, that's what I was thinking, is making `BigDecimal.valueOf(4.0 / 3.0 * Math.PI)` a constant.

PR #1359, opened 2020-06-11
Count wandering traders as NPCs for /butcher.

### LINKED ISSUE #1347: Wandering Villagers are including when butchering
**Versions**

WorldEdit version: 7.1.0;8e55131
Platform version: Paper 1.15.2 b304

**Describe the bug**

  When using the command "/butcher", **Wandering Villagers** are also removed alongside the "mobs"

**To Reproduce**

1. use /butcher

**Expected behavior**

  Wandering Villagers are considered a villager or at least a passive mob.

wizjany:
Fixes #1347.

### DIFF
--- a/worldedit-bukkit/build.gradle.kts
+++ b/worldedit-bukkit/build.gradle.kts
@@ -22,10 +22,10 @@ configurations.all {
 dependencies {
     "api"(project(":worldedit-core"))
     "api"(project(":worldedit-libs:bukkit"))
-    "api"("org.bukkit:bukkit:1.13.2-R0.1-SNAPSHOT") {
+    "api"("org.bukkit:bukkit:1.15.2-R0.1-SNAPSHOT") {
         exclude("junit", "junit")
     }
-    "compileOnly"("com.destroystokyo.paper:paper-api:1.13.2-R0.1-SNAPSHOT")
+    "compileOnly"("com.destroystokyo.paper:paper-api:1.15.2-R0.1-SNAPSHOT")
     "implementation"("io.papermc:paperlib:1.0.2")
     "compileOnly"("com.sk89q:dummypermscompat:1.10")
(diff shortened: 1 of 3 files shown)


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitEntityProperties.java
```diff
@@ -46,6 +47,18 @@
 
 class BukkitEntityProperties implements EntityProperties {
 
+    private static final boolean hasAbstractVillager;
```
octylFractal, line 50: `HAS_ABSTRACT_VILLAGER`?

worldedit-bukkit/src/test/java/com/sk89q/wepif/TestOfflinePermissible.java
```diff
@@ -185,4 +188,94 @@ public Location getBedSpawnLocation() {
     public Map<String, Object> serialize() {
         throw new UnsupportedOperationException("Not supported yet.");
     }
+
```
octylFractal, line 191: Curious if this might be easier to do with a mock or a Proxy.

wizjany: maybe in the long run.

worldedit-bukkit/build.gradle.kts
```diff
@@ -22,10 +22,10 @@ configurations.all {
 dependencies {
     "api"(project(":worldedit-core"))
     "api"(project(":worldedit-libs:bukkit"))
-    "api"("org.bukkit:bukkit:1.13.2-R0.1-SNAPSHOT") {
+    "api"("org.bukkit:bukkit:1.15.2-R0.1-SNAPSHOT") {
```
me4502, line 25: I wonder if there's something we can do to ensure that we don't use APIs that will break 1.13.2 compat

Issue #1358, opened 2020-06-10
Add support for 3D biomes


me4502:
**Is your feature request related to a problem? Please describe.**
As of Minecraft 1.15(?), biomes are stored in 3D space. This should be exposed in the plugin and API.  

**Describe the solution you'd like**
A direct 3D biome system should be implemented, that just hands off to the underlying implementation. On versions prior to 1.15, this would set the entire columns. On versions above, it'd set the 4x4x4 cube to that biome.

Some consideration would need to go into how we'd handle a case where a pattern sets multiple various biomes in the same "biome space" in a single edit.

PR #1361, opened 2020-06-13
Add 3D biomes support (Adds #1358)

me4502:
TODO:

-  [x]  Bukkit support
-  [x]  Fixing history
-  [x]  Modifying locations that use the biomes to iterate over 3D areas
-  [x]  Clipboard support

Adds #1358


worldedit-core/src/main/java/com/sk89q/worldedit/function/operation/ForwardExtentCopy.java
```diff
@@ -307,13 +303,12 @@ public Operation resume(RunContext run) throws WorldEditException {
...
             List<Operation> ops = Lists.newArrayList(blockVisitor);
 
-            if (copyingBiomes && region instanceof FlatRegion) { // double-check here even though we checked before
-                ExtentBiomeCopy biomeCopy = new ExtentBiomeCopy(source, from.toBlockVector2(),
-                        destination, to.toBlockVector2(), currentTransform);
-                Mask2D biomeMask = sourceMask.toMask2D();
-                FlatRegionFunction biomeFunction = biomeMask == null ? biomeCopy
-                        : new FlatRegionMaskingFilter(biomeMask, biomeCopy);
-                FlatRegionVisitor biomeVisitor = new FlatRegionVisitor(((FlatRegion) region), biomeFunction);
+            if (copyingBiomes) { // double-check here even though we checked before
```
octylFractal, line 306: Comment is obsolete AND `setCopyingBiomes` needs the check removed.

worldedit-core/src/main/java/com/sk89q/worldedit/function/pattern/BiomePattern.java
```diff
@@ -31,7 +32,33 @@
...
+     *     Note: This will not be default in WE8.
+     * </p>
+     *
+     * @param position the position
+     * @return a biome
+     */
+    default BiomeType applyBiome(BlockVector3 position) {
+        // TODO Remove default in WE8
+        try {
+            if (getClass().getMethod("apply", BlockVector2.class).getDeclaringClass().equals(BiomePattern.class)) {
```
octylFractal, line 56: At this point I'm wondering if extracting this to a utility is worthwile. Something like `DeprecationUtil`:
```java
public static void checkDelegatingOverride(Class<?> declaringClass, Class<?> implementingClass, String newMethodName, String oldMethodName, Class<?>... parameterTypes) {
  try {
    Method method = implementingClass.getMethod(oldMethodName, parameterTypes);
    if (method.getDeclaringClass().equals(declaringClass)) {
      throw new IllegalStateException("Class " + implementingClass.getName() + " must override " + newMethodName);
    }
  } catch (NoSuchMethodException e) {
    throw new AssertionError(e);
  }
}
```

Usage:
```java
DeprecationUtil.checkDelegatingOverride(getClass(), BiomePattern.class, "applyBiome(BlockVector3)", "apply", BlockVector2.class);
```

Might be nicer to do some sort of stack walking to more easily acquire the new method name, but that's the general idea.

worldedit-core/src/main/java/com/sk89q/worldedit/history/change/BiomeChange.java
```diff
@@ -62,6 +64,7 @@ public BiomeChange(BlockVector2 position, BiomeType previous, BiomeType current)
      * @return the position
      */
     public BlockVector2 getPosition() {
+        // TODO Decide what to do here
```
octylFractal, line 67: Don't quite get this comment.

worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/ArbitraryBiomeShape.java
```diff
@@ -82,14 +78,14 @@ public ArbitraryBiomeShape(Region extent) {
...
+    private BiomeType getBiomeCached(int x, int y, int z, BiomeType baseBiome) {
+        final int index = (y - cacheOffsetY) + (z - cacheOffsetZ) * cacheSizeY + (x - cacheOffsetX) * cacheSizeY * cacheSizeZ;
 
         final BiomeType cacheEntry = cache[index];
         if (cacheEntry == null) {// unknown, fetch material
-            final BiomeType material = getBiome(x, z, baseBiome);
+            final BiomeType material = getBiome(x, y, z, baseBiome);
             if (material == null) {
                 // outside
                 cache[index] = BiomeTypes.THE_VOID;
```
octylFractal, line 91: Use of THE_VOID here is suspect, perhaps we can use `null` or something? I don't really like the fact that this is inconsistent _and_ can cause issues if used in a void world.

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -29,7 +29,7 @@
     "worldedit.brush.operation.equip": "Set brush to {0}.",
     "worldedit.brush.none.equip": "Brush unbound from your current item.",
 
-    "worldedit.setbiome.changed": "Biomes were changed in {0} columns. You may have to rejoin your game (or close and reopen your world) to see a change.",
+    "worldedit.setbiome.changed": "Biomes were changed for {0} blocks. You may have to rejoin your game (or close and reopen your world) to see a change.",
```
octylFractal, line 32: Might be better to say:
```suggestion
    "worldedit.setbiome.changed": "Biomes were changed for approximately {0} blocks. You may have to rejoin your game (or close and reopen your world) to see a change.",
```

PR #1362, opened 2020-06-13
Fabric 1.16

octylFractal:
This is a mostly-complete update to 1.16 for Fabric.

The biggest change is the brand new regen algorithm, which is much faster & safer than before due to loading a new server world from the ground up, and then submitting load tasks all at once for maximum async loading.

This PR will become un-drafted when 1.16 is released and this PR is updated to it.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -30,9 +30,9 @@ configurations.all {
     resolutionStrategy {
         // Fabric needs this.
         force(
-            "commons-io:commons-io:2.5",
-            "org.ow2.asm:asm:7.1",
-            "org.ow2.asm:asm-commons:7.1"
+            "commons-io:commons-io:2.6",
+            "org.ow2.asm:asm:8.0.1",
+            "org.ow2.asm:asm-commons:8.0.1"
         )
     }
 }
--- a/config/checkstyle/import-control.xml
+++ b/config/checkstyle/import-control.xml
@@ -85,6 +85,8 @@
       <allow pkg="org.spongepowered" />
       <allow pkg="com.mojang.brigadier" />
       <allow pkg="com.mojang.datafixers" />
+      <allow pkg="com.mojang.serialization" />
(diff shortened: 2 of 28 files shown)


phemmer:
I see from the description that you consider this "mostly-complete". Is it expected to be at least somewhat functional at this point? If not, then you can ignore the rest of this, and I'll go wait for a release.

Anyway, I'm starting to experiment with 1.16, and attempting to load WorldEdit ([build #5051](https://builds.enginehub.org/job/worldedit/14574)) throws an exception:
```
Exception in thread "main" java.lang.RuntimeException: Failed to read accessWidener file from mod worldedit
	at net.fabricmc.loader.transformer.accesswidener.AccessWidener.loadFromMods(AccessWidener.java:60)
	at net.fabricmc.loader.launch.knot.Knot.init(Knot.java:130)
	at net.fabricmc.loader.launch.knot.KnotClient.main(KnotClient.java:26)
Caused by: java.lang.RuntimeException: Namespace (named) does not match current runtime namespace (intermediary)
	at net.fabricmc.loader.transformer.accesswidener.AccessWidener.read(AccessWidener.java:78)
	at net.fabricmc.loader.transformer.accesswidener.AccessWidener.loadFromMods(AccessWidener.java:58)
	... 2 more
```

This is with Minecraft 1.16-pre5, & Fabric-loader 0.8.7+build.201-1.16-pre5

octylFractal:
I mistakenly though Fabric would map it for me by default, fixed.

i509VCB:
Just a heads up, commands are a crapshoot right now and where you register commands currently (during server start) is too late for anyone using worldedit in functions [all functions that call worldedit fail].

This is a big issue on fabric api as well so I am trying to find a solution to this issue.

octylFractal:
Yea, not really related to this PR, you can file an issue if you want but IIRC it doesn't even work on Bukkit, we need to reload the resource manager after we register or get Mojang to change when the commands are serialized to functions.

PR #1370, opened 2020-06-22
Add clearblocks.py CraftScript

JsKingBoo:
The contents of this CraftScript is less concerned about being useful and more focused on demonstrating the CraftScript API for aspiring developers. Thus the code performs an operation that is short and easy to understand.

### DIFF
--- a/contrib/craftscripts/clearblocks.js
+++ b/contrib/craftscripts/clearblocks.js
@@ -1,0 +1,60 @@
+// Marks all blocks that have at least 3 exposed sides (adjacent to an air block)
+// and then deletes those blocks.
+// Based on jsa's MCEdit Filter clearBlocks.py
+
+importClass(Packages.com.sk89q.worldedit.world.block.BlockTypes)
+
+context.checkArgs(0, -1, "")
+
+var blocks = context.remember()
+var region = context.getSession().getRegionSelector(player.getWorld()).getRegion()
+
+var min = region.getMinimumPoint()
+var xi = region.getWidth()
+var yi = region.getHeight()
+var zi = region.getLength()
+
+function sidesExposed(block) {
+  if (blocks.getBlock(block).getBlockType().equals(BlockTypes.AIR)) {
+    return 0
+  }
+  var sides = 0
+  if (blocks.getBlock(block.add(0, 0, 1)).getBlockType().equals(BlockTypes.AIR)) {
+    sides += 1
+  }
+  if (blocks.getBlock(block.add(0, 1, 0)).getBlockType().equals(BlockTypes.AIR)) {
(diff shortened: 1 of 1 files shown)


octylFractal:
`.py` 🤔

JsKingBoo:
MCEdit filter roots run deep

contrib/craftscripts/clearblocks.js
```diff
@@ -0,0 +1,60 @@
+// Marks all blocks that have at least 3 exposed sides (adjacent to an air block)
+// and then deletes those blocks.
+// Based on jsa's MCEdit Filter clearBlocks.py
+
+importClass(Packages.com.sk89q.worldedit.world.block.BlockTypes)
+
+context.checkArgs(0, -1, "")
+
+var blocks = context.remember()
```
octylFractal, line 9: To make this "easy to understand", this should probably be named `editSession`.

contrib/craftscripts/clearblocks.js
```diff
@@ -0,0 +1,60 @@
...
+var region = context.getSession().getRegionSelector(player.getWorld()).getRegion()
+
+var min = region.getMinimumPoint()
+var xi = region.getWidth()
+var yi = region.getHeight()
+var zi = region.getLength()
```
octylFractal, line 15: I'm not sure if these matter as much, but why not name these `width`, `height`, `length`, rather than the somewhat crypting "x index", which is what I presume `xi` means.

contrib/craftscripts/clearblocks.js
JsKingBoo: Variable renamed

contrib/craftscripts/clearblocks.js
JsKingBoo: Variable renamed

wizjany:
🤔

JsKingBoo:
smh my head

fixed comment

me4502 (approved):

contrib/craftscripts/clearblocks.js
```diff
@@ -0,0 +1,60 @@
...
+
+var editSession = context.remember()
+var region = context.getSession().getRegionSelector(player.getWorld()).getRegion()
+
+var min = region.getMinimumPoint()
+var width = region.getWidth()
+var height = region.getHeight()
+var length = region.getLength()
+
+function sidesExposed(block) {
```
octylFractal, line 17: This entire function can be replaced by the `#surface` mask, don't recall the exact implementing class but it should be sufficient. I would rather not add more clunky code for us to maintain in the craftscripts dir.

octylFractal (changes requested):

PR #1404, opened 2020-06-30
Fix or suppress deprecation warnings in most code

octylFractal:
Didn't touch commands or some of Sponge


worldedit-core/src/main/java/com/sk89q/worldedit/command/ChunkCommands.java
```diff
@@ -162,11 +162,15 @@ public void deleteChunks(Actor actor, World world, LocalSession session,
             throw new StopExecutionException(TextComponent.of("Failed to write chunk list: " + e.getMessage()));
         }
 
-        actor.print(String.format("%d chunk(s) have been marked for deletion the next time the server starts.",
-                newBatch.getChunkCount()));
+        actor.print(TextComponent.of(
```
me4502, line 165: I was using the deprecation marker for these to indicate it's not translatable yet. Maybe worth making these translatable rather than fixing deprecation by using TextComponent directly

PR #1433, opened 2020-07-11
Fix 3D biome support testing

octylFractal:
3D biomes always work on dedicated servers, at least as far as the client knows. This is so terrible though. Mojang why.
the_three-d_biomes_massacre


octylFractal:
@wizjany this needs a matching adapter update for the regen changes at some point, though I don't think it's a huge deal since on Bukkit in 1.16 we always have 3D biomes. Maybe we should elide the check and leave a comment?

PR #1453, opened 2020-07-23
Allow regeneration to accept Extents instead of EditSessions.

dordsor21:
Allows for custom Extents to be parsed to regen, and for clipboard regeneration, etc whilst still allowing EditSessions


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/adapter/BukkitImplAdapter.java
```diff
@@ -227,6 +228,18 @@ default OptionalInt getInternalBlockStateId(BlockState state) {
      * @return true on success, false on failure
      */
     default boolean regenerate(World world, Region region, EditSession session, RegenOptions options) {
```
me4502, line 230: Adapters aren't API, so this isn't needed here anymore.

dordsor21:
It should be noted that the updated WorldEdit-Adapters.jar needs to be placed in the Bukkit module yet

me4502:
Could you please update the included adapters here to be the ones from your adapters PR?

me4502:
Actually nevermind, it's easier to keep it separate just in case another change occurs in the adapters

Issue #1441, opened 2020-07-17
Add a snow brush with a stack mode


me4502:
**Is your feature request related to a problem? Please describe.**
An extension of the snow stacker issue, add a brush version of it

**Describe the solution you'd like**
A brush that places/stacks snow in the same manner as the snow command.

**Describe alternatives you've considered**
Using myself as a mini brush and spamming the snow command everywhere. It's like body paint

**Additional Context**
This probably requires a Brush rewrite :(

PR #1462, opened 2020-07-29
Added a snow brush

me4502:
Quite simple, adds a brush version of the snow simulator system.

After making this I now feel even stronger about the need for a major overhaul of the brush system. This is way too complex.

Also, I'm unsure whether this should be inside the `apply` subcommand or not, the categorisation was rather unclear so that's up in the air.

Closes https://github.com/EngineHub/WorldEdit/issues/1441
Closes https://github.com/EngineHub/WorldEdit/issues/1466

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/factory/ApplyLayer.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/factory/ApplyLayer.java
@@ -21,25 +21,26 @@ package com.sk89q.worldedit.function.factory;
 
 import com.sk89q.worldedit.function.Contextual;
 import com.sk89q.worldedit.function.EditContext;
-import com.sk89q.worldedit.function.RegionFunction;
+import com.sk89q.worldedit.function.LayerFunction;
 import com.sk89q.worldedit.function.operation.Operation;
-import com.sk89q.worldedit.function.visitor.RegionVisitor;
+import com.sk89q.worldedit.function.visitor.LayerVisitor;
+import com.sk89q.worldedit.regions.FlatRegion;
 import com.sk89q.worldedit.regions.NullRegion;
-import com.sk89q.worldedit.regions.Region;
+import com.sk89q.worldedit.regions.Regions;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.sk89q.worldedit.util.GuavaUtil.firstNonNull;
 
-public class Apply implements Contextual<Operation> {
+public class ApplyLayer implements Contextual<Operation> {
 
-    private final Region region;
-    private final Contextual<? extends RegionFunction> function;
+    private final FlatRegion region;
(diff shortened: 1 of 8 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/factory/Snow.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private final boolean stack;
+
+    public Snow(boolean stack) {
+        this.stack = stack;
+    }
+
+    @Override
+    public LayerFunction createFromContext(EditContext context) {
+        return new SnowSimulator(Request.request().getEditSession(), this.stack);
```
wizjany, line 38: iirc this can use `context.getDestination()`, which on one hand I'm not sure will 100% be an EditSession but on the other I'm not sure SnowSimulator requires any more than a raw Extent.

me4502: Yeah it just needs a raw extent, fixed.

me4502:
Also fixes https://github.com/EngineHub/WorldEdit/issues/1466

worldedit-core/src/main/java/com/sk89q/worldedit/function/factory/ApplyLayer.java
```diff
@@ -0,0 +1,65 @@
...
+
+        return new LayerVisitor(localRegion,
+            localRegion.getMinimumPoint().getY(),
+            localRegion.getMaximumPoint().getY(),
+            function.createFromContext(context));
+    }
+
+    @Override
+    public String toString() {
+        return "set " + function;
```
wizjany, line 62: should probably distinguish this from ApplyRegion

me4502: Is that necessary? From a user perspective, it's setting either way - it's just a factory for applying to a LayerVisitor rather than a RegionVisitor

octylFractal: Perhaps we should drive the user facing name via a different method (i18n too maybe?), so that `toString` can be meaningful for us.

worldedit-core/src/main/java/com/sk89q/worldedit/regions/factory/CylinderRegionFactory.java
```diff
@@ -26,12 +26,16 @@
...
 public class CylinderRegionFactory implements RegionFactory {
 
-    private final double height;
+    private double height;
 
     public CylinderRegionFactory(double height) {
         this.height = height;
     }
 
+    public void setHeight(double height) {
```
octylFractal, line 35: Given that this is the only parameter, instead of making this mutable, you could just construct a new one in the brush code?

PR #1476, opened 2020-08-10
Copy the class source validator from WG.

wizjany:
Warns when other plugins incorrectly shade WE or our libs.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/ClassSourceValidator.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/ClassSourceValidator.java
@@ -1,0 +1,131 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.bukkit.util;
+
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableMap;
+import org.bukkit.plugin.Plugin;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.security.CodeSource;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Validates that certain specified classes came from the same source as
+ * a plugin.
+ */
+public class ClassSourceValidator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClassSourceValidator.class);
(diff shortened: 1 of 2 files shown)


worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/ClassSourceValidator.java
```diff
@@ -0,0 +1,128 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Validates that certain specified classes came from the same source as
+ * a plugin.
+ */
+public class ClassSourceValidator {
+
+    private static final Logger log = Logger.getLogger(ClassSourceValidator.class.getCanonicalName());
```
octylFractal, line 42:
```suggestion
    private static final Logger LOGGER = Logger.getLogger(ClassSourceValidator.class);
```

worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/ClassSourceValidator.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final String separatorLine = Strings.repeat("*", 46);
```
octylFractal, line 43:
```suggestion
    private static final String SEPARATOR_LINE = Strings.repeat("*", 46);
```

worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/ClassSourceValidator.java
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @param classes A list of classes to check
+     * @return The results
+     */
+    public Map<Class<?>, CodeSource> findMismatches(List<Class<?>> classes) {
+        checkNotNull(classes, "classes");
+
+        Map<Class<?>, CodeSource> mismatches = Maps.newHashMap();
+
+        if (expectedCodeSource != null) {
```
octylFractal, line 71: Check this before `mismatches` is initialized and return `ImmutableMap.of`?

worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/ClassSourceValidator.java
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @param classes The list of classes to check
+     */
+    public void reportMismatches(List<Class<?>> classes) {
+        if (Boolean.getBoolean("enginehub.disable.class.validation")) {
+            return;
+        }
+        Map<Class<?>, CodeSource> mismatches = findMismatches(classes);
+
+        if (!mismatches.isEmpty()) {
```
octylFractal, line 96: Why not `if (empty) return`? seems like a waste to indent this block of code

worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/ClassSourceValidator.java
```diff
@@ -0,0 +1,128 @@
...
+            StringBuilder builder = new StringBuilder("\n");
+
+            builder.append(separatorLine).append("\n");
+            builder.append("** /!\\    SEVERE WARNING    /!\\\n");
+            builder.append("** \n");
+            builder.append("** A plugin developer has copied and pasted a portion of \n");
```
octylFractal, line 102:
```suggestion
            builder.append("** A plugin developer has included a portion of \n");
```
I don't think we should assume copy/paste, shading is likely too (and while still technically copy/paste, it's not quite the same).

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
```diff
@@ -139,6 +143,12 @@ public void onLoad() {
      */
     @Override
     public void onEnable() {
+
+        // Catch bad things being done by naughty plugins that include
+        // WorldEdit's classes
+        ClassSourceValidator verifier = new ClassSourceValidator(this);
+        verifier.reportMismatches(ImmutableList.of(World.class, CommandManager.class));
```
octylFractal, line 150: Maybe toss in `EditSession` here too?

me4502: Maybe Actor too

Issue #1490, opened 2020-08-14
Left click "jump to" with compass is not working


Fixer-007:
**Versions**

WorldEdit version: WorldEdit 7.2.0 Beta 3 (Fabric for MC 1.16.1)
Platform version: Fabric 0.9.1-build.205

**Describe the bug**
Left click "jump to" not working.

**To Reproduce**
//wand -n to give yourself a compass, try left clicking with it to some distant terrain, it will not teleport you, if you point to smth within player's reach it will teleport you fine. Right click teleport through is working fine. Creative mode, cheats enabled.

Vaelot:
I think this is for fabric-specific bug from commit a0b4f9330, which deletes left button event.
I'm curious about why they delete those event code, but anyway after revert it, it works fine for me. I'll pull the request soon.

booski:
I'm experiencing this bug as well. It worked fine under Forge 35.1.15, but is broken under Fabric 0.6.1.51

Worldedit version: 7.2.1
MC version: 1.16.4

octylFractal:
Duplicate of #1439.

PR #1493, opened 2020-08-16
Fix onLeftClickAir click event missing.

Vaelot:
This pull request is fixing fabric-specific issue #1490.

### DIFF
--- a/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
+++ b/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
@@ -240,6 +240,12 @@ public class FabricWorldEdit implements ModInitializer {
         return ActionResult.PASS;
     }
 
+    public void onLeftClickAir(PlayerEntity playerEntity) {
+        WorldEdit we = WorldEdit.getInstance();
+        FabricPlayer player = adaptPlayer((ServerPlayerEntity) playerEntity);
+        we.handleArmSwing(player);
+    }
+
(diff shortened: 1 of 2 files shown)


wizjany:
iirc the issue with this is that it triggered a lot more than it should, including for some right click actions.

Vaelot:
> iirc the issue with this is that it triggered a lot more than it should, including for some right click actions.

SO, the reason why remove this part is because onLeftClickAir triggers multiple times, which I didn't realized that(I though it might be my hardware issue, especially on my mouse.)

But I'm afraid just removing those lines and left navtool as an uncomfortable feature (for fabric server) is worse than just left that bug.

I think that behavior is not that much as severe as crash(...or there's something missed issues?) That was the reason why I pulled this request.

octylFractal:
Left click air is basically non-functional in 1.16 as-is, I haven't had time to figure out if we can get a reliable method, but firing on both left and right is unacceptable, and considering that Bukkit never has a left click air, I don't think it's much of a loss for now.

Vaelot:
I think this PR is part of a difference of POV.

> Left click air is basically non-functional in 1.16 as-is, 

I didn't realize that was non functional; I had no idea about both how fabric api is working and CoC of worldedit. I just thought that #1490 was annoying me, so I revert those changes and PRed it.

> ...I haven't had time to figure out if we can get a reliable method, but firing on both left and right is unacceptable, and considering that Bukkit never has a left click air, I don't think it's much of a loss for now.

I agree with you. Maybe this PR is not respecting your progresses. I think I should close this PR.

PR #1497, opened 2020-08-19
Use a multipart upload to improve size limits on pastebin uploads

me4502:
This uses the multipart uploader on the paste site, as it has a higher maximum file size limit.

The weird boundary calculation thing is to be "correct" to the format.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/net/HttpRequest.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/net/HttpRequest.java
@@ -19,6 +19,7 @@
 
 package com.sk89q.worldedit.util.net;
 
+import com.sk89q.worldedit.util.concurrency.LazyReference;
 import com.sk89q.worldedit.util.io.Closer;
 
 import java.io.BufferedInputStream;
@@ -38,10 +39,10 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLEncoder;
-import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
+import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
 
 public class HttpRequest implements Closeable {
 
@@ -87,9 +88,21 @@ public class HttpRequest implements Closeable {
      * @param form the form
      * @return this object
      */
-    public HttpRequest bodyForm(Form form) {
+    public HttpRequest bodyUrlEncodedForm(Form form) {
         contentType = "application/x-www-form-urlencoded";
-        body = form.toString().getBytes();
+        body = form.toUrlEncodedString().getBytes();
+        return this;
+    }
+
+    /**
+     * Submit form data.
+     *
+     * @param form the form
+     * @return this object
+     */
+    public HttpRequest bodyMultipartForm(Form form) {
+        contentType = "multipart/form-data;boundary=" + form.getFormDataSeparator();
+        body = form.toFormDataString().getBytes();
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/util/net/HttpRequest.java
```diff
@@ -87,9 +88,21 @@ public HttpRequest body(String data) {
      * @param form the form
      * @return this object
      */
-    public HttpRequest bodyForm(Form form) {
+    public HttpRequest bodyUrlEncodedForm(Form form) {
         contentType = "application/x-www-form-urlencoded";
-        body = form.toString().getBytes();
+        body = form.toUrlEncodedString().getBytes();
```
octylFractal, line 93:
```suggestion
        body = form.toUrlEncodedString().getBytes(StandardCharsets.UTF_8);
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/net/HttpRequest.java
```diff
@@ -87,9 +88,21 @@ public HttpRequest body(String data) {
...
+
+    /**
+     * Submit form data.
+     *
+     * @param form the form
+     * @return this object
+     */
+    public HttpRequest bodyMultipartForm(Form form) {
+        contentType = "multipart/form-data;boundary=" + form.getFormDataSeparator();
+        body = form.toFormDataString().getBytes();
```
octylFractal, line 105:
```suggestion
        body = form.toFormDataString().getBytes(StandardCharsets.UTF_8);
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/net/HttpRequest.java
```diff
@@ -363,18 +376,20 @@ private static URL reformat(URL existing) {
...
 
     /**
-     * Used with {@link #bodyForm(Form)}.
+     * Used with {@link #bodyUrlEncodedForm(Form)}.
      */
     public static final class Form {
-        public final List<String> elements = new ArrayList<>();
+        public final Map<String, String> elements = new LinkedHashMap<>();
+
+        private LazyReference<String> formDataSeparator = LazyReference.from(
```
octylFractal, line 390: I don't get why we're using `LazyReference` here? Just compute it on initialization, it's not expensive.

worldedit-core/src/main/java/com/sk89q/worldedit/util/net/HttpRequest.java
```diff
@@ -387,26 +402,51 @@ private Form() {
...
+        public String getFormDataSeparator() {
+            return formDataSeparator.getValue();
+        }
+
+        public String toFormDataString() {
+            String formSeparator = "--" + formDataSeparator.getValue();
+            StringBuilder builder = new StringBuilder();
+
+            for (Map.Entry<String, String> element : elements.entrySet()) {
+                builder
```
octylFractal, line 418: This might benefit from using some form of Joiner.

worldedit-core/src/main/java/com/sk89q/worldedit/util/net/HttpRequest.java
```diff
@@ -387,26 +402,51 @@ private Form() {
...
+            return builder.toString();
         }
 
-        @Override
-        public String toString() {
+        public String toUrlEncodedString() {
             StringBuilder builder = new StringBuilder();
             boolean first = true;
-            for (String element : elements) {
+            for (Map.Entry<String, String> element : elements.entrySet()) {
```
octylFractal, line 436: This might also use Joiner, esp. MapJoiner.

worldedit-core/src/main/java/com/sk89q/worldedit/util/paste/EngineHubPaste.java
```diff
@@ -50,31 +46,35 @@ private PasteTask(String content) {
...
+                    .bodyMultipartForm(form)
+                    .execute()
+                    .expectResponseCode(200, 204);
 
-            throw new IOException("Failed to save paste; instead, got: " + result);
+            return new URL(response.viewUrl);
         }
     }
 
+    @SuppressWarnings("MismatchedQueryAndUpdateOfCollection")
```
octylFractal, line 74: I don't think we should add non-javac warnings to suppression.

worldedit-core/src/main/java/com/sk89q/worldedit/util/net/HttpRequest.java
me4502: This would be fairly messy to do, due to the need of URLEncoding the values/keys, and the fact that it has a checked exception on it

worldedit-core/src/main/java/com/sk89q/worldedit/util/net/HttpRequest.java
me4502: I feel this format is too complex for a joiner

PR #1503, opened 2020-08-24
Fixed non-Exception throwables being ignored in async tasks, and added a timed message

### LINKED ISSUE #1502: AsyncCommandBuilder drops Throwable/Errors
`AsyncCommandBuilder` only catches `Exception`, so any `Error` slips on by the error message system. This should not occur.

me4502:
Fixes #1502

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
@@ -133,8 +133,8 @@ public class SchematicCommands {
         SchematicLoadTask task = new SchematicLoadTask(actor, f, format);
         AsyncCommandBuilder.wrap(task, actor)
                 .registerWithSupervisor(worldEdit.getSupervisor(), "Loading schematic " + filename)
-                .sendMessageAfterDelay(TranslatableComponent.of("worldedit.schematic.load.loading"))
-                .sendRepeatedMessage(TranslatableComponent.of("worldedit.schematic.load.still-loading"))
+                .setDelayMessage(TranslatableComponent.of("worldedit.schematic.load.loading"))
+                .setWorkingMessage(TranslatableComponent.of("worldedit.schematic.load.still-loading"))
                 .onSuccess(TextComponent.of(filename, TextColor.GOLD)
                                 .append(TextComponent.of(" loaded. Paste it with ", TextColor.LIGHT_PURPLE))
                                 .append(CodeFormat.wrap("//paste").clickEvent(ClickEvent.of(ClickEvent.Action.SUGGEST_COMMAND, "//paste"))),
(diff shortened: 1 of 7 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/util/AsyncCommandBuilder.java
```diff
@@ -125,7 +133,12 @@ private AsyncCommandBuilder(Callable<T> callable, Actor sender) {
     public ListenableFuture<T> buildAndExec(ListeningExecutorService executor) {
         final ListenableFuture<T> future = checkNotNull(executor).submit(this::runTask);
         if (delayMessage != null) {
```
wizjany, line 135: does this handle the case where delayMessage == null && repeatedMessage != null? what's intended behavior there? possibly it should use the repeatedMessage at 1 second and then repeat it every 10 as well?

me4502: It currently doesn't, but IMO the delayMessage should be required to use the repeatedMessage. So I'll just make the builder require delayMessage when repeatedMessage is passed.

worldedit-core/src/main/java/com/sk89q/worldedit/command/util/MessageTimerTask.java
```diff
@@ -24,30 +24,41 @@
...
 
     @Override
     public void run() {
-        sender.printDebug(message);
+        if (!hasRunBefore) {
+            sender.printDebug(message);
+            hasRunBefore = true;
+        } else if (repeatedMessage != null) {
+            sender.printDebug(repeatedMessage);
+        }
```
octylFractal, line 61:
```suggestion
        } else {
            cancel();
        }
```
To save some CPU?

worldedit-core/src/main/java/com/sk89q/worldedit/command/util/AsyncCommandBuilder.java
```diff
@@ -94,6 +96,12 @@ private AsyncCommandBuilder(Callable<T> callable, Actor sender) {
         return this;
     }
 
+    public AsyncCommandBuilder<T> sendDelayedRepeatingMessage(Component message) {
```
octylFractal, line 99: Shouldn't this be more like `sendMessageAfterFurtherDelay` (maybe `Delays`)? Currently this is in the present tense.

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -134,6 +134,7 @@ public void load(Actor actor, LocalSession session,
         AsyncCommandBuilder.wrap(task, actor)
                 .registerWithSupervisor(worldEdit.getSupervisor(), "Loading schematic " + filename)
                 .sendMessageAfterDelay(TranslatableComponent.of("worldedit.schematic.load.loading"))
+                .sendRepeatedMessage(TranslatableComponent.of("worldedit.schematic.load.still-loading"))
```
octylFractal, line 137: These didn't get changed, compile errors abound.

PR #1507, opened 2020-08-30
Remove data versions from the Bukkit adapters

me4502:
Adapters PR: https://github.com/EngineHub/worldedit-adapters/pull/20

There's API to handle this that exists from 1.13.2 onwards, so no need to have it in the adapters.


wizjany:
One notable downside of this is that schematics will be saved with a valid DV even if the data in them lacks properties due to missing adapters. This causes DFU issues later when loading them in an updated WE.

me4502:
I somewhat feel having a valid DV is better than having an invalid one. Technically the blocks are all still valid, they just refer to the default state at that DV

octylFractal (changes requested):
Personally I would like for this to change it so that schematics _cannot_ be saved if no adapter exists, at least not without a configuration option change to confirm they know what they are doing. It's way too common for users to simply not know that the copy does not have all the data it needs.

me4502:
This PR now prevents schematics from being saved when the dataversion is -1, and sets the dataversion to -1 when an adapter is not found.

This is done at the command level, to prevent issues from occurring on the API side.

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -108,6 +108,7 @@
     "worldedit.schematic.delete.failed": "Deletion of {0} failed! Is it read-only?",
     "worldedit.schematic.delete.deleted": "{0} has been deleted.",
     "worldedit.schematic.formats.title": "Available clipboard formats (Name: Lookup names)",
+    "worldedit.schematic.unsupported-minecraft-version": "This version of WorldEdit does not support your Minecraft version. Schematics will not work until this is resolved. Update WorldEdit to restore this functionality.",
```
octylFractal, line 111: Unsure if `Update WorldEdit to restore this functionality.` is the correct wording here, in a potential future where we cut out e.g. 1.13.2 support, _updating_ will do nothing to fix this issue. A bit niche, but I'd like to have consistent messaging. Perhaps just dropping the last line is sufficient?

octylFractal (changes requested):

octylFractal (approved):

PR #1508, opened 2020-08-30
Workaround for snapshot files with not-to-spec slashes.

wizjany:
Some utilities will create zip entries using windows path separators, which despite being against the zip file specification, we should probably read in anyway.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/storage/ZippedMcRegionChunkStore.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/storage/ZippedMcRegionChunkStore.java
@@ -81,12 +81,18 @@ public class ZippedMcRegionChunkStore extends McRegionChunkStore {
             }
         } else {
             Pattern pattern = Pattern.compile(".*\\.mc[ra]$");
+            Pattern worldPattern = Pattern.compile(worldName + "[\\\\/].*");
             for (Enumeration<? extends ZipEntry> e = zip.entries(); e.hasMoreElements(); ) {
                 ZipEntry testEntry = e.nextElement();
                 // Check for world
-                if (testEntry.getName().startsWith(worldName + "/")) {
-                    if (pattern.matcher(testEntry.getName()).matches()) { // does entry end in .mca
-                        folder = testEntry.getName().substring(0, testEntry.getName().lastIndexOf('/'));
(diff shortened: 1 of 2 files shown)


me4502 (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/world/storage/ZippedMcRegionChunkStore.java
```diff
@@ -84,9 +84,14 @@ protected InputStream getInputStream(String name, String worldName) throws IOExc
             for (Enumeration<? extends ZipEntry> e = zip.entries(); e.hasMoreElements(); ) {
                 ZipEntry testEntry = e.nextElement();
                 // Check for world
-                if (testEntry.getName().startsWith(worldName + "/")) {
-                    if (pattern.matcher(testEntry.getName()).matches()) { // does entry end in .mca
-                        folder = testEntry.getName().substring(0, testEntry.getName().lastIndexOf('/'));
+                String entryName = testEntry.getName();
+                if (entryName.startsWith(worldName + "/")) {
```
wizjany, line 88: on second look this line is still broken, it doesn't use the matcher like the TrueZip version does.

worldedit-core/src/main/java/com/sk89q/worldedit/world/storage/ZippedMcRegionChunkStore.java
```diff
@@ -81,12 +81,18 @@ protected InputStream getInputStream(String name, String worldName) throws IOExc
...
                 // Check for world
-                if (testEntry.getName().startsWith(worldName + "/")) {
-                    if (pattern.matcher(testEntry.getName()).matches()) { // does entry end in .mca
-                        folder = testEntry.getName().substring(0, testEntry.getName().lastIndexOf('/'));
+                String entryName = testEntry.getName();
+                if (worldPattern.matcher(entryName).matches()) {
+                    if (pattern.matcher(entryName).matches()) { // does entry end in .mca
+                        int endIndex = entryName.lastIndexOf('/');
+                        if (endIndex < 0) {
+                            endIndex = entryName.indexOf('\\');
```
octylFractal, line 93: Shouldn't this be `lastIndexOf`?

octylFractal (changes requested):

PR #1513, opened 2020-09-02
Remove Y limits if a region has no world

### LINKED ISSUE #1495: Schematic gets cropped
**Versions**

WorldEdit version: 1.16.1
Platform version: Java

**Describe the bug**
 A mapart schematic I had would get cropped at the top after 256 Y, even when I rotated it 90 degrees so it would lay flat on the ground.

**To Reproduce**

1. ... Have a schematic of a MapArt (over 256x256)
2. ... Rotate 90 degrees (//rotate 0 90)
3... //paste

**Expected behavior**
The blocks above 256 what wouldve been Y will be cut off

**Screenshots**

**Additional context**

octylFractal:
This allows schematics to load in without clipping, so that they may be properly rotated later.

Edit: fixes #1495.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/AbstractRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/AbstractRegion.java
@@ -207,11 +207,11 @@ public abstract class AbstractRegion implements Region {
     // Sub-class utilities
 
     protected final int getWorldMinY() {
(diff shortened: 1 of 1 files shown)


me4502 (approved):

wizjany:
any reason to not change min as well?

octylFractal:
it is done

wizjany (approved):

PR #1514, opened 2020-09-03
Create an asset system, and add an image brush that uses it

me4502:
Continue's @KennyTV work, and creates a more generic asset system (for potential future asset types, etc)

Will continue more with a proper writeup tomorrow

Supersedes https://github.com/EngineHub/WorldEdit/pull/1295


worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/AssetLoaders.java
```diff
@@ -0,0 +1,150 @@
...
+     * @param assetClass The asset class
+     * @return The list of files
+     */
+    public List<Path> getFilesForAsset(Class<?> assetClass) {
+        Set<String> extensions = this.assetLoaderMap.getOrDefault(assetClass, ImmutableMap.of()).keySet();
+
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(this.assetsDir, entry -> extensions.contains(MoreFiles.getFileExtension(entry)))) {
+            return ImmutableList.copyOf(stream);
+        } catch (IOException e) {
+            e.printStackTrace();
```
octylFractal, line 136: Use a logger, not STDERR. Also consider moving the `return` inside the `catch`?

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/ImageHeightmapLoader.java
```diff
@@ -0,0 +1,53 @@
...
+    public ImageHeightmap loadAssetFromPath(Path path) throws Exception {
+        if (!Files.exists(path)) {
+            return null;
+        }
+        return new ImageHeightmap(ImageIO.read(path.toFile()));
+    }
+
+    @Override
+    public Set<String> getAllowedExtensions() {
+        return ImmutableSet.of("png", "jpg", "jpeg");
```
octylFractal, line 51: Since you're already using `ImageIO`, perhaps use `ImageIO.getReaderFileSuffixes()`?

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/holder/ImageHeightmap.java
```diff
@@ -0,0 +1,67 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.util.asset.holder;
+
+import java.awt.Graphics2D;
+import java.awt.image.BufferedImage;
+
+public class ImageHeightmap {
```
octylFractal, line 25: Javadocs for this class?

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/holder/ImageHeightmap.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private BufferedImage resizedImage;
+    private int lastSize = -1;
+
+    public ImageHeightmap(BufferedImage image) {
+        this.image = image;
+    }
+
+    public double getHeightAt(int x, int y, int size, double intensity) {
+        if (size != lastSize || resizedImage == null) {
```
octylFractal, line 37: I'm not sure this resize is worth it. Can't you just compute this when accessing the RGB data? e.g.:
```java
double ratioX = image.getWidth() / (double) size;
double ratioY = image.getHeight() / (double) size;

int rgb = image.getRGB((int) (ratioX * x), (int) (ratioY * y));
```

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/holder/ImageHeightmap.java
me4502: That was from KennyTV's code, my assumption there was that resizing it allows for better gradients compared to just directly accessing. Realistically that's probably not a concern though as the image is likely much larger than the brush itself

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -465,4 +533,20 @@ static void setOperationBasedBrush(Player player, LocalSession session, double r
 
         player.printInfo(TranslatableComponent.of("worldedit.brush.operation.equip", TextComponent.of(factory.toString())));
     }
+
+    private static class ImageHeightmapLoadTask implements Callable<ImageHeightmap> {
```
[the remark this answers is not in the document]
me4502: Why?

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/AssetLoaders.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Gets the Asset Loader of the given type.
+     *
+     * @param assetClass The class to get a loader for
+     * @param filename The filename to attempt to load
+     *
+     * @param <T> The returned asset type
+     */
+    @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
me4502: Due to multiple unchecked casts here I feel it'd be rather messy to do so

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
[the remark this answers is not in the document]
octylFractal: Re-use for other `AssetLoader` generics in the future, e.g. if we start loading object files we want those to be using the same load task class.

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -287,7 +287,6 @@ public void list(Actor actor,
             pathComparator = Comparator.naturalOrder();
             flag = "";
         }
-        final int sortType = oldFirst ? -1 : newFirst ? 1 : 0;
```
octylFractal, line 290: Can this be removed and just done in `master`?

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/AssetLoader.java
```diff
@@ -0,0 +1,125 @@
...
+     * Loads an asset from the given file if possible.
+     *
+     * @param path The file to load
+     * @return loaded asset, or null otherwise
+     */
+    @Nullable
+    protected abstract T loadAssetFromPath(Path path) throws Exception;
+
+    /**
+     * Returns an immutable set of asset keys.
```
octylFractal, line 111: Perhaps describe the caching in some form? Or don't expose this at all (seems like an implementation detail of the loader).

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/holder/ImageHeightmap.java
```diff
@@ -0,0 +1,81 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.util.asset.holder;
+
+import java.awt.Graphics2D;
+import java.awt.image.BufferedImage;
+
+/**
+ * Represents an image that acts as a heightmap.
```
octylFractal, line 26: I would like it if it described _how_ it interprets an image as a heightmap.

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -325,6 +337,62 @@ public void butcherBrush(Player player, LocalSession session,
...
+                TextComponent.Builder builder = TextComponent.builder();
+                int i = 0;
+                for (Path path : imageNames) {
+                    builder.append(TextComponent.of(MoreFiles.getNameWithoutExtension(path), i % 2 == 0 ? TextColor.GRAY : TextColor.WHITE));
+                    if (i <= imageNames.size()) {
+                        builder.append(TextComponent.of(", "));
+                    }
+                    i++;
+                }
+                return TranslatableComponent.of("worldedit.brush.image.unknown", TextComponent.of(imageName), builder.build());
```
octylFractal, line 387: Shouldn't this be a paginated box of some sort? I don't really like the idea of dumping a potentially massive list into chat.

octylFractal (changes requested):
Also, still not marked with `@Beta`.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/AssetLoader.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Loads an asset.
+     *
+     * @param path path in assets directory, can be with and without its file extension
+     * @return asset if successfully loaded, null otherwise
+     */
+    @Nullable
+    public T getAsset(String path) {
+        // Remove file extension from mapped name
```
octylFractal, line 58: So if I have `foo.jpg` and `foo.png`, they map to the same thing? Why isn't the full path used for caching?

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/AssetLoaders.java
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * @param <T> The asset type
+     */
+    @SuppressWarnings("unchecked")
+    public <T> List<AssetLoader<T>> getAssetLoaders(Class<T> assetClass) {
+        if (!assetLoaderRegistration.containsRow(assetClass)) {
+            return ImmutableList.of();
+        }
+
+        return ImmutableList.copyOf((Collection<AssetLoader<T>>) (Collection<?>) assetLoaderRegistration.values());
```
octylFractal, line 140: Pretty sure this should be keyed by `assetClass` somehow?

worldedit-core/src/main/java/com/sk89q/worldedit/util/asset/holder/ImageHeightmap.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * <p>
+ *     Height is determined by how light each pixel of the image is,
+ *     from black (0) to white (1).
+ *
+ *     Lightness is determined by an average of the 3 color channels.
+ * </p>
```
octylFractal, line 35:
```suggestion
 * <p>
 * Height is determined by how light each pixel of the image is,
 * from black (0) to white (1). Lightness is determined by an
 * average of the 3 color channels.
 * </p>
```

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -325,6 +332,46 @@ public void butcherBrush(Player player, LocalSession session,
         player.printInfo(TranslatableComponent.of("worldedit.brush.butcher.equip", TextComponent.of((int) radius)));
     }
 
+    @Command(
+        name = "image",
+        desc = "Image brush, uses images as input"
```
octylFractal, line 337: This does not just "use images as input" as a user would expect, it creates a heightmap from it. This should give some sort of overview of how the brush actually functions, at least in a very general sense.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -325,6 +332,46 @@ public void butcherBrush(Player player, LocalSession session,
         player.printInfo(TranslatableComponent.of("worldedit.brush.butcher.equip", TextComponent.of((int) radius)));
     }
 
+    @Command(
+        name = "image",
```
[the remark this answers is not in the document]
octylFractal: @wizjany help us bikeshed a better name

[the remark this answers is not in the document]
me4502: 🚲🏡

octylFractal (approved):

PR #1525, opened 2020-09-10
Fixed //snow going through slabs

### LINKED ISSUE #1516: //snow goes through slabs
WorldEdit version: Forge-Official(7.2.0-SNAPSHOT+5332-0cb1eea)
Platform version: Forge 33.0.37 1.16.2

**Describe the bug**
Slabs of all types/states are treated as air in the snow command.

**To Reproduce**
1. place some slabs above an area
2. //snow

**Screenshots**

NoteInBlock:
This fixes the bug that snow was placed under slabs when using //snow command.

It also places snow on blocks now when the block is a double slab or if the slab is on the top side of its block.

fixes #1516

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
@@ -24,6 +24,8 @@ import com.sk89q.worldedit.extent.Extent;
 import com.sk89q.worldedit.function.LayerFunction;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.registry.state.Property;
+import com.sk89q.worldedit.world.block.BlockCategories;
+import com.sk89q.worldedit.world.block.BlockCategory;
 import com.sk89q.worldedit.world.block.BlockState;
 import com.sk89q.worldedit.world.block.BlockTypes;
 
@@ -35,6 +37,7 @@ public class SnowSimulator implements LayerFunction {
 
     private final Property<Integer> snowLayersProperty = BlockTypes.SNOW.getProperty("layers");
     private final Property<Integer> waterLevelProperty = BlockTypes.WATER.getProperty("level");
+    private final Property<String> slabTypeProperty = BlockTypes.OAK_SLAB.getProperty("type");
 
     private final Extent extent;
     private final boolean stack;
@@ -66,6 +69,13 @@ public class SnowSimulator implements LayerFunction {
             return true;
         }
 
+        // Slabs are also valid but not "FullCube"
+        BlockCategory blockCategorySlabs = new BlockCategory(BlockCategories.SLABS.getId());
+
+        if (blockCategorySlabs.contains(block.getBlockType())) {
+            return true;
+        }
+
         // Can only place on full solid blocks
         return block.getBlockType().getMaterial().isFullCube()
                 && block.getBlockType().getMaterial().isSolid();
@@ -86,6 +96,13 @@ public class SnowSimulator implements LayerFunction {
                     affected++;
                 }
             }
+            return false;
+        }
+
+        // Prevent placing snow on slabs if on bottom
+        BlockCategory blockCategorySlabs = new BlockCategory(BlockCategories.SLABS.getId());
+        if (blockCategorySlabs.contains(block.getBlockType())
+                && block.getState(slabTypeProperty).equals("bottom")) {
             return false;
         }
 


worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
```diff
@@ -35,6 +37,7 @@
 
     private final Property<Integer> snowLayersProperty = BlockTypes.SNOW.getProperty("layers");
     private final Property<Integer> waterLevelProperty = BlockTypes.WATER.getProperty("level");
+    private final Property<String> slabTypeProperty = BlockTypes.OAK_SLAB.getProperty("type");
```
wizjany, line 40: has this been tested with slabs other than oak? the same property of two different blocks is likely a different property instance. I'm not sure that getting e.g. stone slabs' "type" property with an instance of oak slabs' property is "reliable" behavior, even if it happens to work.

worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
```diff
@@ -66,6 +69,13 @@ public boolean isGround(BlockVector3 position) {
             return true;
         }
 
+        // Slabs are also valid but not "FullCube"
+        BlockCategory blockCategorySlabs = new BlockCategory(BlockCategories.SLABS.getId());
```
wizjany, line 73: shouldn't be constructing a new instance. `BlockCategories` exists.

worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
```diff
@@ -89,6 +99,13 @@ public boolean apply(BlockVector3 position, int depth) throws WorldEditException
             return false;
         }
 
+        // Prevent placing snow on slabs if on bottom
+        BlockCategory blockCategorySlabs = new BlockCategory(BlockCategories.SLABS.getId());
```
wizjany, line 103: as above, don't construct new categories.

wizjany (changes requested):

wizjany:
in general, is there another rule MC uses for this? is it a hard-coded check for slabs or is there something else we can check?

octylFractal:
I feel like the proper solution here is to simply return `true` in the last part of `isGround`, and bail out of `apply` if the original condition (full cube & solid) isn't met. That might make it not go through things such as torches, but I think that's reasonable.

MC itself uses `blockState.getMaterial().blocksMovement() || !blockState.getFluidState().isEmpty()` as the condition for "ground", saved in a chunk heightmap; and the condition for placement is in `SnowBlock.canPlaceAt`:
```java
      BlockState blockState = world.getBlockState(pos.down());
      if (!blockState.isOf(Blocks.ICE) && !blockState.isOf(Blocks.PACKED_ICE) && !blockState.isOf(Blocks.BARRIER)) {
         if (!blockState.isOf(Blocks.HONEY_BLOCK) && !blockState.isOf(Blocks.SOUL_SAND)) {
            return Block.isFaceFullSquare(blockState.getCollisionShape(world, pos.down()), Direction.UP) || blockState.getBlock() == this && (Integer)blockState.get(LAYERS) == 8;
         } else {
            return true;
         }
      } else {
         return false;
      }
```
Which looks to be the complement of [ice, packed_ice, barrier] + [honey_block, soul_sand] + [$B where isFaceFullSquare($B)] (this is basically isFullCube) + [snow[layers=8]]. Pretty complex.

worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
NoteInBlock: I tested this with several different slabs and it worked. Unfortunately I couldn't find something like "BlockCategories.SLABS.getProperty()".

worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
NoteInBlock: Yep, sorry for that, my bad.

NoteInBlock:
> I feel like the proper solution here is to simply return `true` in the last part of `isGround`, and bail out of `apply` if the original condition (full cube & solid) isn't met. That might make it not go through things such as torches, but I think that's reasonable.

That will solve the bug #1516 but won't place snow on top of two slabs that have been placed on top of each other and thus forming one full block. Also there won't be snow on top of a slab that has been placed on the upper half of a full block. Is that the intended behavior?

me4502:
I feel a "good enough" stop-gap until we're actually hooking into the internal canPlace systems in Minecraft would be to just do what octyl suggested. Not placing on top of top slabs isn't too big of a deal in comparison to it falling through top slabs in general.

me4502:
Superseded by https://github.com/EngineHub/WorldEdit/pull/1541

Thanks for making this, but it made more sense to use the internal canPlace checks to fix all cases rather than just slabs

PR #1526, opened 2020-09-10
Added an OffsetsMask, with a parser for an Exposed/Surface mask

me4502:
Adds an AdjacentMask, a mask that checks whether any touching blocks match a given mask. Also includes an option to not allow itself to match the mask as well.

Due to the limitations of our mask parser, the full power of this mask isn't exposed yet. I have however, created a `#exposed` or `#surface` mask that matches all blocks that are exposed to air (excluding air itself). This is a somewhat common feature request/question we get in the Discord.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask.java
@@ -20,11 +20,11 @@
 package com.sk89q.worldedit.function.mask;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Lists;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.util.Direction;
 
-import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -35,12 +35,11 @@ import static com.google.common.base.Preconditions.checkNotNull;
  */
 public class OffsetsMask extends AbstractMask {
 
-    private static final ImmutableList<BlockVector3> OFFSET_LIST = ImmutableList.copyOf(
+    private static final ImmutableSet<BlockVector3> OFFSET_LIST =
         Direction.valuesOf(Direction.Flag.CARDINAL | Direction.Flag.UPRIGHT)
             .stream()
             .map(Direction::toBlockVector)
-            .collect(Collectors.toList())
-    );
+            .collect(ImmutableSet.toImmutableSet());
 
     /**
      * Create an offsets mask for a single offset.
@@ -70,7 +69,7 @@ public class OffsetsMask extends AbstractMask {
         private boolean excludeSelf;
         private int minMatches = 1;
         private int maxMatches = Integer.MAX_VALUE;
-        private ImmutableList<BlockVector3> offsets = OFFSET_LIST;
+        private ImmutableSet<BlockVector3> offsets = OFFSET_LIST;
 
         private Builder() {
         }
@@ -130,7 +129,7 @@ public class OffsetsMask extends AbstractMask {
          * @return this builder, for chaining
          */
         public Builder offsets(Iterable<BlockVector3> offsets) {
-            this.offsets = ImmutableList.copyOf(offsets);
+            this.offsets = ImmutableSet.copyOf(offsets);
             return this;
         }
 
@@ -148,9 +147,9 @@ public class OffsetsMask extends AbstractMask {
     private final boolean excludeSelf;
     private final int minMatches;
     private final int maxMatches;
-    private final ImmutableList<BlockVector3> offsets;
+    private final ImmutableSet<BlockVector3> offsets;
 
-    private OffsetsMask(Mask mask, boolean excludeSelf, int minMatches, int maxMatches, ImmutableList<BlockVector3> offsets) {
+    private OffsetsMask(Mask mask, boolean excludeSelf, int minMatches, int maxMatches, ImmutableSet<BlockVector3> offsets) {
         checkNotNull(mask);
         checkNotNull(offsets);
         // Validate match args. No need to test maxMatches as it must be >=0 based on the conditions here.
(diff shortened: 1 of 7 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask2D.java
```diff
@@ -0,0 +1,180 @@
...
+     * @param excludeSelf excludes blocks where the mask matches itself
+     * @param minMatches the minimum number of matches
+     * @param maxMatches the maximum number of matches
+     * @param offsets the block offsets to test with
+     */
+    public OffsetsMask2D(Mask2D mask, boolean excludeSelf, int minMatches, int maxMatches, List<BlockVector2> offsets) {
+        // Validate match args. No need to test maxMatches as it must be >=0 based on the conditions here.
+        checkArgument(minMatches <= maxMatches, "minMatches must be less than or equal to maxMatches");
+        checkArgument(minMatches >= 0, "minMatches must be greater than or equal to 0");
+        checkArgument(minMatches <= offsets.size(), "minMatches must be less than or equal to the number of offsets");
```
octylFractal, line 96: Validate `maxMatches <= offsets.size()` too?

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask2D.java
```diff
@@ -0,0 +1,180 @@
...
+        checkNotNull(mask);
+        checkNotNull(offsets);
```
octylFractal, line 98: This is redundant by this point, should go at the top of the checks if you want to keep it.

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask2D.java
```diff
@@ -0,0 +1,180 @@
...
+    public Mask2D getMask() {
+        return mask;
+    }
+
+    /**
+     * Set the mask.
+     *
+     * @param mask the mask
+     */
+    @Deprecated
```
octylFractal, line 122: Why is this deprecated?

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask2D.java
```diff
@@ -0,0 +1,180 @@
...
+    public List<BlockVector2> getOffsets() {
+        return ImmutableList.copyOf(this.offsets);
+    }
+
+    /**
+     * Set the offsets.
+     *
+     * @param offsets the offsets
+     */
+    @Deprecated
```
octylFractal, line 143: Why is this deprecated?

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask2D.java
```diff
@@ -0,0 +1,180 @@
...
+            return false;
+        }
+
+        int matches = 0;
+
+        for (BlockVector2 offset : offsets) {
+            if (mask.test(vector.add(offset))) {
+                matches++;
+            }
+            if (matches > maxMatches) {
```
octylFractal, line 162: Nest this inside the previous `if`?

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask2D.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * Create an OffsetsMask2D for a single offset.
+     *
+     * @param mask The mask to use
+     * @param offset The offset
+     * @return The offsets mask
+     */
+    public static OffsetsMask2D single(Mask2D mask, BlockVector2 offset) {
```
octylFractal, line 177: Move to the top.

octylFractal (changes requested):
Changes on `OffsetsMask2D` should also be applied to `OffsetsMask`.

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask.java
```diff
@@ -0,0 +1,172 @@
...
+    public Mask getMask() {
+        return mask;
+    }
+
+    /**
+     * Get the offsets.
+     *
+     * @return the offsets
+     */
+    public List<BlockVector3> getOffsets() {
```
octylFractal, line 132:
```suggestion
    public ImmutableList<BlockVector3> getOffsets() {
```
It's recommended to do this if the API declares the list immutable. I would also recommend doing the copy on creation and storing it as the `offsets` field.

worldedit-core/src/main/java/com/sk89q/worldedit/world/RegenOptions.java
```diff
@@ -20,14 +20,14 @@
...
-import com.sk89q.worldedit.EditSession;
+import com.sk89q.worldedit.extent.Extent;
 import com.sk89q.worldedit.regions.Region;
 
 import java.util.OptionalLong;
 import javax.annotation.Nullable;
 
 /**
- * Regeneration options for {@link World#regenerate(Region, EditSession, RegenOptions)}.
+ * Regeneration options for {@link World#regenerate(Region, Extent, RegenOptions)}.
```
octylFractal, line 30: Can this be pulled out of this PR into a commit on master?

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetsMask.java
```diff
@@ -0,0 +1,199 @@
...
+     * Get the offsets.
+     *
+     * @return the offsets
+     */
+    public ImmutableList<BlockVector3> getOffsets() {
+        return this.offsets;
+    }
+
+    /**
+     * Whether matching the current block should fail the mask.
```
octylFractal, line 137: Sentence fragment :(
```suggestion
     * Get the flag determining if matching the current block should fail the mask.
```
or something like that.

octylFractal (changes requested):
Can you also document if min/max matches are inclusive or not, the default set of offsets, and the range of min/max matches? I know there are validation checks, but it's good to have this all in Javadoc too.

octylFractal (approved):

PR #1536, opened 2020-09-21
Use block change limit for //copy and //cut

### LINKED ISSUE #979: Block limits should apply to //copy
WORLDEDIT-3232 - Reported by Zane.Woodard

I had to write a plugin that wrapped the command to achieve this functionality in WorldEdit 5.x. Unfortunately, it broke for WorldEdit 6.x, and I'm looking to patch it but thought I'd try getting an official patch for this.

Basically, either //copy should get its own limit in the config, or it should be restricted to the same max block change limits as every other command. Currently, no limits apply to copy. This makes it a very unsafe command to give to regular players, as it can be easily abused to lag out the server or cause excessive memory usage.

If the entire clipboard system had to be rewritten to make it obey limits, I would understand why //copy had no limits. However, it's a simple patch to add a size check to the region at the top of the copy function, so I don't think this is too much to ask for.

me4502:
Fixes #979

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
@@ -42,6 +42,7 @@ import com.sk89q.worldedit.internal.annotation.Selection;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.math.Vector3;
 import com.sk89q.worldedit.math.transform.AffineTransform;
+import com.sk89q.worldedit.regions.CuboidRegion;
 import com.sk89q.worldedit.regions.Region;
 import com.sk89q.worldedit.regions.RegionSelector;
 import com.sk89q.worldedit.regions.selector.CuboidRegionSelector;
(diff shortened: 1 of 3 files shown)


wizjany:
This is a super hacky and inaccurate workaround :\
On one hand, we can't exactly add checked exceptions to InputExtent now, but I don't know if this is the correct solution honestly.

me4502:
I do agree, I just feel not allowing people to allocate giant chunks of memory when limits are in place may be worthwhile adding a hacky solution for

worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
```diff
@@ -66,6 +68,22 @@
...
 
+    /**
+     * Throws if the region would allocate a clipboard larger than the block change limit.
+     *
+     * @param region The region to check
+     * @param session The session
+     * @throws MaxChangedBlocksException if the volume exceeds the limit
+     */
+    private void checkRegionBounds(Region region, LocalSession session) throws MaxChangedBlocksException {
+        Region clipboardBounds = region instanceof CuboidRegion
```
octylFractal, line 79: Can you make this a method on `Region`:
```java
default CuboidRegion getBoundingBox() {
    return new CuboidRegion(getMinimumPoint(), getMaximumPoint());
}
```
and just override in `CuboidRegion`? I feel like this is generically useful.

worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
```diff
@@ -66,6 +68,22 @@
...
+            ? region
+            : new CuboidRegion(region.getMinimumPoint(), region.getMaximumPoint());
+        if (clipboardBounds.getVolume() > session.getBlockChangeLimit()) {
+            throw new MaxChangedBlocksException(session.getBlockChangeLimit());
```
octylFractal, line 83:
```suggestion
        int limit = session.getBlockChangeLimit();
        if (clipboardBounds.getVolume() > limit) {
            throw new MaxChangedBlocksException(limit);
```

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
```diff
@@ -49,6 +49,15 @@
      */
     BlockVector3 getMaximumPoint();
 
+    /**
+     * Get the bounding box of this region as a CuboidRegion.
```
octylFractal, line 53:
```suggestion
     * Get the bounding box of this region as a {@link CuboidRegion}.
```

octylFractal (changes requested):

octylFractal (approved):

PR #1539, opened 2020-09-26
Allow stacking by block units with //stack

### LINKED ISSUE #1443: Stack via block units, not selection units
**Is your feature request related to a problem? Please describe.**
Stack only operates on selection-sized units, not block units, so `//stack 1,0,0` is always a full selection in the X direction. This limits interesting stacking opportunities.

**Describe the solution you'd like**
`//stack` should have a new flag, `-r` (similar to `//gen`), that changes the offset units to blocks. It will be checked to ensure you do not stack inside of your selection, you _must_ stack outside of the selection.

**Describe alternatives you've considered**
You can use alternating `//stack` and `//move` to shift things into place, but it is tedious.

**Additional context**
Forked from #1363.

me4502:
Adds a method of using //stack with block units rather than selection units (`-r` flag). 

Closes #1443

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -99,6 +99,7 @@ import com.sk89q.worldedit.regions.CylinderRegion;
 import com.sk89q.worldedit.regions.EllipsoidRegion;
 import com.sk89q.worldedit.regions.FlatRegion;
 import com.sk89q.worldedit.regions.Region;
+import com.sk89q.worldedit.regions.RegionOperationException;
 import com.sk89q.worldedit.regions.Regions;
 import com.sk89q.worldedit.regions.shape.ArbitraryBiomeShape;
 import com.sk89q.worldedit.regions.shape.ArbitraryShape;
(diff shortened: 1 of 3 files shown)


wizjany:
haven't tested yet but doesn't this need to use a buffering extent like //move if you stack with a number of blocks less than the selection along that axis? (i.e. when the stacked blocks would overlap the original blocks)

me4502:
This specifically disallows stacking that would overlap the existing region, the issue octyl made mentioned to disallow that so I implemented it that way.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1417,6 +1418,42 @@ public int stackCuboidRegion(Region region, BlockVector3 offset, int count,
...
+     * @param count the number of times to stack
+     * @param copyEntities true to copy entities
+     * @param copyBiomes true to copy biomes
+     * @param mask source mask for the operation (only matching blocks are copied)
+     * @return number of blocks affected
+     * @throws MaxChangedBlocksException thrown if too many blocks are changed
+     * @throws RegionOperationException thrown if the region operation is invalid
+     */
+    public int stackCuboidRegionBlockUnits(CuboidRegion region, BlockVector3 offset, int count,
+                                 boolean copyEntities, boolean copyBiomes, Mask mask) throws MaxChangedBlocksException, RegionOperationException {
```
octylFractal, line 1435: format pls

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1417,6 +1418,42 @@ public int stackCuboidRegion(Region region, BlockVector3 offset, int count,
         return copy.getAffected();
     }
 
+    /**
+     * Stack a cuboid region using block units.
```
octylFractal, line 1422: Maybe suggest using `getBoundingBox()`?

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1417,6 +1418,42 @@ public int stackCuboidRegion(Region region, BlockVector3 offset, int count,
...
+     * @throws RegionOperationException thrown if the region operation is invalid
+     */
+    public int stackCuboidRegionBlockUnits(CuboidRegion region, BlockVector3 offset, int count,
+                                 boolean copyEntities, boolean copyBiomes, Mask mask) throws MaxChangedBlocksException, RegionOperationException {
+        checkNotNull(region);
+        checkNotNull(offset);
+        checkArgument(count >= 1, "count >= 1 required");
+
+        BlockVector3 size = region.getMaximumPoint().subtract(region.getMinimumPoint()).add(1, 1, 1);
+        if (offset.getX() < size.getX() && offset.getY() < size.getY() && offset.getZ() < size.getZ()) {
```
octylFractal, line 1441: Is this reliable for negative offsets?

octylFractal (changes requested):
You can still implement `stackCuboidRegion` in terms of `stackCuboidRegionBlockUnits`, just do the size multiply.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
me4502: I'm unsure if it makes sense to tbh, as that includes areas outside the region itself

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
```diff
@@ -381,7 +382,12 @@ public int stack(Actor actor, World world, EditSession editSession, LocalSession
...
 
-        int affected = editSession.stackCuboidRegion(region, offset, count, copyEntities, copyBiomes, combinedMask);
+        int affected = 0;
+        if (blockUnits) {
+            editSession.stackRegionBlockUnits(region, offset, count, copyEntities, copyBiomes, combinedMask);
+        } else {
+            editSession.stackCuboidRegion(region, offset, count, copyEntities, copyBiomes, combinedMask);
+        }
 
         if (moveSelection) {
```
octylFractal, line 392: As a reminder, this still needs to be updated.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
```diff
@@ -381,19 +382,25 @@ public int stack(Actor actor, World world, EditSession editSession, LocalSession
             combinedMask = mask;
         }
 
-        int affected = editSession.stackCuboidRegion(region, offset, count, copyEntities, copyBiomes, combinedMask);
+        int affected = 0;
```
octylFractal, line 385: Shouldn't this be unset, and assigned in both `if` branches?

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1401,13 +1402,44 @@ public int stackCuboidRegion(Region region, BlockVector3 offset, int count,
                                  boolean copyEntities, boolean copyBiomes, Mask mask) throws MaxChangedBlocksException {
         checkNotNull(region);
         checkNotNull(offset);
+
+        BlockVector3 size = region.getMaximumPoint().subtract(region.getMinimumPoint()).add(1, 1, 1);
+        try {
+            return stackRegionBlockUnits(region, offset.multiply(size), count, copyEntities, copyBiomes, mask);
+        } catch (RegionOperationException ignored) {
+            // Should never be able to happen
+            return 0;
```
octylFractal, line 1411: Can we just throw an `AssertionError` here? No reason to silently fail.

octylFractal (changes requested):

octylFractal (approved):

PR #1541, opened 2020-09-29
Fixed snow being placed through slabs

me4502:
Uses the game's internal "canPlace" system to determine whether snow is capable of being placed at the location.

Adapters PR: https://github.com/EngineHub/worldedit-adapters/pull/21

Fixes https://github.com/EngineHub/WorldEdit/issues/1516


worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
```diff
@@ -67,8 +68,7 @@ public boolean isGround(BlockVector3 position) {
         }
 
         // Can only place on full solid blocks
```
octylFractal, line 70: Update the comment plox.

worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
```diff
@@ -116,6 +116,9 @@ public boolean apply(BlockVector3 position, int depth) throws WorldEditException
             }
             return false;
         }
+        if (this.extent instanceof World && !((World) this.extent).canPlaceAt(abovePosition, snow)) {
```
octylFractal, line 119: Just take and store `World`, if we want to use `Extent` in the future we can just change the constructor (that's backwards-compatible).

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
me4502: Due to the extent here being the one that applies blocks, we'd need to pass in a second world to do that. Instead, I've made this require a world or EditSession to be passed in

worldedit-core/src/main/java/com/sk89q/worldedit/function/block/SnowSimulator.java
```diff
@@ -42,6 +46,10 @@
     private int affected;
 
     public SnowSimulator(Extent extent, boolean stack) {
+        checkArgument(extent instanceof World
```
octylFractal, line 49: Unwrap the world into a separate field in constructor maybe?

octylFractal (changes requested):

octylFractal (approved):

PR #1543, opened 2020-10-01
Add a way to get a safe teleport location in the world object

JOO200:
I want to use a method to get safe teleport locations with WorldEdit methods. Currently the only method to teleport a player to a safe place teleports the player directly to that place.

I need some better way for my WorldGuard PR: https://github.com/EngineHub/WorldGuard/pull/1621

Basically I moved the code from AbstractPlayerActor to World. I don't think that code is perfect now and I want to have some hints to find a safe teleport location. E.g. the current method teleports you into liquids like water and lava. That's not a safe teleport location. If that's intended I should rename the getSafeTeleportLocation to getFreeTeleportLocation.

Maybe we could add some boolean value to force safe teleport locations for that method.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/AbstractPlayerActor.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/AbstractPlayerActor.java
@@ -108,39 +108,14 @@ public abstract class AbstractPlayerActor implements Actor, Player, Cloneable {
     public void findFreePosition(Location searchPos) {
         Extent world = searchPos.getExtent();
 
+        if (!(world instanceof World)) {
+            throw new IllegalArgumentException("Unable to teleport a player to a non world type.");
+        }
+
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/AbstractWorld.java
```diff
@@ -160,6 +162,33 @@ public void setWeather(WeatherType weatherType) {
...
+        int x = searchPos.getBlockX();
+        int y = Math.max(yMin, searchPos.getBlockY());
+        int origY = y;
+        int z = searchPos.getBlockZ();
+        int yPlusSearchHeight = y + WorldEdit.getInstance().getConfiguration().defaultVerticalHeight;
+        int maxY = Math.min(yMax, yPlusSearchHeight) + 2;
+
+        byte free = 0;
+
+            while (y <= maxY) {
```
me4502, line 176: Formatting seems broken here

Issue #1442, opened 2020-07-17
Take offset input via facing + arbitrary vector


octylFractal:
**Is your feature request related to a problem? Please describe.**
Currently, we accept arbitrary offsets, but they cannot be auto-adjusted to match your facing direction. We need new syntax and complicated math.

**Describe the solution you'd like**
I don't have a solution in mind, this is open to suggestions on how we could parse a "facing arbitrary vector".

**Describe alternatives you've considered**
The alternative is we do not add this feature, and users must manually change vectors to match their facing direction. It's not a great experience.

**Additional context**
Forked from #1363.

octylFractal:
Note: this will be `status:pending` and not worked on until syntax is suggested. That is the big blocker on this.

octylFractal:
Current proposal is to duplicate Mojang's syntax, `^`: https://minecraft.gamepedia.com/Commands#Local_coordinates:_Caret_notation

Usage is `//stack ^5,^0,^3`, i.e. you must label each axis for it to work entirely. Partial labelling will be supported, but what happens if you manage to loop a stack through itself is undefined :)

PR #1545, opened 2020-10-02
Add ^x,y,z relative offset support to the offset parser

me4502:
Adds a `^` prefix to the offset parser that signifies the vector is a `forwards,up,left` offset rather than raw x/y/z
 
Fixes https://github.com/EngineHub/WorldEdit/issues/1442

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/OffsetConverter.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/OffsetConverter.java
@@ -22,8 +22,14 @@ package com.sk89q.worldedit.command.argument;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.sk89q.worldedit.WorldEdit;
+import com.sk89q.worldedit.extension.platform.Actor;
+import com.sk89q.worldedit.extension.platform.Locatable;
 import com.sk89q.worldedit.internal.annotation.Offset;
 import com.sk89q.worldedit.math.BlockVector3;
+import com.sk89q.worldedit.math.MathUtils;
+import com.sk89q.worldedit.math.Vector3;
+import com.sk89q.worldedit.math.transform.AffineTransform;
+import com.sk89q.worldedit.util.Location;
 import com.sk89q.worldedit.util.formatting.text.Component;
 import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import org.enginehub.piston.CommandManager;
@@ -33,6 +39,7 @@ import org.enginehub.piston.inject.InjectedValueAccess;
 import org.enginehub.piston.inject.Key;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class OffsetConverter implements ArgumentConverter<BlockVector3> {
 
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/OffsetConverter.java
```diff
@@ -62,6 +67,11 @@ public Component describeAcceptableArguments() {
 
     @Override
     public List<String> getSuggestions(String input, InjectedValueAccess context) {
+        if (input.startsWith("^") && context.injectedValue(Key.of(Player.class)).isPresent()) {
```
wizjany, line 70: any reason we require `Player` here? would any `Locatable` sender do? (even tho commandblocks' pitch/yaw are probably always `0,0` in theory mobs can be commandsenders too though we don't support them yet)

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/OffsetConverter.java
```diff
@@ -70,7 +83,47 @@ public Component describeAcceptableArguments() {
...
+                m1.getZ(), m2.getZ(), m3.getZ(), 0
+            );
+
+            return vectorConverter.convert(input.substring(1), context).map(blockVector3s ->
+                blockVector3s.stream().map(vector -> transform
+                    .apply(vector.toVector3())
+                    // This *MUST* be rounded before converting to block points
+                    .round().toBlockPoint()
+                ).collect(Collectors.toList())
+            );
```
octylFractal, line 123: Extract to a method.

worldedit-core/src/main/java/com/sk89q/worldedit/math/MathUtils.java
```diff
@@ -66,11 +66,10 @@ public static double dCos(double degrees) {
                 case 0:
                     return 1.0;
                 case 90:
+                case 270:
```
octylFractal, line 69: Revert these changes.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/OffsetConverter.java
```diff
@@ -62,15 +69,66 @@ public Component describeAcceptableArguments() {
...
+    }
+
     @Override
     public ConversionResult<BlockVector3> convert(String input, InjectedValueAccess context) {
-        return directionVectorConverter.convert(input, context)
-            .orElse(vectorConverter.convert(input, context));
+        if (input.startsWith("^")) {
+            // Looking at a relative vector.
+            Actor actor = context.injectedValue(Key.of(Actor.class))
+                .orElseThrow(() -> new IllegalStateException("An actor is required to use relative offsets"));
```
octylFractal, line 116: oh yea, you're uh, not supposed to throw here. Return a `FailedConversion.from(ex)` instead, or just wrap the whole thing in a `try-catch` :P

octylFractal (changes requested):

octylFractal (approved):

PR #1546, opened 2020-10-03
Make block positions copy on click for //size

### LINKED ISSUE #1329: Allow copying of region corner coordinates from //size to clipboard
It would be awesome if there were a command, that just pastes the coordinates of the corners of your (cuboid) selection to your cliboard. You basically do your selection, write //xywhatever and then you can use Ctrl + V to paste the corner coordinates into a command block for example.

I hope you can't do this already... but I didnt find something.

lewisjb:
Fixes #1329

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/CylinderRegionSelector.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/CylinderRegionSelector.java
@@ -248,7 +248,7 @@ public class CylinderRegionSelector implements RegionSelector, CUIRegion {
         if (!region.getCenter().equals(Vector3.ZERO)) {
             Vector3 center = region.getCenter();
             lines.add(TranslatableComponent.of("worldedit.selection.cylinder.info.center", TextComponent.of(center.toString())
-                    .clickEvent(ClickEvent.of(ClickEvent.Action.COPY_TO_CLIPBOARD, center.getX() + "," + center.getY() + "," + center.getZ()))
+                    .clickEvent(ClickEvent.of(ClickEvent.Action.COPY_TO_CLIPBOARD, center.toParserString()))
                     .hoverEvent(HoverEvent.of(HoverEvent.Action.SHOW_TEXT, TextComponent.of("Click to copy")))));
         }
         if (!region.getRadius().equals(Vector2.ZERO)) {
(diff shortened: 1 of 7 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/CylinderRegionSelector.java
```diff
@@ -244,7 +246,10 @@ public String getTypeName() {
         final List<Component> lines = new ArrayList<>();
 
         if (!region.getCenter().equals(Vector3.ZERO)) {
-            lines.add(TranslatableComponent.of("worldedit.selection.cylinder.info.center", TextComponent.of(region.getCenter().toString())));
+            Vector3 center = region.getCenter();
+            lines.add(TranslatableComponent.of("worldedit.selection.cylinder.info.center", TextComponent.of(center.toString())
+                    .clickEvent(ClickEvent.of(ClickEvent.Action.COPY_TO_CLIPBOARD, center.getX() + "," + center.getY() + "," + center.getZ()))
```
octylFractal, line 251: Why add toCopyableString for BlockVector3 but not Vector?

Might we also want the 2D ones to acquire this ability? @wizjany @me4502

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/CylinderRegionSelector.java
lewisjb: Good question - no real reason, thought Vector3 might've been too generic to add something like this to?

wizjany: tbh I'm slightly confused as to why toCopyableString exists at all. Is there something undesirable about parens?

me4502: Yeah, I feel that we should have it on all the vectors if we have it on any.

Also I'm unsure whether 'toCopyableString' best represents what it is, as it's mostly just a string representation of the vector

lewisjb: My understanding is that the parens are undesirable for use in commands, such as `//pos x,y,z`, you don't have to manually strip the parens yourself. And I didn't want to change the default `toString` as having parens in other places can be useful?

me4502 (approved):

wizjany:
Anyone know off-hand/tested what happens if we send click-to-copy json to a 1.13/1.14 client, as per the linked issue? I don't mind lack of functionality but I'd prefer to avoid errors. Otherwise lgtm.

me4502:
kashike confirmed that it'll get ignored if it's sent to an older client, so I'll merge this.

Thanks!

PR #1547, opened 2020-10-03
Add splatter brush

### LINKED ISSUE #1524: new brush: splatter
**Is your feature request related to a problem? Please describe.**
Currently, there is no simple way to create a splatter design over a large surface that I am aware of. You can slowly do it by creating different brushes, but it's slow and painful to do.

**Describe the solution you'd like**
A brush that essentially does this:
image

It's parameters would be a **block**, **radius**, and **fall-off** strength. A higher fall-off strength would create a sphere brush pattern. A lower fall-off strength would create a messier pattern.

image
The left would be high fall-off, the right would be low fall-off.

**Describe alternatives you've considered**
Alternative solutions are to use GoPaint (which offers this brush) however that plugin requires FAWE which is completely unusable. Another alternative is to create different brushes with percentage patterns to slowly create this effect. Emphasis on slowly.

**Additional context**
I believe that everything. If you have any questions let me know.

lewisjb:
Fixes #1524

image
(going left-to-right, it's the different decay values of 0-10 for radius 5)

Tried a bunch of different approaches to achieve this and I felt this was pretty close to the original ticket - happy if others have a better approach

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -1740,15 +1740,15 @@ public class EditSession implements Extent, AutoCloseable {
      *
      * @param pos Center of the splatter
      * @param block The block pattern to use
-     * @param radius The cylinder's radius
+     * @param radius The sphere's radius
      * @param decay The higher the number the more decay, between 0 and 10
      * @return number of blocks changed
      * @throws MaxChangedBlocksException thrown if too many blocks are changed
      */
     public int makeSplatter(BlockVector3 pos, Pattern block, double radius, int decay) throws MaxChangedBlocksException {
         //  decay is 0 - 10, 0 should be a circle.
-        checkArgument(decay >= 0, "decay >= 0");
-        checkArgument(decay <= 10, "decay <= 10");
+        checkArgument(decay >= 0, "decay must be >= 0");
+        checkArgument(decay <= 10, "decay must be <= 10");
         radius += 0.5;
 
         int affected = 0;
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
@@ -182,7 +182,10 @@ public class BrushCommands {
                                   int decay) throws WorldEditException {
         worldEdit.checkMaxBrushRadius(radius);
 
-        decay = Math.max(Math.min(decay, 10), 0);
+        if (decay < 0 || decay > 10) {
+            player.printError(TranslatableComponent.of("worldedit.brush.splatter.decay-out-of-range", TextComponent.of(decay)));
+            return;
+        }
 
         BrushTool tool = session.getBrushTool(player.getItemInHand(HandSide.MAIN_HAND).getType());
         tool.setFill(pattern);
--- a/worldedit-core/src/main/resources/lang/strings.json
+++ b/worldedit-core/src/main/resources/lang/strings.json
@@ -22,6 +22,7 @@
     "worldedit.brush.sphere.equip": "Sphere brush shape equipped ({0}).",
     "worldedit.brush.cylinder.equip": "Cylinder brush shape equipped ({0} by {1}).",
     "worldedit.brush.splatter.equip": "Splatter brush shape equipped ({0} with {1} decay).",
+    "worldedit.brush.splatter.decay-out-of-range": "Splatter brush decay value: {0} is out of range 0 - 10.",
     "worldedit.brush.clipboard.equip": "Clipboard brush shape equipped.",
     "worldedit.brush.smooth.equip": "Smooth brush equipped ({0} x {1}x using {2}).",
     "worldedit.brush.extinguish.equip": "Extinguisher equipped ({0}).",


me4502 (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1732,6 +1733,87 @@ public int makeCylinder(BlockVector3 pos, Pattern block, double radiusX, double
...
+     * @param pos Center of the splatter
+     * @param block The block pattern to use
+     * @param radius The cylinder's radius
+     * @param decay The higher the number the more decay
+     * @return number of blocks changed
+     * @throws MaxChangedBlocksException thrown if too many blocks are changed
+     */
+    public int makeSplatter(BlockVector3 pos, Pattern block, double radius, int decay) throws MaxChangedBlocksException {
+        //  decay is 0 - 10, 0 should be a circle.
+        final Random random = new Random();
```
wizjany, line 1748: Can probably used a shared random here.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1732,6 +1733,87 @@ public int makeCylinder(BlockVector3 pos, Pattern block, double radiusX, double
...
+
+        radius += 0.5;
+
+        int affected = 0;
+
+        if (decay <= 0) {
```
wizjany, line 1754: could just as well be `< 0`.
these hard limits should probably also be noted in the param javadoc, and probably also `checkArgument`ed, at which point this clamping can be removed. (invisibly correcting mistakes leads to confusion, errors and documentation are explicit).

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1732,6 +1733,87 @@ public int makeCylinder(BlockVector3 pos, Pattern block, double radiusX, double
...
+            decay = 10;
+        }
+
+        final float normDecay = (float) decay / 10;
+
+        // Inner radius will be a normal circle, outer radius will have specks
+        double outerRadius = radius + (radius * 0.5 * normDecay);
+        double innerRadius = radius - radius * normDecay;
+
+        if (pos.getBlockY() < world.getMinY()) {
```
wizjany, line 1766: both checks here can be replaced with (a single) `BlockVector3.clampY`

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1732,6 +1733,87 @@ public int makeCylinder(BlockVector3 pos, Pattern block, double radiusX, double
...
+
+                double distanceSq = lengthSq(xn, zn);
+                if (distanceSq > 1) {
+                    if (z == 0) {
+                        break forX;
+                    }
+                    break forZ;
+                }
+
+                double distanceFromCenter = Math.sqrt(lengthSq(x, z));
```
wizjany, line 1793: `lengthSq` calculation is being repeated here (already `distanceSq`) and you can just store an `innerRadiusSq` instead of doing sqrt every time here.

wizjany (changes requested):
Aside from review comments, it seems odd to me that this just does a single layer x-z circle. Any reason this can't be a sphere and allow people to mask it to solid/surface blocks if that's what they want (as in your screenshot)? Feel like this could be just as useful on walls, etc.

lewisjb:
Good point - updated to be a 3d splatter - along with other feedback

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1732,6 +1735,85 @@ public int makeCylinder(BlockVector3 pos, Pattern block, double radiusX, double
...
+     * @param pos Center of the splatter
+     * @param block The block pattern to use
+     * @param radius The cylinder's radius
+     * @param decay The higher the number the more decay, between 0 and 10
+     * @return number of blocks changed
+     * @throws MaxChangedBlocksException thrown if too many blocks are changed
+     */
+    public int makeSplatter(BlockVector3 pos, Pattern block, double radius, int decay) throws MaxChangedBlocksException {
+        //  decay is 0 - 10, 0 should be a circle.
+        checkArgument(decay >= 0, "decay >= 0");
```
octylFractal, line 1750: This should probably be something like `"decay must be greater than or equal to 0"`, or just `"decay must be >= 0"`.

Similar for the line below.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -1732,6 +1735,85 @@ public int makeCylinder(BlockVector3 pos, Pattern block, double radiusX, double
         return affected;
     }
 
+    /**
+     * Makes a splatter.
+     *
+     * @param pos Center of the splatter
+     * @param block The block pattern to use
+     * @param radius The cylinder's radius
```
octylFractal, line 1743: Isn't this a sphere?

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -166,6 +167,33 @@ public void cylinderBrush(Player player, LocalSession session,
...
+    public void splatterBrush(Player player, LocalSession session,
+                              @Arg(desc = "The pattern of blocks to set")
+                                  Pattern pattern,
+                              @Arg(desc = "The radius of the splatter", def = "2")
+                                  double radius,
+                              @Arg(desc = "The decay of the splatter between 0 and 10", def = "1")
+                                  int decay) throws WorldEditException {
+        worldEdit.checkMaxBrushRadius(radius);
+
+        decay = Math.max(Math.min(decay, 10), 0);
```
octylFractal, line 185: I would rather bail out with an error than silently correct.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
lewisjb: Sure thing - I just copied the pattern from other functions like `fillXZ` and `removeAbove`

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
lewisjb: Good catch

me4502 (approved):

wizjany (approved):

PR #1580, opened 2020-11-09
Add Snow Smooth Tools

JOO200:
I want to add some tools to make smooth snow areas. I added a //snowsmooth command and a snowsmooth brush for that.

e.g.:
grafik

The implemented brush does work like the //snowsmooth command. I added some parameters to //snowsmooth to change the GaussianKernel but I have no problem to remove that parameter.

Currently there is some minor issue for smoothing areas without a block at the lowest level.
grafik
I'll dig into that a little further to remove, feel free to make notes.


SirYwell:
If this doesn't *only* smooth snow, I'd personally prefer a `-s` switch over separate commands

JOO200:
It does smooth blocks but has some other parameters. I agree that some `-s` switch could be a nice thing but with the property to define the amount of snow blocks under snow layers it's a little bit more complex.

I open this as PR now as I fixed all issues with offset-errors and weird snow-blocks at lower bounds.

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
```diff
@@ -0,0 +1,213 @@
...
+        for (int z = 0; z < height; ++z) {
+            for (int x = 0; x < width; ++x) {
+                int highestBlockY = session.getHighestTerrainBlock(x + minX, z + minZ, minY, maxY, mask);
+                BlockState upper = session.getBlock(BlockVector3.at(x + minX, highestBlockY + 1, z + minZ));
+                if (upper.getBlockType() == BlockTypes.SNOW) {
+                    Integer amountLayers = (Integer) upper.getState(layers);
+                    data[z * width + x] = (highestBlockY + 1 + (((float) amountLayers - 1) / 8));
+                } else {
+                    BlockState block = session.getBlock(BlockVector3.at(x + minX, highestBlockY, z + minZ));
+                    if (block.getBlockType() == BlockTypes.AIR) {
```
me4502, line 85: This should be using the `block.getBlockType().getMaterial().isAir()` method rather than doing this equality check

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
```diff
@@ -0,0 +1,213 @@
...
+                // We are keeping the topmost blocks so take that in account for the scale
+                double scale = (double) (curHeight - originY) / (double) (newHeight - originY);
+
+                // Depending on growing or shrinking we need to start at the bottom or top
+                if (newHeight >= curHeight) {
+                    // Set the top block of the column to be the same type (this might go wrong with rounding)
+                    BlockState existing = session.getBlock(BlockVector3.at(xr, curHeight, zr));
+
+                    // Skip water/lava
+                    if (existing.getBlockType() != BlockTypes.WATER && existing.getBlockType() != BlockTypes.LAVA) {
```
me4502, line 164: Should use isLiquid on the material rather than checking exact types

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/HeightMap.java
```diff
@@ -90,7 +91,7 @@ public int applyFilter(HeightMapFilter filter, int iterations) throws MaxChanged
         System.arraycopy(data, 0, newData, 0, data.length);
 
         for (int i = 0; i < iterations; ++i) {
-            newData = filter.filter(newData, width, height);
+            newData = filter.filter(newData, width, height, 0.5F);
```
me4502, line 94: Does this modify prior behaviour?

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
```diff
@@ -0,0 +1,213 @@
...
+ */
+public class SnowHeightMap {
+    private final float[] data;
+    private final int width;
+    private final int height;
+
+    private final Region region;
+    private final EditSession session;
+
+    private final Property<Object> layers;
```
me4502, line 48: You should store this as Property<Integer> and cast when getting the property, rather than casting on the getters and setters

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -203,6 +204,7 @@
     "worldedit.naturalize.naturalized": "{0} block(s) have been made to look more natural.",
     "worldedit.center.changed": "Center set. ({0} blocks changed)",
     "worldedit.smooth.changed": "Terrain's height map smoothed. {0} blocks changed.",
+    "worldedit.snowsmooth.changed": "Terrain's height map smoothed. {0} blocks changed.",
```
me4502, line 207: Should this make some mention of snow? Like "Snow's height map smoothed" etc

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
JOO200: I'm a little bit confused. I should store the Property as Propert<?>? But why?
e.g. I found [this class](https://github.com/EngineHub/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/BannerBlockCompatibilityHandler.java#L40) using Properties and there is the cast on the getter. I would change the Property<Object> to Property<Integer> and add a try-catch.

wizjany: can probably be stored as an IntegerProperty directly, we're not very consistent throughout the project on using this correctly lol.

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
```diff
@@ -0,0 +1,218 @@
...
+                        ++blocksChanged;
+
+                        // Grow -- start from 1 below top replacing airblocks
+                        for (int y = (int) newHeight - 1 - originY; y >= 0; --y) {
+                            if (y >= newHeight - 1 - originY - layerBlocks) {
+                                session.setBlock(BlockVector3.at(xr, originY + y, zr), fillerSnow);
+                            } else {
+                                int copyFrom = (int) (y * scale);
+                                session.setBlock(BlockVector3.at(xr, originY + y, zr),
+                                        session.getBlock(BlockVector3.at(xr, originY + copyFrom, zr)));
```
octylFractal, line 180: I think I might prefer if you pulled this out to its own variable?

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
```diff
@@ -0,0 +1,218 @@
...
+                    }
+                } else {
+                    // Shrink -- start from bottom
+                    for (int y = 0; y < (int) newHeight - originY; ++y) {
+                        if (y >= (int) newHeight - originY - layerBlocks) {
+                            session.setBlock(BlockVector3.at(xr, originY + y, zr), fillerSnow);
+                        } else {
+                            int copyFrom = (int) (y * scale);
+                            session.setBlock(BlockVector3.at(xr, originY + y, zr),
+                                    session.getBlock(BlockVector3.at(xr, originY + copyFrom, zr)));
```
octylFractal, line 193: > I think I might prefer if you pulled this out to its own variable?

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
```diff
@@ -0,0 +1,218 @@
...
+                    }
+                }
+            }
+        }
+
+        // Drop trees to the floor -- TODO
+        return blocksChanged;
+    }
+
+    private void setSnowLayer(int x, int z, float newHeight, IntegerProperty property) throws MaxChangedBlocksException {
```
octylFractal, line 214: Why the hEcK are you taking `property` here? `layers` is an instance variable, you can use it in this method?

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
```diff
@@ -0,0 +1,218 @@
...
+        int layers = (int) ((newHeight % 1) * 8) + 1;
+        session.setBlock(BlockVector3.at(x, (int) newHeight, z), BlockTypes.SNOW.getDefaultState().with(property, layers));
```
octylFractal, line 216:
```suggestion
        session.setBlock(BlockVector3.at(x, (int) newHeight, z), BlockTypes.SNOW.getState(ImmutableMap.of(property, layers));
```
slightly more efficient :)

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
```diff
@@ -291,6 +292,32 @@ public int smooth(Actor actor, EditSession editSession, @Selection Region region
         return affected;
     }
 
+    @Command(
+        name = "/snowsmooth",
+        desc = "Smooth the elevation in the selection with snow layers",
+        descFooter = "Example '//snowsmooth 1 snow_block,snow' would only smooth snow terrain."
```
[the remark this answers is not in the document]
JOO200: Shoul be consistent to the //smooth command. The smooth comman does have
`Example: '//smooth <cmd example>' would only smooth natural surface terrain`

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
```diff
@@ -0,0 +1,218 @@
...
+        checkNotNull(region);
+
+        this.session = session;
+        this.region = region;
+
+        this.width = region.getWidth();
+        this.height = region.getLength();
+
+        try {
+            layers = (IntegerProperty) (Property<?>) BlockTypes.SNOW.getProperty("layers");
```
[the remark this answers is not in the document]
JOO200: Copied from here: https://github.com/EngineHub/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/legacycompat/NoteBlockCompatibilityHandler.java#L37
I can use a Property<Integer> too, @wizjany suggested to use IntegerProprty

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/SnowHeightMap.java
```diff
@@ -0,0 +1,218 @@
...
+        } catch (NullPointerException | IllegalArgumentException | ClassCastException e) {
```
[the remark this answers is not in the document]
JOO200: Agree, will be catched by the exception handler for commands

octylFractal:
@JOO200 poking you for updates on this, it's outdated & there are still some review comments that weren't addressed.

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
```diff
@@ -291,6 +292,32 @@ public int smooth(Actor actor, EditSession editSession, @Selection Region region
...
+    @CommandPermissions("worldedit.region.snowsmooth")
+    @Logging(REGION)
+    public int snowSmooth(Actor actor, EditSession editSession, @Selection Region region,
+                          @Arg(desc = "# of iterations to perform", def = "1")
+                              int iterations,
+                          @Arg(desc = "The mask of blocks to use as the height map", def = "")
+                              Mask mask,
+                          @ArgFlag(name = 'l', desc = "Set the amount of snow blocks under the snow layers", def = "1")
+                              int snowBlockGroundLayers,
+                          @ArgFlag(name = 'r', desc = "Set the radius for the GaussianKernel", def = "20")
```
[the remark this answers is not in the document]
JOO200: i removed the flags. it should be no big deal to add the flags later

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/HeightMapFilter.java
```diff
@@ -80,6 +80,20 @@ public void setKernel(Kernel kernel) {
...
+     * Filter with a 2D kernel.
+     *
+     * @param inData the data
+     * @param width the width
+     * @param height the height
+     * @param offset the offset added to the height
+     *
+     * @return the modified height map
+     */
+    public int[] filter(int[] inData, int width, int height, float offset) {
```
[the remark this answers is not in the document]
JOO200: I wanted to have a matching float and int signature for that filter method.
Not sure about the delegation, I don't see a good method to convert a int array to a float array (and back to an int array)

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -272,6 +273,36 @@ public void smoothBrush(Player player, LocalSession session,
...
+
+        BrushTool tool = session.getBrushTool(player.getItemInHand(HandSide.MAIN_HAND).getType());
+        tool.setSize(radius);
+        tool.setBrush(new SnowSmoothBrush(iterations, snowBlockCount, mask), "worldedit.brush.snowsmooth");
+
+        player.printInfo(TranslatableComponent.of(
+                "worldedit.brush.snowsmooth.equip",
+                TextComponent.of((int) radius),
+                TextComponent.of(iterations),
+                TextComponent.of(mask == null ? "any block" : "filter"),
```
octylFractal, line 301: This should probably be translatable. They're nestable, so you just need to add keys for it and change the component.

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -272,6 +273,36 @@ public void smoothBrush(Player player, LocalSession session,
...
+        desc = "Choose the snow terrain softener brush",
+        descFooter = "Example: '/brush snowsmooth 5 1 3'"
+    )
+    @CommandPermissions("worldedit.brush.snowsmooth")
+    public void snowSmoothBrush(Player player, LocalSession session,
+                                @Arg(desc = "The radius to sample for softening", def = "2")
+                                    double radius,
+                                @Arg(desc = "The number of iterations to perform", def = "4")
+                                    int iterations,
+                                @Arg(desc = "The number of snow blocks under snow", def = "1")
```
octylFractal, line 287: Make this a flag like in `//snowsmooth`.

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -272,6 +273,36 @@ public void smoothBrush(Player player, LocalSession session,
...
+                                    int snowBlockCount,
+                                @Arg(desc = "The mask of blocks to use for the heightmap", def = "")
```
octylFractal, line 289: I think `mask` belong as an `@ArgFlag(name = 'm')`, like with other commands. It's different from `//smooth`, but it's more usable IMO. If you change it, also change `//snowsmooth`.

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/HeightMapFilter.java
```diff
@@ -80,50 +80,109 @@ public void setKernel(Kernel kernel) {
      * @return the modified height map
      */
     public int[] filter(int[] inData, int width, int height) {
+        return filter(inData, width, height, 0.5F);
```
octylFractal, line 83: I'm still not sure why you added this override:
> I am not entirely sure why you added this method, the only use of it also passed 0.5F?

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/math/convolution/HeightMapFilter.java
JOO200: Same method signature as filter(float[]...) and it doesn't hurt. maybe someone want's to use a different offset as 0.5?

octylFractal: Sure, I suppose.

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -267,11 +268,45 @@ public void smoothBrush(Player player, LocalSession session,
                 "worldedit.brush.smooth.equip",
                 TextComponent.of((int) radius),
                 TextComponent.of(iterations),
-                TextComponent.of(mask == null ? "any block" : "filter")
+                mask == null
+                        ? TranslatableComponent.of("worldedit.brush.smooth.nofilter")
+                        : TranslatableComponent.of("worldedit.brush.smooth.filter")
```
octylFractal, line 273:
```suggestion
                TranslatableComponent.of("worldedit.brush.smooth." + (mask == null ? "no" : "") + "filter")
```

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -267,11 +268,45 @@ public void smoothBrush(Player player, LocalSession session,
...
+        tool.setSize(radius);
+        tool.setBrush(new SnowSmoothBrush(iterations, snowBlockCount, mask), "worldedit.brush.snowsmooth");
+
+        player.printInfo(TranslatableComponent.of(
+                "worldedit.brush.snowsmooth.equip",
+                TextComponent.of((int) radius),
+                TextComponent.of(iterations),
+                mask == null
+                        ? TranslatableComponent.of("worldedit.brush.snowsmooth.nofilter")
+                        : TranslatableComponent.of("worldedit.brush.snowsmooth.filter"),
```
octylFractal, line 305:
```suggestion
                TranslatableComponent.of("worldedit.brush.snowsmooth." + (mask == null ? "no" : "") + "filter")
```

worldedit-core/src/main/java/com/sk89q/worldedit/command/RegionCommands.java
```diff
@@ -291,6 +292,28 @@ public int smooth(Actor actor, EditSession editSession, @Selection Region region
...
+    )
+    @CommandPermissions("worldedit.region.snowsmooth")
+    @Logging(REGION)
+    public int snowSmooth(Actor actor, EditSession editSession, @Selection Region region,
+                          @Arg(desc = "# of iterations to perform", def = "1")
+                              int iterations,
+                          @ArgFlag(name = 'm', desc = "The mask of blocks to use as the height map")
+                              Mask mask,
+                          @ArgFlag(name = 'l', desc = "Set the amount of snow blocks under the snow", def = "1")
+                              int snowBlockCount) throws WorldEditException {
```
octylFractal, line 308: Can you order these the other way around? I would like to keep the flag ordering the same.

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/SnowSmoothBrush.java
```diff
@@ -0,0 +1,68 @@
...
+        this.iterations = iterations;
+        this.mask = mask;
+        this.snowBlockLayer = snowBlockLayer;
+    }
+
+    @Override
+    public void build(EditSession editSession, BlockVector3 position, Pattern pattern, double size) throws MaxChangedBlocksException {
+        Vector3 posDouble = position.toVector3();
+        BlockVector3 min =  posDouble.subtract(size, size, size).toBlockPoint();
+        BlockVector3 max = posDouble.add(size, size + 10, size).toBlockPoint();
```
octylFractal, line 61: I wonder if this height should be a parameter. @wizjany @me4502 thoughts?

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/SnowSmoothBrush.java
JOO200: I copied that code from the SmoothBrush. The offset size there is hardcoded too.

octylFractal (approved):
Approving, but I'd like another pass from @wizjany or @me4502 to ensure everything is good.

me4502 (approved):

PR #1591, opened 2020-11-24
Schematic Share system

me4502:
This PR adds a command in game, that shares a schematic with a URL.

`/schem share`.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
@@ -19,6 +19,7 @@
 
 package com.sk89q.worldedit.command;
 
+import com.google.common.base.Charsets;
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.MoreFiles;
 import com.sk89q.worldedit.LocalConfiguration;
@@ -54,6 +55,7 @@ import com.sk89q.worldedit.util.formatting.text.format.TextColor;
 import com.sk89q.worldedit.util.io.Closer;
 import com.sk89q.worldedit.util.io.file.FilenameException;
 import com.sk89q.worldedit.util.io.file.MorePaths;
+import com.sk89q.worldedit.util.paste.EngineHubPaste;
 import org.enginehub.piston.annotation.Command;
 import org.enginehub.piston.annotation.CommandContainer;
 import org.enginehub.piston.annotation.param.Arg;
@@ -66,14 +68,17 @@ import org.slf4j.LoggerFactory;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.net.URL;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Comparator;
 import java.util.List;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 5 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -372,6 +408,49 @@ public Void call() throws Exception {
...
+                baos = closer.register(new ByteArrayOutputStream());
+                BufferedOutputStream bos = closer.register(new BufferedOutputStream(baos));
+                ClipboardWriter writer = closer.register(format.getWriter(bos));
+                writer.write(target);
+            } catch (IOException e) {
+                throw new CommandException(TextComponent.of(e.getMessage()), e, ImmutableList.of());
+            }
+
+            EngineHubPaste pasteService = new EngineHubPaste();
+            return pasteService.paste(new String(Base64.getEncoder().encode(baos.toByteArray()), Charsets.UTF_8)).call();
```
kashike, line 450:
```suggestion
            return pasteService.paste(new String(Base64.getEncoder().encode(baos.toByteArray()), StandardCharsets.UTF_8)).call();
```
No need to use Guava, the JDK has this.

kashike (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -146,7 +152,7 @@ public void load(Actor actor, LocalSession session,
 
     @Command(
         name = "save",
-        desc = "Save a schematic into your clipboard"
+        desc = "Save a schematic of your clipboard"
```
octylFractal, line 155:
```suggestion
        desc = "Save your clipboard into a schematic file"
```

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -205,6 +211,39 @@ public void save(Actor actor, LocalSession session,
                 .buildAndExec(worldEdit.getExecutorService());
     }
 
+    @Command(
+        name = "share",
+        desc = "Share a schematic into your clipboard"
```
octylFractal, line 216: smh you corrected the `save` and then made the same mistake here
```suggestion
        desc = "Share your clipboard as a schematic online"
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/paste/PasteMetadata.java
```diff
@@ -0,0 +1,29 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.util.paste;
+
+public class PasteMetadata {
+    public PasteMetadata() {
```
octylFractal, line 23: This is kinda pointless. Use the bits of syntax sugar Java gives you :P

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -372,6 +411,55 @@ public Void call() throws Exception {
...
+        SchematicShareTask(Actor actor, ClipboardFormat format, ClipboardHolder holder, String name) {
+            this.actor = actor;
+            this.format = format;
+            this.holder = holder;
+            this.name = name;
+        }
+
+        @Override
+        public URL call() throws Exception {
+            Clipboard clipboard = holder.getClipboard();
```
octylFractal, line 429: Can we merge the saving code here and with `save` into a single method? DRY applies here, especially with how much custom logic is part of saving.

octylFractal (changes requested):

octylFractal (changes requested):

octylFractal (approved):

PR #1595, opened 2020-11-27
Add instructions for unbinding tools/brushes on equip.

wizjany:
Will hopefully cut down on this question.

examples:

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
@@ -66,8 +66,11 @@ import com.sk89q.worldedit.util.TreeGenerator;
 import com.sk89q.worldedit.util.asset.AssetLoadTask;
 import com.sk89q.worldedit.util.asset.AssetLoader;
 import com.sk89q.worldedit.util.asset.holder.ImageHeightmap;
+import com.sk89q.worldedit.util.formatting.text.Component;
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -88,6 +91,10 @@
 
     private final WorldEdit worldEdit;
 
+    private static Component unbindCommandComponent = TextComponent.builder("/brush none", TextColor.AQUA)
+                                                                   .clickEvent(ClickEvent.runCommand("/brush none"))
```
octylFractal, line 95: I think we prefer suggesting instead.

worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolCommands.java
```diff
@@ -69,6 +71,10 @@
 @CommandContainer(superTypes = CommandPermissionsConditionGenerator.Registration.class)
 public class ToolCommands {
 
+    private static Component unbindCommandComponent = TextComponent.builder("/tool none", TextColor.AQUA)
+                                                                   .clickEvent(ClickEvent.runCommand("/tool none"))
```
octylFractal, line 75: Ditto here

octylFractal (changes requested):

octylFractal (approved):

me4502 (approved):

PR #1609, opened 2020-12-11
Properly repeat extents for negative coordinates

SirYwell:
An extent's content was returned mirrored/flipped when applied for negative positions, as e.g. `Math.abs(-2) % 3` returns 2 instead of 1 (as 1 + -1 * 3 = -2).

This PR fixes it by using [Math.floorMod()](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#floorMod-int-int-) which provides the properties of the mathematical congruence relation for positive divisors.


octylFractal:
Sorry, we haven't updated documentation but you need to base this off of the `version/7.2.x` branch for bug fixes. `master` is for features only.

PR #1618, opened 2020-12-19
Add ability to paste from clipboard to absolute position [-p x,y,z]

Virus610:
In case you want to paste something in a specific place, but you don't want to be there (Or from console)

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
@@ -163,14 +163,24 @@ public class ClipboardCommands {
                           boolean pasteBiomes,
                       @ArgFlag(name = 'm', desc = "Only paste blocks matching this mask")
                       @ClipboardMask
-                          Mask sourceMask) throws WorldEditException {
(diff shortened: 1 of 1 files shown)


wizjany:
pos1 x,y,z/toggleplace? rather not add redundant parameters that can already be accomplished. if we accept this we'd need to add it to literally every other command that uses placement position.

Virus610:
> 
> 
> pos1 x,y,z/toggleplace? rather not add redundant parameters that can already be accomplished. if we accept this we'd need to add it to literally every other command that uses placement position. (notably, toggleplace is already forced for console since it doesn't have an actual location, so "or from console" makes it sound like you didn't actually test pasting things from console...)

My bad. Didn't realize /toggleplace did that.

That said, I wouldn't dare say my code supports something if I hadn't first verified that.

wizjany:
I meant you didn't understand that you already have to set a pos1 to paste from console before your change.

Virus610:
Oh, you mean didn't test beforehand. Apologies, I thought you mean I hadn't tested my own code.

Honestly, this was just a feature request from a friend who said he looked stuff up, and it wasn't supported. ¯\_(ツ)_/¯

Apologies for the waste of time.

PR #1623, opened 2020-12-22
Add a command to apply side effects

me4502:
Adds a command to apply a side effect to the given region, or when no side effect is provided, the currently selected ones.

This fits a few use-cases:

* Allows re-calculating lighting (a fix lighting command)
* Allows telling fences, etc, to attempt to re-connect
* Allows triggering block updates (make sand fall, etc)

https://user-images.githubusercontent.com/546754/102883299-87915380-449b-11eb-8e11-4dbd439d2308.mp4

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
@@ -286,9 +286,7 @@ public class GeneralCommands {
 
     @Command(
         name = "/update",
-        desc = "Apply a side effect",
-        descFooter = "Note that this command is GOING to change in the future."
-            + " Do not depend on the exact format of this command yet."
+        desc = "Apply a side effect"
     )
     @CommandPermissions("worldedit.update")
     void update(Actor actor, LocalSession session, World injectedWorld,
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
@@ -32,13 +32,11 @@ import com.sk89q.worldedit.command.util.WorldEditAsyncCommandBuilder;
 import com.sk89q.worldedit.entity.Player;
 import com.sk89q.worldedit.extension.platform.Actor;
 import com.sk89q.worldedit.extension.platform.Capability;
-import com.sk89q.worldedit.function.RegionFunction;
 import com.sk89q.worldedit.function.mask.Mask;
 import com.sk89q.worldedit.function.operation.Operations;
 import com.sk89q.worldedit.function.visitor.RegionVisitor;
 import com.sk89q.worldedit.internal.command.CommandRegistrationHandler;
 import com.sk89q.worldedit.internal.command.CommandUtil;
-import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.util.SideEffect;
 import com.sk89q.worldedit.util.SideEffectSet;
 import com.sk89q.worldedit.util.auth.AuthorizationException;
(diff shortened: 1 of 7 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -282,6 +287,33 @@ void perf(Actor actor, LocalSession session,
         }
     }
 
+    @Command(
+        name = "/update",
+        desc = "Apply a side effect"
+    )
+    @CommandPermissions("worldedit.update")
+    void update(Actor actor, LocalSession session, World injectedWorld,
+                   @Arg(desc = "The side effect", def = "")
```
octylFractal, line 296: `@Arg` should align with `(` in `update(`

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -282,6 +287,33 @@ void perf(Actor actor, LocalSession session,
...
+                       SideEffect sideEffect) throws WorldEditException {
```
octylFractal, line 297: Can we take multiple side-effects here? Seems like a nicer option.

worldedit-core/src/main/java/com/sk89q/worldedit/util/SideEffect.java
```diff
@@ -29,6 +29,7 @@
     ENTITY_AI(State.OFF),
     EVENTS(State.OFF);
 
+    // TODO Make these components in WE8
```
octylFractal, line 32: Why not just do this now (maybe in separate PR, remove comments here)? It'd be easy to deprecate out.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -282,6 +287,29 @@ void perf(Actor actor, LocalSession session,
         }
     }
 
+    @Command(
+        name = "/update",
+        desc = "Apply a side effect"
```
octylFractal, line 292:
```suggestion
        desc = "Apply side effects to your selection"
```

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/CommaSeparatedValuesConverter.java
```diff
@@ -73,7 +86,17 @@ public Component describeAcceptableArguments() {
     public List<String> getSuggestions(String input, InjectedValueAccess context) {
         String lastInput = Iterables.getLast(COMMA.split(input), "");
         assert lastInput != null;
-        return delegate.getSuggestions(lastInput, context);
+        List<String> suggestions = delegate.getSuggestions(lastInput, context);
+        if (input.length() > lastInput.length()) {
```
octylFractal, line 90: Can we make this condition `input.contains(",")` instead? (Not sure if that's valid Java tbh, but you know what I mean)

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/SideEffectSetConverter.java
```diff
@@ -0,0 +1,80 @@
...
+import org.enginehub.piston.converter.SuccessfulConversion;
+import org.enginehub.piston.inject.InjectedValueAccess;
+import org.enginehub.piston.inject.Key;
+
+import java.util.List;
+
+public class SideEffectSetConverter implements ArgumentConverter<SideEffectSet> {
+
+    public static void register(CommandManager commandManager) {
+        commandManager.getConverter(Key.of(SideEffect.class)).ifPresent(sideEffectConverter ->
```
octylFractal, line 39: You might want to make this use `orElseThrow` instead, so that an incorrect ordering causes a hard fail rather than errors at command execution time.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/SideEffectSetConverter.java
```diff
@@ -0,0 +1,82 @@
...
+
+public class SideEffectSetConverter implements ArgumentConverter<SideEffectSet> {
+
+    public static void register(CommandManager commandManager) {
+        Optional<ArgumentConverter<SideEffect>> sideEffectConverter = commandManager.getConverter(Key.of(SideEffect.class));
+        commandManager.registerConverter(
+            Key.of(SideEffectSet.class),
+            new SideEffectSetConverter(CommaSeparatedValuesConverter.wrapNoRepeats(sideEffectConverter
+                .orElseThrow(() -> new IllegalStateException("SideEffectSetConverter must be registered after SideEffectConverter"))))
+        );
```
octylFractal:
```suggestion
        ArgumentConverter<SideEffect> sideEffectConverter = commandManager.getConverter(Key.of(SideEffect.class))
           .orElseThrow(() -> new IllegalStateException("SideEffectSetConverter must be registered after SideEffectConverter"));
        commandManager.registerConverter(
            Key.of(SideEffectSet.class),
            new SideEffectSetConverter(CommaSeparatedValuesConverter.wrapNoRepeats(sideEffectConverter))
        );
```

octylFractal (approved):

PR #1632, opened 2020-12-31
Correct ellipsoid region selector initialization

zml2008:
When the previous selection was a non-ellipsoidal region, the
started flag would not be appropriately set until a new selection
had been started.

## Reproduction

1. Make a cuboid selection (or any other type really)
2. Change to an ellipsoid with `/sel ellipsoid`
3. Try to modify the selection with a right-click.

Without this fix, step 3 will fail, requiring starting a whole new selection to make modifications.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/EllipsoidRegionSelector.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/EllipsoidRegionSelector.java
@@ -79,6 +79,7 @@ public class EllipsoidRegionSelector implements RegionSelector, CUIRegion {
             final EllipsoidRegionSelector ellipsoidRegionSelector = (EllipsoidRegionSelector) oldSelector;
 
             region = new EllipsoidRegion(ellipsoidRegionSelector.getIncompleteRegion());
+            started = ellipsoidRegionSelector.started;
         } else {
             Region oldRegion;
             try {
@@ -93,6 +94,7 @@ public class EllipsoidRegionSelector implements RegionSelector, CUIRegion {
             BlockVector3 center = pos1.add(pos2).divide(2).floor();
             region.setCenter(center);
             region.setRadius(pos2.subtract(center).toVector3());
+            started = true;
         }
     }
 


worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/EllipsoidRegionSelector.java
```diff
@@ -208,6 +209,7 @@ public EllipsoidRegion getIncompleteRegion() {
 
     @Override
     public void learnChanges() {
+        this.started = !region.getCenter().equals(Vector3.ZERO);
```
me4502, line 212: Would this not prevent creating an ellipsoid around the zero point?

zml2008: hm, I suppose -- I'd taken that value since it was the default value for `center` from `clear`, but I suppose directly setting the started flag to `true` in the constructor would work as well

octylFractal: Yea, I'd rather not use magic values that are in-bounds.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/EllipsoidRegionSelector.java
```diff
@@ -93,6 +93,7 @@ public EllipsoidRegionSelector(RegionSelector oldSelector) {
             BlockVector3 center = pos1.add(pos2).divide(2).floor();
             region.setCenter(center);
             region.setRadius(pos2.subtract(center).toVector3());
+            started = true;
```
octylFractal, line 96: Should this flag also be inherited in the other branch of the `if`?

zml2008: Ah yes - applied there.

octylFractal (approved):

PR #1633, opened 2020-12-31
Alter the CUI lifecycle to be more consistent and reliable

zml2008:
This handling reduces the platform-specific behaviour of CUI handshakes, and ensures that the appropriate state is maintained when a client connection ends but a LocalSession remains cached.

The underlying issue here is that the lifetime of a LocalSession is longer than the lifetime of any single client connection, and if the session thinks it already has CUI support enabled it wouldn't fully re-initialize the selection. 

This issue can be reproduced on either a dedicated server, or as a player on an existing LAN world, by creating a non-cuboid selection, logging out and logging back in. The selection will no longer be visible, and the CUI will have logged some errors.

If we assume that once a player has connected from a client with CUI support once they will always continue connecting from a client with CUI support the change is a bit more straightforward -- just resend a full selection initialization every time the client sends a CUI handshake packet. However, it is entirely possible for a player to connect from a modded client, quit, and then shortly after connect from a vanilla client. When that happens, we don't want to send selection data to a client that will, at best, ignore it. To do that, player quits have to be tracked in order to clear out selection data as soon as a player disconnects.

## Further work

While this PR will make things work for now, there is potential to improve the handshake from the client's side. Currently the client just sends the handshake packet without waiting for a `minecraft:register`packet from the server, and resets when it detects a "world change". Every time the client sends a handshake, it resets its state, expecting a full re-initialization from the server. I'd like to make this more reliant on the `minecraft:register` channel, but would need to test behaviour across proxies, multiworld, and the various platforms WE supports.

This fixes mikroskeem/WorldEditCUI#13

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
@@ -78,12 +78,17 @@ import static com.google.common.base.Preconditions.checkNotNull;
  */
 public class LocalSession {
 
+    private static final transient int CUI_VERSION_UNINITIALIZED = -1;
     public static transient int MAX_HISTORY_SIZE = 15;
 
     // Non-session related fields
     private transient LocalConfiguration config;
     private final transient AtomicBoolean dirty = new AtomicBoolean();
     private transient int failedCuiAttempts = 0;
+
+    // Single-connection lifetime fields
+    private transient boolean hasCUISupport = false;
+    private transient int cuiVersion = CUI_VERSION_UNINITIALIZED;
 
     // Session related
     private transient RegionSelector selector = new CuboidRegionSelector();
@@ -99,8 +104,6 @@ public class LocalSession {
     private transient boolean useInventory;
     private transient com.sk89q.worldedit.world.snapshot.Snapshot snapshot;
     private transient Snapshot snapshotExperimental;
-    private transient boolean hasCUISupport = false;
-    private transient int cuiVersion = -1;
     private transient SideEffectSet sideEffectSet = SideEffectSet.defaults();
     private transient Mask mask;
     private transient ZoneId timezone = ZoneId.systemDefault();
@@ -907,7 +910,12 @@ public class LocalSession {
      */
     public void handleCUIInitializationMessage(String text, Actor actor) {
         checkNotNull(text);
-        if (this.hasCUISupport || this.failedCuiAttempts > 3) {
+        if (this.hasCUISupport) {
+            // WECUI is a bit aggressive about re-initializing itself
+            // the last attempt to touch handshakes didn't go well, so this will do... for now
+            dispatchCUISelection(actor);
+            return;
+        } else if (this.failedCuiAttempts > 3) {
             return;
         }
 
@@ -1163,5 +1171,16 @@ public class LocalSession {
      */
     public void setLastDistribution(List<Countable<BlockState>> dist) {
         lastDistribution = dist;
+    }
+
+    /**
+     * Call when this session has become inactive.
+     *
+     * <p>This is for internal use only.</p>
+     */
+    public void didBecomeIdle() {
+        this.cuiVersion = CUI_VERSION_UNINITIALIZED;
+        this.hasCUISupport = false;
+        this.failedCuiAttempts = 0;
     }
 }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
@@ -84,9 +84,9 @@ public class LocalSession {
     // Non-session related fields
     private transient LocalConfiguration config;
     private final transient AtomicBoolean dirty = new AtomicBoolean();
-    private transient int failedCuiAttempts = 0;
 
     // Single-connection lifetime fields
+    private transient int failedCuiAttempts = 0;
     private transient boolean hasCUISupport = false;
     private transient int cuiVersion = CUI_VERSION_UNINITIALIZED;
 
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
@@ -973,6 +973,10 @@ public class LocalSession {
      * @param cuiVersion the CUI version
      */
     public void setCUIVersion(int cuiVersion) {
+        if (cuiVersion < 0) {
+            throw new IllegalArgumentException("CUI protocol version must be non-negative, but '" + cuiVersion + "' was received.");
+        }
+
         this.cuiVersion = cuiVersion;
     }
 
@@ -1178,7 +1182,7 @@ public class LocalSession {
      *
      * <p>This is for internal use only.</p>
      */
-    public void didBecomeIdle() {
+    public void onIdle() {
         this.cuiVersion = CUI_VERSION_UNINITIALIZED;
         this.hasCUISupport = false;
         this.failedCuiAttempts = 0;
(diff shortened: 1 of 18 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
```diff
@@ -78,13 +78,18 @@
...
     public static transient int MAX_HISTORY_SIZE = 15;
 
     // Non-session related fields
     private transient LocalConfiguration config;
     private final transient AtomicBoolean dirty = new AtomicBoolean();
     private transient int failedCuiAttempts = 0;
 
+    // Single-connection lifetime fields
+    private transient boolean hasCUISupport = false;
+    private transient int cuiVersion = CUI_VERSION_UNINITIALIZED;
```
octylFractal, line 91: Can you validate that this is not set to `< 0` in `setCUIVersion`?

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
```diff
@@ -1164,4 +1172,15 @@ public boolean isNavWandItemDefault() {
     public void setLastDistribution(List<Countable<BlockState>> dist) {
         lastDistribution = dist;
     }
+
+    /**
+     * Call when this session has become inactive.
+     *
+     * <p>This is for internal use only.</p>
+     */
+    public void didBecomeIdle() {
```
octylFractal, line 1181: I would rather call this:
```suggestion
    public void onIdle() {
```

worldedit-core/src/main/java/com/sk89q/worldedit/event/platform/SessionIdleEvent.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.sk89q.worldedit.event.platform;
+
+import com.sk89q.worldedit.event.Event;
+import com.sk89q.worldedit.session.SessionKey;
+
+/**
+ * An event called when a session becomes idle.
+ *
+ * <p>This can happen when a player leaves the server.
```
octylFractal, line 28:
```suggestion
 * <p>This can happen when a player leaves the server.</p>
```

worldedit-core/src/main/java/com/sk89q/worldedit/event/platform/SessionIdleEvent.java
```diff
@@ -0,0 +1,45 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.event.platform;
+
+import com.sk89q.worldedit.event.Event;
+import com.sk89q.worldedit.session.SessionKey;
+
+/**
+ * An event called when a session becomes idle.
```
octylFractal, line 26:
```suggestion
 * An event fired when a session becomes idle.
```

worldedit-core/src/main/java/com/sk89q/worldedit/session/SessionManager.java
```diff
@@ -356,13 +360,26 @@ public void onConfigurationLoad(ConfigurationLoadEvent event) {
...
+    }
+
     /**
      * Stores the owner of a session, the session, and the last active time.
      */
     private static final class SessionHolder {
         private final SessionKey key;
         private final LocalSession session;
         private long lastActive = System.currentTimeMillis();
+        private boolean hasBecomeIdle = false;
```
octylFractal, line 382:
```suggestion
        private boolean sessionIdle = false;
```

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricPlayer.java
```diff
@@ -128,7 +128,7 @@ public void dispatchCUIEvent(CUIEvent event) {
         if (params.length > 0) {
             send = send + "|" + StringUtil.joinString(params, "|");
         }
-        PacketByteBuf buffer = new PacketByteBuf(Unpooled.copiedBuffer(send.getBytes(WECUIPacketHandler.UTF_8_CHARSET)));
+        PacketByteBuf buffer = new PacketByteBuf(Unpooled.wrappedBuffer(send.getBytes(StandardCharsets.UTF_8)));
```
octylFractal, line 131:
```suggestion
        PacketByteBuf buffer = new PacketByteBuf(Unpooled.copiedBuffer(send, StandardCharsets.UTF_8));
```

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
```diff
@@ -284,6 +284,12 @@ private ActionResult onRightClickBlock(PlayerEntity playerEntity, World world, H
 
     // TODO Pass empty left click to server
 
+    // TODO: Use the fabric-networking-api-v1 event when updating
```
octylFractal, line 287: Would you mind just updating to v1 for us? Or I can do it separately.

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgePlayer.java
```diff
@@ -128,7 +129,7 @@ public void dispatchCUIEvent(CUIEvent event) {
         if (params.length > 0) {
             send = send + "|" + StringUtil.joinString(params, "|");
         }
-        PacketBuffer buffer = new PacketBuffer(Unpooled.copiedBuffer(send.getBytes(WECUIPacketHandler.UTF_8_CHARSET)));
+        PacketBuffer buffer = new PacketBuffer(Unpooled.wrappedBuffer(send.getBytes(StandardCharsets.UTF_8)));
```
octylFractal, line 132:
```suggestion
        PacketBuffer buffer = new PacketBuffer(Unpooled.copiedBuffer(send, StandardCharsets.UTF_8));
```

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorldEdit.java
```diff
@@ -293,6 +295,14 @@ public void onCommandEvent(CommandEvent event) throws CommandSyntaxException {
         ));
     }
 
+    @SubscribeEvent
+    public void onPlayerLogOut(PlayerEvent.PlayerLoggedOutEvent event) {
+        if (event.getPlayer() instanceof ServerPlayerEntity) {
```
octylFractal, line 300: every day I pray for mojang to move to higher Java versions so we may `instanceof ServerPlayerEntity serverPlayer`.

octylFractal (changes requested):

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorldEdit.java
zml2008: soon:tm:

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
zml2008: Sure, I've updated to networking-v1

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
```diff
@@ -78,13 +78,18 @@
  */
 public class LocalSession {
 
+    private static final transient int CUI_VERSION_UNINITIALIZED = -1;
     public static transient int MAX_HISTORY_SIZE = 15;
 
     // Non-session related fields
     private transient LocalConfiguration config;
     private final transient AtomicBoolean dirty = new AtomicBoolean();
     private transient int failedCuiAttempts = 0;
```
octylFractal, line 87: This should be moved under `Single-connection lifetime fields` as well.

octylFractal (changes requested):

octylFractal (approved):

wizjany (approved):

me4502 (approved):

PR #1638, opened 2021-01-05
Support Lucko's Fabric Permissions API

octylFractal:
Fairly straightforward, falls back to the existing provider if the mod isn't available or the returned perm is `false`.

### DIFF
--- a/worldedit-fabric/build.gradle.kts
+++ b/worldedit-fabric/build.gradle.kts
@@ -99,6 +99,9 @@ dependencies {
         "modImplementation"(it)
     }
 
+    // No need for this at runtime
+    "modCompileOnly"("me.lucko:fabric-permissions-api:0.1-SNAPSHOT")
+
     // Hook these up manually, because Fabric doesn't seem to quite do it properly.
     "compileOnly"("net.fabricmc:sponge-mixin:${project.versions.mixin}")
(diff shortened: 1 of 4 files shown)


me4502 (approved):

octylFractal:
also gonna @lucko on here to make sure they see this.

lucko:
Hey! Looks good :)

The only thing I'd say is - currently, it wouldn't be possible for admins to negate WorldEdit permissions (set them to `false`) because that would be overridden by the vanilla provider.

To fix this, you could consider replacing
```java
return Permissions.check(player, permission) || super.hasPermission(player, permission);
```

with
```java
return Permissions.check(player, permission, super.hasPermission(player, permission));
```

or with
```java
return Permissions.getPermissionValue(player, permission)
    .orElseGet(() -> super.hasPermission(player, permission));
```

wizjany:
While maybe not ideal, I'd note that that's in-line with behavior on other platforms, e.g. bukkit where a user would have to specifically have to turn off op-perms in config to bypass that. That's kind of just a relic of before perms systems existed/were standardized to any extent, and if we did have a good crossplatform perms system (at least standardized behavior if not API), I'd like to see that dropped from bukkit as well.

octylFractal:
I decided to change it to your suggestion, as that's what I actually wanted to do in the first place but I misread it as package-private.

PR #1657, opened 2021-01-29
Add splatter brush 

me4502:
Continuation of https://github.com/EngineHub/WorldEdit/pull/1547

Uses a slightly different algorithm (and therefore different output, this has more of a dropoff effect with high decay rather than splotches), but moves the concept into a mask for more WorldEdit-y usage


worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -134,6 +134,7 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
```
octylFractal, line 137: why

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -175,6 +176,36 @@ public void cylinderBrush(Player player, LocalSession session,
...
+            return;
+        }
+
+        BrushTool tool = session.getBrushTool(player.getItemInHand(HandSide.MAIN_HAND).getType());
+        tool.setFill(pattern);
+        tool.setSize(radius);
+
+        tool.setBrush(new SplatterBrush(decay), "worldedit.brush.splatter");
+
+        player.printInfo(TranslatableComponent.of("worldedit.brush.splatter.equip", TextComponent.of((int) radius), TextComponent.of(decay)));
```
octylFractal, line 205: Why aren't we passing down the double radius directly?

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/SplatterMask.java
```diff
@@ -0,0 +1,59 @@
...
+
+        double decayChance = distRatio * decay * 2;
+
+        return ThreadLocalRandom.current().nextDouble() > decayChance;
+    }
+
+    @Nullable
+    @Override
+    public Mask2D toMask2D() {
+        return null;
```
octylFractal, line 57: Please do implement a Mask2D variant.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
me4502: All of them do this - I'd rather keep this as-is for now.

octylFractal (approved):

PR #1682, opened 2021-02-21
Show a message with install instructions when running the jar file

me4502:
Should lessen the number of people trying to run the jar file and coming to support out of confusion.


GMpet:
I would like to see more plugin devs do things like this. 
Great job

wizjany (approved):

PR #1707, opened 2021-03-18
Allow plugins to register new clipboard share destinations

me4502:
Currently, the share command is rigid and only supports the EngineHub paste service. It also allows a format arg that will break for anything non-Sponge.

This change allows plugins to register their own destinations, as well as what formats each destination can support.

Some potential cool use cases:
* Uploading straight to the many schematic sharing sites
* Uploading to other servers on a network
* Uploading to a general file storage drive like Dropbox or Google Drive

This PR also adds completions for clipboard formats and clipboard destinations.

Due to a bug in Piston, this appears to behave oddly when after generic String args.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
@@ -427,11 +427,12 @@ public class SchematicCommands {
         @Override
         public URL call() throws Exception {
             ClipboardShareMetadata metadata = new ClipboardShareMetadata(
+                format,
                 this.actor.getName(),
                 name == null ? actor.getName() + "-" + System.currentTimeMillis() : name
             );
 
-            return destination.share(format, metadata, this::writeToOutputStream).toURL();
+            return destination.share(metadata, this::writeToOutputStream).toURL();
         }
     }
 
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
@@ -33,11 +33,13 @@ import com.sk89q.worldedit.extension.platform.Actor;
 import com.sk89q.worldedit.extension.platform.Capability;
 import com.sk89q.worldedit.extent.clipboard.BlockArrayClipboard;
 import com.sk89q.worldedit.extent.clipboard.Clipboard;
+import com.sk89q.worldedit.extent.clipboard.FlattenedClipboardTransform;
 import com.sk89q.worldedit.extent.clipboard.io.BuiltInClipboardFormat;
 import com.sk89q.worldedit.extent.clipboard.io.ClipboardFormat;
 import com.sk89q.worldedit.extent.clipboard.io.ClipboardFormats;
 import com.sk89q.worldedit.extent.clipboard.io.ClipboardReader;
 import com.sk89q.worldedit.extent.clipboard.io.ClipboardWriter;
+import com.sk89q.worldedit.extent.clipboard.io.share.ClipboardShareDestination;
 import com.sk89q.worldedit.function.operation.Operations;
 import com.sk89q.worldedit.internal.util.LogManagerCompat;
 import com.sk89q.worldedit.math.transform.Transform;
@@ -55,7 +57,6 @@ import com.sk89q.worldedit.util.formatting.text.format.TextColor;
 import com.sk89q.worldedit.util.io.Closer;
 import com.sk89q.worldedit.util.io.file.FilenameException;
 import com.sk89q.worldedit.util.io.file.MorePaths;
-import com.sk89q.worldedit.util.paste.EngineHubPaste;
 import com.sk89q.worldedit.util.paste.PasteMetadata;
 import org.apache.logging.log4j.Logger;
 import org.enginehub.piston.annotation.Command;
@@ -68,19 +69,16 @@ import org.enginehub.piston.exception.StopExecutionException;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Base64;
 import java.util.Comparator;
 import java.util.List;
 import java.util.concurrent.Callable;
@@ -116,7 +114,7 @@ public class SchematicCommands {
                      @Arg(desc = "File name.")
                          String filename,
                      @Arg(desc = "Format name.", def = "sponge")
-                         String formatName) throws FilenameException {
+                         ClipboardFormat format) throws FilenameException {
         LocalConfiguration config = worldEdit.getConfiguration();
 
         File dir = worldEdit.getWorkingDirectoryPath(config.saveDir).toFile();
@@ -129,13 +127,9 @@ public class SchematicCommands {
             return;
         }
 
-        ClipboardFormat format = ClipboardFormats.findByFile(f);
-        if (format == null) {
-            format = ClipboardFormats.findByAlias(formatName);
-        }
-        if (format == null) {
-            actor.printError(TranslatableComponent.of("worldedit.schematic.unknown-format", TextComponent.of(formatName)));
-            return;
+        ClipboardFormat inferredFormat = ClipboardFormats.findByFile(f);
+        if (inferredFormat != null) {
+            format = inferredFormat;
         }
 
         SchematicLoadTask task = new SchematicLoadTask(actor, f, format);
@@ -160,7 +154,7 @@ public class SchematicCommands {
                      @Arg(desc = "File name.")
                          String filename,
                      @Arg(desc = "Format name.", def = "sponge")
-                         String formatName,
+                         ClipboardFormat format,
                      @Switch(name = 'f', desc = "Overwrite an existing file.")
                          boolean allowOverwrite) throws WorldEditException {
         if (worldEdit.getPlatformManager().queryCapability(Capability.GAME_HOOKS).getDataVersion() == -1) {
@@ -171,12 +165,6 @@ public class SchematicCommands {
         LocalConfiguration config = worldEdit.getConfiguration();
 
         File dir = worldEdit.getWorkingDirectoryPath(config.saveDir).toFile();
-
-        ClipboardFormat format = ClipboardFormats.findByAlias(formatName);
-        if (format == null) {
-            actor.printError(TranslatableComponent.of("worldedit.schematic.unknown-format", TextComponent.of(formatName)));
-            return;
-        }
 
         File f = worldEdit.getSafeSaveFile(actor, dir, filename, format.getPrimaryFileExtension());
 
@@ -220,22 +208,31 @@ public class SchematicCommands {
     public void share(Actor actor, LocalSession session,
                       @Arg(desc = "Schematic name. Defaults to name-millis", def = "")
                           String schematicName,
-                      @Arg(desc = "Format name.", def = "sponge")
-                          String formatName) throws WorldEditException {
+                      @Arg(desc = "Share location", def = "enginehub")
+                          ClipboardShareDestination destination,
+                      @Arg(desc = "Format name", def = "")
+                          ClipboardFormat format) throws WorldEditException {
         if (worldEdit.getPlatformManager().queryCapability(Capability.GAME_HOOKS).getDataVersion() == -1) {
             actor.printError(TranslatableComponent.of("worldedit.schematic.unsupported-minecraft-version"));
             return;
         }
 
-        ClipboardFormat format = ClipboardFormats.findByAlias(formatName);
         if (format == null) {
-            actor.printError(TranslatableComponent.of("worldedit.schematic.unknown-format", TextComponent.of(formatName)));
+            format = destination.getDefaultFormat();
+        }
+
+        if (!destination.supportsFormat(format)) {
+            actor.printError(TranslatableComponent.of(
+                "worldedit.schematic.share.unsupported-format",
+                TextComponent.of(destination.getName()),
+                TextComponent.of(format.getName())
+            ));
             return;
         }
 
         ClipboardHolder holder = session.getClipboard();
 
-        SchematicShareTask task = new SchematicShareTask(actor, format, holder, schematicName);
+        SchematicShareTask task = new SchematicShareTask(actor, holder, destination, format, schematicName);
         AsyncCommandBuilder.wrap(task, actor)
             .registerWithSupervisor(worldEdit.getSupervisor(), "Sharing schematic")
             .setDelayMessage(TranslatableComponent.of("worldedit.schematic.save.saving"))
@@ -366,18 +363,22 @@ public class SchematicCommands {
         }
     }
 
-    private abstract static class SchematicOutputTask<T> implements Callable<T> {
-        protected final Actor actor;
+    private static class SchematicSaveTask implements Callable<Void> {
+        private final Actor actor;
         private final ClipboardFormat format;
         private final ClipboardHolder holder;
+        private final File file;
+        private final boolean overwrite;
 
-        SchematicOutputTask(Actor actor, ClipboardFormat format, ClipboardHolder holder) {
+        SchematicSaveTask(Actor actor, File file, ClipboardFormat format, ClipboardHolder holder, boolean overwrite) {
             this.actor = actor;
             this.format = format;
             this.holder = holder;
+            this.file = file;
+            this.overwrite = overwrite;
         }
 
-        protected void writeToOutputStream(OutputStream outputStream) throws Exception {
+        private void writeToOutputStream(OutputStream outputStream) throws Exception {
             Clipboard clipboard = holder.getClipboard();
             Transform transform = holder.getTransform();
             Clipboard target;
@@ -399,17 +400,6 @@ public class SchematicCommands {
                 writer.write(target);
             }
         }
-    }
-
-    private static class SchematicSaveTask extends SchematicOutputTask<Void> {
-        private final File file;
-        private final boolean overwrite;
-
-        SchematicSaveTask(Actor actor, File file, ClipboardFormat format, ClipboardHolder holder, boolean overwrite) {
-            super(actor, format, holder);
-            this.file = file;
-            this.overwrite = overwrite;
-        }
 
         @Override
         public Void call() throws Exception {
@@ -424,29 +414,33 @@ public class SchematicCommands {
         }
     }
 
-    private static class SchematicShareTask extends SchematicOutputTask<URL> {
+    private static class SchematicShareTask implements Callable<URL> {
+        private final Actor actor;
+        private final ClipboardHolder holder;
         private final String name;
+        private final ClipboardShareDestination destination;
+        private final ClipboardFormat format;
 
-        SchematicShareTask(Actor actor, ClipboardFormat format, ClipboardHolder holder, String name) {
-            super(actor, format, holder);
+        SchematicShareTask(Actor actor,
+                           ClipboardHolder holder,
+                           ClipboardShareDestination destination,
+                           ClipboardFormat format,
+                           String name) {
+            this.actor = actor;
+            this.holder = holder;
             this.name = name;
+            this.destination = destination;
+            this.format = format;
         }
 
         @Override
         public URL call() throws Exception {
-            ByteArrayOutputStream baos = new ByteArrayOutputStream();
-            try {
-                writeToOutputStream(baos);
-            } catch (Exception e) {
-                throw new CommandException(TextComponent.of(e.getMessage()), e, ImmutableList.of());
-            }
-
-            EngineHubPaste pasteService = new EngineHubPaste();
             PasteMetadata metadata = new PasteMetadata();
             metadata.author = this.actor.getName();
             metadata.extension = "schem";
             metadata.name = name == null ? actor.getName() + "-" + System.currentTimeMillis() : name;
-            return pasteService.paste(new String(Base64.getEncoder().encode(baos.toByteArray()), StandardCharsets.UTF_8), metadata).call();
+
+            return destination.share(holder, format, metadata);
         }
     }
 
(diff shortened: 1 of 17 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/FlattenedClipboardTransform.java
```diff
@@ -41,7 +40,7 @@
  * @see Clipboard
  * @see Transform
  */
-class FlattenedClipboardTransform {
+public class FlattenedClipboardTransform {
```
octylFractal, line 43: Rename to something better maybe, like `ClipboardTransformFuser`

octylFractal (changes requested):
more incoming when I get on my computer

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -220,22 +208,31 @@ public void save(Actor actor, LocalSession session,
     public void share(Actor actor, LocalSession session,
                       @Arg(desc = "Schematic name. Defaults to name-millis", def = "")
                           String schematicName,
-                      @Arg(desc = "Format name.", def = "sponge")
-                          String formatName) throws WorldEditException {
+                      @Arg(desc = "Share location", def = "enginehub")
```
wizjany, line 211: this doesn't strictly match a destination, should probably be "eh_paste" or whatever

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -424,29 +414,33 @@ public Void call() throws Exception {
...
-            try {
-                writeToOutputStream(baos);
-            } catch (Exception e) {
-                throw new CommandException(TextComponent.of(e.getMessage()), e, ImmutableList.of());
-            }
-
-            EngineHubPaste pasteService = new EngineHubPaste();
             PasteMetadata metadata = new PasteMetadata();
             metadata.author = this.actor.getName();
             metadata.extension = "schem";
```
wizjany, line 440: should this be based on the format field instead? is our PasteMetadata really something that should be generically exposed to the ClipboardShareDestination class? seems more specific to EH paste.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/ClipboardTransformFuser.java
```diff
@@ -41,7 +40,7 @@
  * @see Clipboard
  * @see Transform
  */
-class FlattenedClipboardTransform {
+public class ClipboardTransformFuser {
```
wizjany, line 43: This class may need additional rethinking. It definitely wasn't meant to be exposed in the current state.
Aside from the name (bake pls), it would probably be more useful to API users (including WE internally) to just provide the resultant Clipboard object instead of just exposing copyTo. This would also prevent it from being misused as a pastebuilder as a lot of clipboard manipulation can (and should) be done off main thread, and pre-baking should not be done at paste-time.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ClipboardShareDestinations.java
```diff
@@ -0,0 +1,79 @@
...
+public class ClipboardShareDestinations {
+
+    private static final Map<String, ClipboardShareDestination> aliasMap = new HashMap<>();
+    private static final List<ClipboardShareDestination> registeredDestinations = new ArrayList<>();
+
+    public static void registerClipboardShareDestination(ClipboardShareDestination destination) {
+        checkNotNull(destination);
+
+        for (String key : destination.getAliases()) {
+            String lowKey = key.toLowerCase(Locale.ENGLISH);
```
wizjany, line 44: weren't we using ROOT throughout? is there a reason to use ENGLISH here? (same applies to the clipboardformats class this was copied from - not sure why it uses english either. also one usage in GeneralCommands for some reason. throughout the repo there's 38 ROOTS and 2 ENGLISH (+1 english for the default i18n locale which makes sense))

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ClipboardShareDestinations.java
```diff
@@ -0,0 +1,79 @@
...
+     * Find the clipboard format named by the given alias.
+     *
+     * @param alias
+     *            the alias
+     * @return the format, otherwise null if none is matched
+     */
+    @Nullable
+    public static ClipboardShareDestination findByAlias(String alias) {
+        checkNotNull(alias);
+        return aliasMap.get(alias.toLowerCase(Locale.ENGLISH).trim());
```
wizjany, line 70: as above, I think we should prefer ROOT.

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -107,6 +107,7 @@
     "worldedit.schematic.save.failed-directory": "Could not create folder for schematics!",
     "worldedit.schematic.save.saving": "(Please wait... saving schematic.)",
     "worldedit.schematic.save.still-saving": "(Please wait... still saving schematic.)",
+    "worldedit.schematic.share.unsupported-format": "The schematic share destination \"{0}\" does not support \"{1}\" schematics.",
```
wizjany, line 110: I think this leads to awkward/redundant messages like `does not support "SPONGE_SCHEMATIC" schematics`. Maybe just `does not support the \"{1}\" format.`?

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ClipboardShareDestinations.java
```diff
@@ -0,0 +1,79 @@
...
+        for (BuiltInClipboardShareDestinations destination : BuiltInClipboardShareDestinations.values()) {
+            registerClipboardShareDestination(destination);
+        }
+    }
+
+    /**
+     * Find the clipboard format named by the given alias.
+     *
+     * @param alias
+     *            the alias
```
wizjany, line 64: awk newline.

wizjany (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
me4502: Hmm, fair - I feel most of this metadata makes sense from a general perspective, but extension maybe less so.

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/ClipboardFormatConverter.java
```diff
@@ -0,0 +1,83 @@
...
+            .filter(format -> format.startsWith(input))
+            .collect(Collectors.toList());
+    }
+
+    @Override
+    public ConversionResult<ClipboardFormat> convert(String s, InjectedValueAccess injectedValueAccess) {
+        ClipboardFormat result = ClipboardFormats.findByAlias(s);
+        return result == null
+            ? FailedConversion.from(new IllegalArgumentException(
+            "Not a valid schematic format: " + s))
```
octylFractal:
```suggestion
            ? FailedConversion.from(new IllegalArgumentException("Not a valid schematic format: " + s))
```

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/ClipboardFormatConverter.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public List<String> getSuggestions(String input, InjectedValueAccess context) {
+        ClipboardShareDestination destination = context.injectedValue(Key.of(ClipboardShareDestination.class)).orElse(null);
+
+        return getFormats()
+            .filter(format -> destination == null || destination.supportsFormat(format))
+            .map(ClipboardFormat::getAliases)
+            .flatMap(Set::stream)
+            .filter(format -> format.startsWith(input))
```
octylFractal, line 71: don't we have a utility for this already?

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/ClipboardFormatConverter.java
```diff
@@ -0,0 +1,83 @@
...
+    private ClipboardFormatConverter() {
+        this.choices = TextComponent.of("any clipboard format");
+    }
+
+    @Override
+    public Component describeAcceptableArguments() {
+        return this.choices;
+    }
+
+    private Stream<ClipboardFormat> getFormats() {
```
octylFractal, line 59: I don't see the point of extracting this method out. What's wrong with calling it inline in the one usage it has?

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/ClipboardShareDestinationConverter.java
```diff
@@ -0,0 +1,79 @@
...
+    private ClipboardShareDestinationConverter() {
+        this.choices = TextComponent.of("any clipboard share destination");
+    }
+
+    @Override
+    public Component describeAcceptableArguments() {
+        return this.choices;
+    }
+
+    private Stream<ClipboardShareDestination> getDestinations() {
```
octylFractal, line 58: Again, not sure why this was factored out. Just inline it.

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/ClipboardShareDestinationConverter.java
```diff
@@ -0,0 +1,79 @@
...
+            .filter(destination -> destination.startsWith(input))
+            .collect(Collectors.toList());
+    }
+
+    @Override
+    public ConversionResult<ClipboardShareDestination> convert(String s, InjectedValueAccess injectedValueAccess) {
+        ClipboardShareDestination result = ClipboardShareDestinations.findByAlias(s);
+        return result == null
+            ? FailedConversion.from(new IllegalArgumentException(
+            "Not a valid clipboard share destination: " + s))
```
octylFractal:
```suggestion
            ? FailedConversion.from(new IllegalArgumentException("Not a valid clipboard share destination: " + s))
```

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/ClipboardTransformBaker.java
```diff
@@ -36,12 +37,10 @@
 /**
  * Helper class to 'bake' a transform into a clipboard.
  *
- * <p>This class needs a better name and may need to be made more generic.</p>
- *
  * @see Clipboard
  * @see Transform
  */
-class FlattenedClipboardTransform {
+public class ClipboardTransformBaker {
```
octylFractal, line 43: Move to `internal.util`, add `Clipboard#transform(Transform);Clipboard` which delegates to `bakeTransform`.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/BuiltInClipboardShareDestinations.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * A collection of natively supported clipboard share destinations.
+ */
+public enum BuiltInClipboardShareDestinations implements ClipboardShareDestination {
+
+    /**
+     * The EngineHub pastebin service, at https://paste.enginehub.org/
+     */
+    ENGINEHUB_PASTEBIN("enginehub_paste", "ehpaste") {
```
octylFractal, line 49: I think I would seriously prefer only having this as `enginehub:paste`, and maybe allow implicitly dropping the `enginehub` since we're the primary source here, like we do with `minecraft:` on blocks and items.

So, singular `enginehub:paste` identifier, parsing from user input can implicitly add `enginehub:`.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/BuiltInClipboardShareDestinations.java
```diff
@@ -0,0 +1,110 @@
...
+            Clipboard clipboard = holder.getClipboard();
+            Transform transform = holder.getTransform();
+            Clipboard target;
+
+            // If we have a transform, bake it into the copy
+            if (transform.isIdentity()) {
+                target = clipboard;
+            } else {
+                target = ClipboardTransformBaker.bakeTransform(clipboard, transform);
+            }
```
octylFractal, line 67: I wonder if we shouldn't move this into `bakeTransform`'s logic. Only issue there is defining that the input value may be returned as the output, and as such, an additional copy may be needed.

Or perhaps we can have an optimized variant, e.g. `transformOptimized`, or a flag.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ClipboardShareDestinations.java
```diff
@@ -0,0 +1,78 @@
...
+
+        for (String key : destination.getAliases()) {
+            String lowKey = key.toLowerCase(Locale.ROOT);
+            ClipboardShareDestination old = aliasMap.put(lowKey, destination);
+            if (old != null) {
+                aliasMap.put(lowKey, old);
+                WorldEdit.logger.warn(destination.getClass().getName() + " cannot override existing alias '" + lowKey + "' used by " + old.getClass().getName());
+            }
+        }
+        registeredDestinations.add(destination);
```
octylFractal, line 51: Can you add a check that the destination accepts its default format?

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ClipboardShareMetadata.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.sk89q.worldedit.extent.clipboard.io.share;
+
+/**
+ * Items of metadata about shared clipboards.
+ */
+public class ClipboardShareMetadata {
+    public String name;
+    public String author;
```
octylFractal, line 27: Please please please no. Private fields, public getters. I know you want this to be a record, but this forbids us from making it one in WE 7. A public getter, OTOH, allows for us to deprecate those getters and delegate to the record entries. Or you could simply style this like a record to begin with, using `name()` and `author()` as the getters.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/ClipboardFormatConverter.java
```diff
@@ -0,0 +1,82 @@
...
+    public static void register(CommandManager commandManager) {
+        commandManager.registerConverter(Key.of(ClipboardFormat.class),
+            new ClipboardFormatConverter()
+        );
+    }
+
+    private final TextComponent choices;
+
+    private ClipboardFormatConverter() {
+        this.choices = TextComponent.of("any clipboard format");
```
kashike, line 51: translatable?

worldedit-core/src/main/java/com/sk89q/worldedit/command/argument/ClipboardShareDestinationConverter.java
```diff
@@ -0,0 +1,78 @@
...
+    public static void register(CommandManager commandManager) {
+        commandManager.registerConverter(Key.of(ClipboardShareDestination.class),
+            new ClipboardShareDestinationConverter()
+        );
+    }
+
+    private final TextComponent choices;
+
+    private ClipboardShareDestinationConverter() {
+        this.choices = TextComponent.of("any clipboard share destination");
```
kashike, line 50: translatable?

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/Clipboard.java
```diff
@@ -68,4 +71,19 @@
...
+    /**
+     * Returns a clipboard with a given transform baked in.
+     *
+     * <p>
+     * Note: If an identity transform is given, the original clipboard will be returned unedited.
+     * </p>
+     *
+     * @param transform The transform
+     * @return The new clipboard
+     * @throws WorldEditException If it failed to copy
```
octylFractal, line 84:
```suggestion
     * @throws WorldEditException if the copy encounters an error
```

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ClipboardShareDestination.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Share a clipboard output stream and return a URL.
+     *
+     * @param holder The clipboard holder
+     * @param format The clipboard format
+     * @param metadata The clipboard metadata
+     * @return The URL
+     * @throws Exception if it failed to share
+     */
+    URL share(ClipboardHolder holder, ClipboardFormat format, ClipboardShareMetadata metadata) throws Exception;
```
octylFractal, line 53: Consider changing to:
```suggestion
   URI share(ClipboardFormat format, ClipboardShareMetadata metadata, IOConsumer<OutputStream> serialize) throws Exception;
```
`URI` works better as the return value for full compatibility with all destinations, and `serialize` allows us to own the serialization code and implementers don't need to worry about it.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/Clipboard.java
```diff
@@ -68,4 +71,19 @@
     default boolean hasBiomes() {
         return false;
     }
+
+    /**
+     * Returns a clipboard with a given transform baked in.
+     *
+     * <p>
+     * Note: If an identity transform is given, the original clipboard will be returned unedited.
```
octylFractal, line 79: I'd like to be less specific about the exact implementation, and just say this instead:
```suggestion
     * Note: This method may return the same clipboard object, if a copy is needed then you should check the returned value for identity equality and copy if needed.
```

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ClipboardShareDestination.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Get a set of aliases.
+     *
+     * @return a set of aliases
+     */
+    Set<String> getAliases();
+
+    /**
+     * Share a clipboard output stream and return a URL.
```
octylFractal, line 44:
```suggestion
     * Share a clipboard output stream and return a URL.
     *
     * The serialized schematic can be retrieved by providing an {@link java.io.OutputStream} to {@code serializer}.
```

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ClipboardShareDestination.java
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @param format The clipboard format
+     * @param metadata The clipboard metadata
+     * @param serialize A consumer taking the {@link java.io.OutputStream}
+     * @return The URI
+     * @throws Exception if it failed to share
+     */
+    URI share(ClipboardFormat format, ClipboardShareMetadata metadata, ShareOutputConsumer serialize) throws Exception;
```
octylFractal, line 52: IMO `format` is part of `metadata`, isn't it?

octylFractal, line 52:
```suggestion
    URI share(ClipboardFormat format, ClipboardShareMetadata metadata, ShareOutputConsumer serializer) throws Exception;
```

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ClipboardShareDestinations.java
```diff
@@ -0,0 +1,80 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ClipboardShareDestinations {
+
+    private static final Map<String, ClipboardShareDestination> aliasMap = new HashMap<>();
+    private static final List<ClipboardShareDestination> registeredDestinations = new ArrayList<>();
+
+    public static void registerClipboardShareDestination(ClipboardShareDestination destination) {
+        checkNotNull(destination);
+        checkArgument(destination.supportsFormat(destination.getDefaultFormat()), "Destination must accept its default format");
```
octylFractal, line 43:
```suggestion
        checkState(destination.supportsFormat(destination.getDefaultFormat()), "Destination must accept its default format");
```

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ShareOutputConsumer.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.sk89q.worldedit.extent.clipboard.io.share;
+
+import com.sk89q.worldedit.WorldEditException;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+@FunctionalInterface
+public interface ShareOutputConsumer {
```
octylFractal, line 28: Consider renaming to `ShareOutputProvider`? Or something? not sure what name is good, I just don't like `Consumer` here.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ShareOutputConsumer.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Consumes the given output stream.
```
octylFractal, line 31: In tune with the above,
```suggestion
     * Provides the share output to {@code stream}.
```

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/ShareOutputConsumer.java
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @throws IOException if it failed
+     * @throws WorldEditException if WorldEdit failed to serialize to the stream
+     */
+    void accept(OutputStream stream) throws IOException, WorldEditException;
```
octylFractal, line 36: In tune with the above,
```suggestion
    void writeTo(OutputStream stream) throws IOException, WorldEditException;
```
(or maybe call it `provideTo`?)

octylFractal (changes requested):

SirYwell:
I'm late to the party, but from looking at the code it seems that the success message is hardcoded with the URL, also adding a `.schem`. That looks a bit odd to me and is a limitation for [Arkitektonika](https://github.com/IntellectualSites/Arkitektonika), where we generate an access url and a separate deletion url. In best case, we'd be able to print both of them to the player. Would that be possible?

me4502:
@SirYwell I believe I've resolved that in my most recent commit. I'm not entirely happy with returning a consumer rather than the URL, but it provides the most flexibility.

SirYwell:
I'd be already totally happy with returning a `TextComponent`, but maybe it's worth to have that much flexibility there. Thanks!

octylFractal (approved):

PR #1715, opened 2021-03-21
Scrapped Logging Framework 4 Java

### LINKED ISSUE #1709: WorldEdit 7.2.4 and 7.2.3 crashes when loading world
### WorldEdit Version

7.2.4+5723-e3bfc64 and 7.2.3+5680-135941e

### Platform Version

Forge 36.1.1 and 36.1.0

### Bug Description

WorldEdit 7.2.4 and 7.2.3 crashes when loading world.
WorldEdit 7.2.2 does not crash.

* Windows 10, updated.
* AdoptOpenJDK Hotspot 1.8 (jdk8u282-b08)
* Updated MultiMC
* Minecraft 1.16.5
* Freshly downloaded WorldEdit .jar-files from curseforge

I am using MultiMC, ~~and have not tested the behaviour with the vanilla launcher as this probably wouldn't make a difference.~~
_EDIT: It turns out that it indeed **is** a conflict when using MultiMC.
But I don't know if it is MultiMC that fails or it is MultiMC that trips over an error induced by WorldEdit._

All combinations of above mentioned configurations have been tested separately on clean Minecraft instances, as to make sure no old data was lingering and no other mods were to blame.

A combination of Forge 36.0.63 and WorldEdit 7.2.4 has also been done and does also crash.

Log for the combination of Forge 36.1.1 and  WorldEdit 7.2.4 is available at:
https://paste.ee/p/uwfKT

### Expected Behavior

Loading world without crash.

### Reproduction Steps

1. Create a new Minecraft 1.16.5. instance
2. Add Forge 36.1.1 or 36.1.0
3. Add WorldEdit 7.2.4 or 7.2.3
4. Launch the Minecraft instance
5. Create a new world or load an existing one.
6. Watch the loading percentage reach 100%
7. Game crashes to desktop.

octylFractal:
SLF4J is basically unmaintained and Log4J 2 provides essentially a replacement in the form of `log4j-api`, so let's swap to that.

Bonuses:
- Less bundling on all platforms! Everyone has Log4J on the classpath at runtime!
- Less overhead! Because Log4J is on the classpath already, there's no translation from SLF4J to Log4J like we have been doing
- Future proof! Log4J is actively maintained, with improvements and bug fixes.

~~Note: this is dependent on my Paper PR (PaperMC/Paper#5400) being merged for WorldEdit to continue to work on Paper in Java 11+. Additionally, it will likely break on Spigot in Java 11+, but I am of the opinion that people should either (a) use Paper, (b) downgrade to 8, or (c) convince Spigot to take on my patch as well.~~ This is no longer true, but should the PR be merged and we end up requiring paper, we can replace `LogManagerCompat` to save a few kb.

Fixes #1709


octylFractal:
Oh, I also need to release Piston before this can be merged. But that should be done only if this is accepted as a path forward.

me4502 (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/LogManagerCompat.java
```diff
@@ -0,0 +1,44 @@
...
+    public static Logger getLogger() {
+        return LogManager.getLogger(getCallerCallerClassName());
+    }
+
+    private static String getCallerCallerClassName() {
+        List<StackTraceElement> lazyStack = Throwables.lazyStackTrace(new Throwable());
+        // 0 - this method
+        // 1 - caller
+        // 2 - caller caller
+        return lazyStack.get(2).getClassName();
```
me4502, line 39: Ngl this feels hella hacky, is it guaranteed to be accurate across all JVM versions?

octylFractal: To my knowledge, it should work everywhere. I don't know why stackframes would differ elsewhere.

worldedit-sponge/build.gradle.kts
```diff
@@ -20,7 +20,13 @@ repositories {
 dependencies {
     api(project(":worldedit-core"))
     api(project(":worldedit-libs:sponge"))
-    api("org.spongepowered:spongeapi:7.1.0")
+    api("org.spongepowered:spongeapi:7.1.0") {
+        exclude(group = "org.slf4j", module = "slf4j-api")
+    }
+    implementation(enforcedPlatform("org.apache.logging.log4j:log4j-bom:2.8.1") {
```
wizjany, line 26:
```suggestion
    implementation(enforcedPlatform("org.apache.logging.log4j:log4j-bom:${Versions.LOG4J}") {
```

worldedit-fabric/build.gradle.kts
```diff
@@ -47,9 +47,8 @@ repositories {
 dependencies {
     "api"(project(":worldedit-core"))
     "implementation"(enforcedPlatform("org.apache.logging.log4j:log4j-bom:2.8.1") {
```
wizjany, line 49:
```suggestion
    "implementation"(enforcedPlatform("org.apache.logging.log4j:log4j-bom:${Versions.LOG4J}") {
```

wizjany (changes requested):
These versions should probably use the Versions constants. Otherwise LGTM.

wizjany (approved):

PR #1723, opened 2021-04-07
Allow toggleplace to work on any Actor + Locatable

octylFractal:
Not sure if we wish to change the `.player` message, or key.


me4502 (approved):

CDWimmer:
Fab, thanks for doing this so quickly when I asked in discord!

PR #1736, opened 2021-04-23
Address issue template deprecations

NotMyFault:
A few recently introduced strings for issue templates were deprecated and scheduled for removal making templates no longer work properly.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,9 +1,6 @@
 name: Bug report
 description: Report a way in which WorldEdit is not working as intended
-title: ''
 labels: ['type:bug', 'status:pending']
-assignees: []
-issue_body: true
 
 body:
 - type: markdown
@@ -55,6 +52,6 @@ body:
   validations:
     required: true
 
-- type: markdown
+- type: textarea
   attributes:
     value: Add any additional context you can provide below.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,9 +1,6 @@
 name: Bug report
 description: Report a way in which WorldEdit is not working as intended
-title: ''
 labels: ['type:bug', 'status:pending']
-assignees: []
-issue_body: true
```
octylFractal, line 6: Can we add the catchall "Anything else?" textarea to replace this, in both templates? I want to keep that.

octylFractal (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
NotMyFault: I did that here: https://github.com/EngineHub/WorldEdit/pull/1736/files#diff-637f7b97bba458badb691a1557c3d4648686292e948dbe3e8360564378b653efR55
Using `textarea` now makes use of the gh-typical WYSIWYG comment box.

octylFractal: Hmm, yea. Not sure why we had `issue_body` then.

octylFractal (approved):

PR #1750, opened 2021-05-19
Sponge Schematic Version 3

octylFractal:
This PR updates WorldEdit to save and load Sponge Schematic Version 3.

I've rewritten large chunks of the reading class, but it shouldn't have resulted in any changes to how data is read.

~~It might be worth splitting up the Sponge format into 3 separate formats, keeping 1 and 2 writable to allow for users to use software that hasn't had a chance to upgrade to 3 yet. Something like `sponge.1`, `sponge.2`, and `sponge.3`, with `sponge.3` being our new default read and write format.~~ This has been done. Older API consumers should still get version 2 by default, but we now expose each version separately. For command users, reading is unaffected, writing happens with V3 by default but you can specify V2 if you need it. Converting between the two is pretty painless.

~~This PR is currently a draft because the specification itself is still being reviewed. This implementation is to ensure that the spec is sound and will not face any implementation issues.~~ The specification is now complete. Ready for review!


octylFractal:
The schematic specification is now out of draft, and this should be implemented accordingly. Please review!

PR #1766, opened 2021-06-02
Register platforms and commands in a more proper way

octylFractal:
Platforms are now register "on load"/initialize/etc., and preferreds are chosen just prior to command registration events. This allows us to properly hook in to reload. Additionally, we now have the capability to reload data with /minecraft:reload, though this is **not yet implemented**.

Additionally, we will no longer pass out platforms before preferreds are chosen. This is technically breaking, but using platforms in that way was inherently unstable and prone to bugs anyways, so I consider it a bug-fix :).

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/lifecycle/Token.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/lifecycle/Token.java
@@ -17,15 +17,17 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-package com.sk89q.worldedit.event.platform;
-
-import com.sk89q.worldedit.extension.platform.Platform;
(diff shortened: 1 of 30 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/util/lifecycle/Token.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.sk89q.worldedit.util.lifecycle;
+
+/**
+ * Used to create a new strong reference to an object that can be separately dropped.
+ *
+ * @param <T> the inner object
+ */
+class Token<T> {
```
me4502, line 27: Is there not a JVM class that could be used instead?

me4502 (approved):

PR #1770, opened 2021-06-04
Make BukkitWorld lazier for frequent use-case

PaulBGD:
The adapters are used quite frequently by plugins using the WorldEdit
API, with BukkitWorld being one of the more common ones among plugins
such as WorldGuard. As a new BukkitWorld may need to be initialized
every time a physics (redstone and others) event happens, there is a
significant cost to anything put inside the constructor for BukkitWorld.

This commit attempts to reduce the initialization cost by storing a
reference to the world's name instead of holding a WeakReference, which
may actually fix bugs relating to code assuming the world is still
loaded because the reference was not collected yet.

As well it lazily initializes the WorldNativeAccess when it's needed,
which is usually only for modifications that are actually applied to the
world. It does the same for the effect queue. These changes should be
beneficial to everyone, and significantly improve allocations.

A real-world example of a problem this fixes:
image

20% of the allocations on this server are due to redstone triggering WorldGuard to create a BukkitWorld. I personally believe this fix belongs in WorldEdit however, as there shouldn't be a penalty for using a wrapper class.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java
@@ -47,6 +47,7 @@ import com.sk89q.worldedit.world.weather.WeatherType;
 import com.sk89q.worldedit.world.weather.WeatherTypes;
 import io.papermc.lib.PaperLib;
 import org.apache.logging.log4j.Logger;
+import org.bukkit.Bukkit;
 import org.bukkit.Effect;
 import org.bukkit.TreeType;
 import org.bukkit.World;
@@ -58,7 +59,6 @@ import org.bukkit.inventory.DoubleChestInventory;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.InventoryHolder;
 
-import java.lang.ref.WeakReference;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.EnumMap;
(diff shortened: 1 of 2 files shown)


octylFractal:
I would rather just set up a `WeakHashMap<bukkit.World, BukkitWorld>` -- this should reduce allocations even further, and is something we should really be doing anyways. If we make such a change, I would also like to apply it generally -- if you're not up for that, I'll take care of it after Sponge 8.

Even if we don't opt for that, caching the WorldNativeAccess like that would cause the same "bugs relating to code assuming the world is still loaded because the reference was not collected yet." :P

PaulBGD:
The reason for no map is due to the fact that a lookup in a map is likely more expensive than just creating a new object here. The JVM will allocate short lived objects on the stack, so there's no cost assuming it doesn't make more allocations.

As for caching the WorldNativeAccess, my thought was to maintain existing behavior and generally not assume whether the adapter internally uses a weak reference or not. I figured maintaining a cache of it would be better for large world modifications, as to prevent a million instances from being created when pasting a million blocks.

octylFractal:
The JVMs definition of "short-lived" is tedious, and escape analysis may not always kick in, especially not here where it is across more than one class _and_ classloader, and then passed into another method where multiple methods are called on it. I don't think the JVM is that powerful, but feel free to prove me wrong with some allocation profiling with these changes in place.

PaulBGD:
I didn't consider cross classloader, I'd wager you're probably right. Seems the correct solution here is to rewrite all the worldguard listeners to use the standardized class, so that seems to be the proper next move.

PR #1790, opened 2021-06-17
Ensure that all chunks are loaded when copying entities

lukalt:
When using the ForwardExtendCopy to copy entities along with blocks on a Bukkit 1.16.5 server, entities in unloaded chunks won't be copied. 
I think this problem only occurs with the Bukkit implementation, however I have not tested in on sponge etc...

To solve this, I have added a new method to the Extend interface which can be used to load all chunks in a specific region. 
As the method has a default implementation, legacy compatibility is ensured. This method is called before entities are copied. 

Signed-off-by: lukas81298 <lukas81298@gommehd.net>


octylFractal:
I don't think this is a sufficient solution, we probably need to hold a chunk ticket for each chunk for the duration of the operation or the server will likely just unload the chunks again before we query for entities.

octylFractal:
Closing this as the issues mentioned above have not been responded to.

PR #1793, opened 2021-06-21
Improve mask api

BloodEko:
Makes **toMask2D()** a default method, which spares some clutter on client code.
From there on, allows the Mask interface also to be used as a functional interface.

As discussed on discord, here is the PR.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/Mask.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/Mask.java
@@ -26,10 +26,12 @@ import javax.annotation.Nullable;
 /**
  * Tests whether a given vector meets a criteria.
  */
+@FunctionalInterface
 public interface Mask {
 
     /**
-     * Returns true if the criteria is met.
+     * Returns true if the criteria is met and
+     * placement is allowed for the block.
      *
      * @param vector the vector to test
      * @return true if the criteria is met
@@ -42,6 +44,8 @@ public interface Mask {
      * @return a 2D mask version or {@code null} if this mask can't be 2D
      */
     @Nullable
-    Mask2D toMask2D();
+    default Mask2D toMask2D() {
+        return null;
+    }
 
 }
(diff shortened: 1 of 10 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/Mask.java
```diff
@@ -26,10 +26,12 @@
 /**
  * Tests whether a given vector meets a criteria.
  */
+@FunctionalInterface
 public interface Mask {
 
     /**
-     * Returns true if the criteria is met.
+     * Returns true if the criteria is met and
+     * placement is allowed for the block.
```
octylFractal, line 34: Why was this change made? Masks do not necessarily only apply to blocks, I would leave this doc comment as is.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/Mask.java
BloodEko: To me the docs felt a bit unclear wheather it means "the criteria is met to allow" or "the criteria is met to block".
So it might be nice if the ambiguity could be stripped too, but I would leave the phrasing open to you then.

octylFractal: It's neither. Masks can be used to block or allow. It's literally up to the usage of the mask object.

BloodEko: At least the masks I've found so far seem to follow the **true=allow** convention.
So I assume manifesting this in the docs, would make the interface easier to use.
But not sure if any **true=deny** masks exist yet, so I leave the decision up you then.

octylFractal: I would prefer just reverting this for now.

BloodEko: Alright then.

octylFractal (approved):

PR #1800, opened 2021-06-24
Fixed javadocs in Constants.java

MattBDev:
It looks like someone accidently forgot to update the javadocs on these constant fields. This PR updates the javadoc with the proper Minecraft version per field.


octylFractal:
Can you rebase this against `version/7.2.x`? It's technically a (documentation) bug-fix, so it belongs in that branch.

PR #1826, opened 2021-07-12
Support 48x48x48 drawsel on 1.16+

### LINKED ISSUE #1741: The maximum selection area of the structure blocks is outdated
### WorldEdit Version

7.2.5

### Platform Version

Fabric

### Bug Description

Since the version of **Minecraft 1.16**, the **maximum limit** of **structure blocks** has been increased, going from **32x32x32** to **48x48x48**, however when a selection greater than 32 blocks is made, it is not indicated as it should

### Expected Behavior

Selections greater than 32 blocks and less than or equal to 48 blocks should be marked with the structure blocks in any of the axes.

### Reproduction Steps

1. Run the command **//drawsel**
2. Make a selection greater than 32 blocks

### Anything Else?

Add any additional context you can provide below.

me4502:
Fixes #1741

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/ServerCUIHandler.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/ServerCUIHandler.java
@@ -46,7 +46,16 @@ import javax.annotation.Nullable;
  */
 public class ServerCUIHandler {
 
+    private final static int MAX_DISTANCE = 32;
+
     private ServerCUIHandler() {
+    }
+
+    public static int getMaxServerCuiSize() {
+        int dataVersion = WorldEdit.getInstance().getPlatformManager().queryCapability(Capability.GAME_HOOKS).getDataVersion();
+
+        // 1.16 increased maxSize to 48.
+        return dataVersion >= 2566 ? 48 : 32;
     }
 
     /**
@@ -113,11 +122,7 @@ public class ServerCUIHandler {
             return null;
         }
 
-        int dataVersion = WorldEdit.getInstance().getPlatformManager().queryCapability(Capability.GAME_HOOKS).getDataVersion();
-
-        // 1.16 increased maxSize to 48.
-        int maxSize = dataVersion >= 2566 ? 48 : 32;
-        int maxDistance = 32;
+        int maxSize = getMaxServerCuiSize();
 
         if (width > maxSize || length > maxSize || height > maxSize) {
             // Structure blocks have a limit of maxSize^3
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/ServerCUIHandler.java
```diff
@@ -112,8 +113,14 @@ public static BaseBlock createStructureBlock(Player player) {
             return null;
         }
 
-        if (width > 32 || length > 32 || height > 32) {
-            // Structure blocks have a limit of 32x32x32
+        int dataVersion = WorldEdit.getInstance().getPlatformManager().queryCapability(Capability.GAME_HOOKS).getDataVersion();
+
+        // 1.16 increased maxSize to 48.
+        int maxSize = dataVersion >= 2566 ? 48 : 32;
```
octylFractal, line 119: Can we extract this logic somewhere? Feels awkward to have it repeated twice, and if it gets expanded in the future we'd need to remember both places too.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/cui/ServerCUIHandler.java
```diff
@@ -112,8 +113,14 @@ public static BaseBlock createStructureBlock(Player player) {
...
+        int maxDistance = 32;
```
octylFractal, line 120: Why is this declared here? Why isn't it just a constant field?

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -35,7 +35,7 @@
     "worldedit.setbiome.warning": "You may have to re-join your game (or close and re-open your world) to see changes.",
 
     "worldedit.drawsel.disabled": "Server CUI disabled.",
-    "worldedit.drawsel.enabled": "Server CUI enabled. This only supports cuboid regions, with a maximum size of 32x32x32.",
+    "worldedit.drawsel.enabled": "Server CUI enabled. This only supports cuboid regions, with a maximum size of {0}x{0}x{0}.",
```
octylFractal, line 38: I wonder if we should make this `{0}x{1}x{2}` in case the dimensions become different in the future.

octylFractal (changes requested):

worldedit-core/src/main/resources/lang/strings.json
me4502: I somewhat doubt it'd ever get to the point where it's different in the different axis

octylFractal: I'd still like to make this change, I think it's better overall. It doesn't really harm anything, and it allows forward-compatibility.

octylFractal (changes requested):

octylFractal (approved):

PR #1847, opened 2021-07-23
[Forge] Update to 1.17.1

octylFractal:
Includes minor updates to Fabric as well.


worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeEntity.java
```diff
@@ -84,19 +85,19 @@ public boolean setLocation(Location location) {
...
         }
     }
 
     @Override
     public boolean remove() {
-        net.minecraft.entity.Entity entity = entityRef.get();
+        net.minecraft.world.entity.Entity entity = entityRef.get();
         if (entity != null) {
-            entity.remove();
+            entity.remove(net.minecraft.world.entity.Entity.RemovalReason.KILLED);
```
me4502, line 100: Is this the right removal reason?

me4502 (approved):

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorld.java
```diff
@@ -244,46 +237,46 @@ public boolean setBiome(BlockVector3 position, BiomeType biome) {
         checkNotNull(position);
         checkNotNull(biome);
 
-        IChunk chunk = getWorld().getChunk(position.getBlockX() >> 4, position.getBlockZ() >> 4);
-        BiomeContainer container = checkNotNull(chunk.getBiomes());
+        LevelChunk chunk = getWorld().getChunk(position.getBlockX() >> 4, position.getBlockZ() >> 4);
+        ChunkBiomeContainer container = checkNotNull(chunk.getBiomes());
         int idx = BiomeMath.computeBiomeIndex(position.getX(), position.getY(), position.getZ()); // TODO 1.17 pass min/maxY
```
wizjany, line 242: should be changed to new method and old method can be removed.

wizjany (dismissed):

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeEntity.java
octylFractal: It's what I used with Fabric. The other option is `DISCARDED`, but I think we would rather fire the kill event since `remove()` is a lot like `/kill`.

In the future we should probably add a removal reason of our own, to mark cut entities as "discarded"? Not sure about that but it sorta makes sense.

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeEntityProperties.java
```diff
@@ -20,28 +20,28 @@
...
+import net.minecraft.world.entity.decoration.Painting;
+import net.minecraft.world.entity.item.FallingBlockEntity;
+import net.minecraft.world.entity.item.ItemEntity;
+import net.minecraft.world.entity.item.PrimedTnt;
+import net.minecraft.world.entity.npc.Npc;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.entity.projectile.Projectile;
+import net.minecraft.world.entity.vehicle.AbstractMinecart;
+import net.minecraft.world.entity.vehicle.Boat;
+import net.minecraft.world.item.trading.Merchant;
```
me4502, line 44: Yes Mojang, the Merchant NPC is an item

Issue #1722, opened 2021-04-05
ThreadLocal kills interactions


octylFractal:
### WorldEdit Version

7.2.x, 7.3.0-SNAPSHOT

### Platform Version

Any Bukkit Platform, Maybe All

### Bug Description

ThreadLocals have a weird bug that can slow down interact events to a crawl, #1668. We might be able to implement some sort of hack to only have it slow down those with WorldEdit permissions, preventing TPS hits. Or we could just try to fix the bug, either with a WorldEdit refactor or JVM report / patch.

### Expected Behavior

TPS is fine and the JVM actually works.

### Reproduction Steps

????
TODO

Note: this issue is for discussion of some sort of solution to this problem, which currently has an unknown cause but is 100% confirmed.

me4502:
Worth noting that this can only be reproduced with G1GC-based garbage collectors, the CMS GC does not exhibit this behaviour.

Relevant ElasticSearch thread on the issue: https://discuss.elastic.co/t/indexing-performance-degrading-over-time/40229
Prior JDK issue: https://bugs.openjdk.java.net/browse/JDK-8182982

The above issue was closed, mainly due to not enough information on how to reliably reproduce in order to prove that it's actually a Java bug.

octylFractal:
"G1GC-based" is an odd term, but that essentially means ZGC and Shenandoah as well, right? They're not really the same kind as G1GC, just have some similar flaw.

me4502:
Yeah - as far as I'm aware they share common code

octylFractal:
@v9-l10n would you be able to give details on the machine from #1668? I'm wondering if this might be hardware/OS specific, as I haven't really been able to reproduce it at all.

epserv:
Same issue on
- Ubuntu Server 20.04.2 LTS x86_64
- CPU: i7-3770
- Manually compiled jdk15u
- Flags:
```
[...]

Issue #1668, opened 2021-02-08
WE lag at event handler


v9-l10n:
**Versions**

WorldEdit version: 7.2.2
Platform version: Paper 1.16.5 build 464

**Describe the bug**

The server runs normally and TPS drops from 20 to 1 for some time.

**To Reproduce**

I have no idea, this issue has been around us for a while. We thought it would be some players that flooding with packets, but even we added packet limiter with a lower pps limit we still see this happens from time to time.

There is no WE operation ongoing while such issue happens.

**Expected behavior**

Server running normally.

**Screenshots**

None.

**Additional context**

Some stack trace

```
[10:39:01] [Paper Watchdog Thread/ERROR]: --- DO NOT REPORT THIS TO PAPER - THIS IS NOT A BUG OR A CRASH  - git-Paper-445 (MC: 1.16.5) ---
[10:39:01] [Paper Watchdog Thread/ERROR]: The server has not responded for 20 seconds! Creating thread dump
[10:39:01] [Paper Watchdog Thread/ERROR]: ------------------------------
[10:39:01] [Paper Watchdog Thread/ERROR]: Server thread dump (Look for plugins here before reporting to Paper!):
[10:39:01] [Paper Watchdog Thread/ERROR]: ------------------------------
[10:39:01] [Paper Watchdog Thread/ERROR]: Current Thread: Server thread
[10:39:01] [Paper Watchdog Thread/ERROR]: 	PID: 35 | Suspended: false | Native: false | State: RUNNABLE
[10:39:01] [Paper Watchdog Thread/ERROR]: 	Stack:
[10:39:01] [Paper Watchdog Thread/ERROR]: 		java.base@14.0.2/java.lang.ThreadLocal$ThreadLocalMap.expungeStaleEntry(ThreadLocal.java:637)
[10:39:01] [Paper Watchdog Thread/ERROR]: 		java.base@14.0.2/java.lang.ThreadLocal$ThreadLocalMap.remove(ThreadLocal.java:519)
[10:39:01] [Paper Watchdog Thread/ERROR]: 		java.base@14.0.2/java.lang.ThreadLocal.remove(ThreadLocal.java:242)
[10:39:01] [Paper Watchdog Thread/ERROR]: 		com.sk89q.worldedit.session.request.Request.reset(Request.java:111)
[...]

PR #1861, opened 2021-08-02
Replace ThreadLocal with HashMap to eliminate server crash

ReinWD:
Workaround #1668 and #1722.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/session/request/Request.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/session/request/Request.java
@@ -24,13 +24,15 @@ import com.sk89q.worldedit.LocalSession;
 import com.sk89q.worldedit.world.World;
 
 import javax.annotation.Nullable;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
- * Describes the current request using a {@link ThreadLocal}.
+ * Describes the current request using a Thread.name() -> request map(avoid server freeze on scale of 50+ players).
  */
 public final class Request {
 
-    private static final ThreadLocal<Request> threadLocal = ThreadLocal.withInitial(Request::new);
+    private static Map<String, Request> reqMap = new HashMap<>();
 
     private @Nullable World world;
     private @Nullable LocalSession session;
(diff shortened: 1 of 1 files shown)


me4502:
It's worth noting for the servers this affects, if worldedit doesn't have the threadlocal the server still dies from Brigadier's threadlocal

phoenixlzx:
> It's worth noting for the servers this affects, if worldedit doesn't have the threadlocal the server still dies from Brigadier's threadlocal

Yes but we haven't seen Brigadier crashed a server.

Maybe just because player interaction events are way more frequent.

worldedit-core/src/main/java/com/sk89q/worldedit/session/request/Request.java
```diff
@@ -24,13 +24,15 @@
...
+import java.util.Map;
 
 /**
- * Describes the current request using a {@link ThreadLocal}.
+ * Describes the current request using a Thread.name() -> request map(avoid server freeze on scale of 50+ players).
  */
 public final class Request {
 
-    private static final ThreadLocal<Request> threadLocal = ThreadLocal.withInitial(Request::new);
+    private static Map<String, Request> reqMap = new HashMap<>();
```
zml2008, line 35: HashMap is also not thread-safe

zml2008 (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/session/request/Request.java
ReinWD: So how about changing the key of this HashMap?
like thread id or something

zml2008: Do you know what "thread-safe" means?

ReinWD: considering the visibility, I think if we can find a key to identify each thread, it's no need to achieve full thread safe in this case.

octylFractal:
As noted by @zml2008, this PR introduces a thread-safety bug by using `HashMap`, plus it leaks memory if a request is not removed from the map before a thread closes. I've opened #1862 to fix these issues while still doing a workaround.

PR #1862, opened 2021-08-02
Use a Guava Cache instead of a ThreadLocal

### LINKED ISSUE #1722: ThreadLocal kills interactions

octylFractal:
This allows high performance without leaking memory, and works around the JVM bug with ThreadLocals. Fixes #1722.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/session/request/Request.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/session/request/Request.java
@@ -19,6 +19,9 @@
 
 package com.sk89q.worldedit.session.request;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.sk89q.worldedit.EditSession;
 import com.sk89q.worldedit.LocalSession;
 import com.sk89q.worldedit.world.World;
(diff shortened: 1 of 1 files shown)


wizjany (approved):

TomyLobo:
hmm, this JDK bug has been closed as "not a java bug", with reference to a hadoop bug.
what they say in that bug thread is that weak references dont get cleaned up reliably with G1GC.
You told that cache to use weakly referenced keys that we're now relying on in order to remove entries for the dead threads from the cache.

So I don't know how deep you looked into this, but at a surface level this looks like it might have made it worse, not better, but I'm not sure.

me4502:
I'm under the impression that one of the main reasons it was closed was because they had no reliable reproduction environment and therefore no way to confirm it was a Java bug? Either way - if a part of Java depends on weak reference cleanups, and the default GC doesn't handle that reliably, that's a Java bug.

octylFractal:
The main difference is that we are using weakly-referenced `Thread` objects, rather than `ThreadLocal` objects -- we expect there to be usually only a handful of these `Thread`s (main and WorldEdit's processing pools), and for them to be long-lived. They should rarely need to be GC'd, and even then Guava's `Cache` employs reference queues to more eagerly clean-up these weak references. I think this is a safe hold-over solution.

PR #1865, opened 2021-08-05
Use MethodHandle for faster event bus

me4502:
This is mostly for test purposes, gotta do more extensive performance checks

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/eventbus/AnnotatedSubscriberFinder.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/eventbus/AnnotatedSubscriberFinder.java
@@ -22,6 +22,9 @@ package com.sk89q.worldedit.util.eventbus;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.lang.reflect.Method;
 
 /**
@@ -59,7 +62,15 @@ class AnnotatedSubscriberFinder implements SubscriberFindingStrategy {
                         );
                     }
                     Class<?> eventType = parameterTypes[0];
-                    EventHandler handler = new MethodEventHandler(annotation.priority(), listener, method);
+                    MethodHandle handle;
+                    try {
+                        handle = MethodHandles.lookup().unreflect(method);
+                        handle = handle.asType(handle.type().generic().changeReturnType(void.class));
+                    } catch (IllegalAccessException e) {
+                        throw new IllegalArgumentException("Method " + method + " failed to unreflect.", e);
+                    }
+
+                    EventHandler handler = new MethodHandleEventHandler(annotation.priority(), listener, handle);
                     methodsInListener.put(eventType, handler);
                 }
             }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/eventbus/AnnotatedSubscriberFinder.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/eventbus/AnnotatedSubscriberFinder.java
@@ -24,7 +24,6 @@ import com.google.common.collect.Multimap;
 
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/util/eventbus/AnnotatedSubscriberFinder.java
```diff
@@ -59,7 +62,15 @@
                         );
                     }
                     Class<?> eventType = parameterTypes[0];
-                    EventHandler handler = new MethodEventHandler(annotation.priority(), listener, method);
+                    MethodHandle handle;
+                    try {
+                        handle = MethodHandles.lookup().unreflect(method);
+                        handle = handle.asType(handle.type().generic().changeReturnType(void.class));
```
octylFractal, line 68:
```suggestion
                        handle = MethodHandles.lookup().unreflect(method);
                        handle = handle.asType(MethodType.methodType(void.class, Object.class));
```
or maybe even extract the method type to a constant and just do:
```suggestion
                        handle = MethodHandles.lookup().unreflect(method).asType(EVENT_HANDLER_TYPE);
```

octylFractal: Also consider caching `MethodHandles.lookup()`, iirc it's expensive to construct. Or use `publicLookup()`, which is probably more appropriate.

worldedit-core/src/main/java/com/sk89q/worldedit/util/eventbus/MethodHandleEventHandler.java
```diff
@@ -0,0 +1,74 @@
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        MethodHandleEventHandler that = (MethodHandleEventHandler) o;
+
+        if (!methodHandle.equals(that.methodHandle)) {
```
octylFractal, line 68: This is absolutely unusable, method handles don't have equality. You'll need some other equality check, maybe just storing the name of the method and checking the name (method type is always the same so no need to check that).

worldedit-core/src/main/java/com/sk89q/worldedit/util/eventbus/AnnotatedSubscriberFinder.java
octylFractal: ... Also I'd rather move the `asType` to `MethodHandleEventHandler`, to be closer to where the `invokeExact` that needs it is.

worldedit-core/src/main/java/com/sk89q/worldedit/util/eventbus/MethodHandleEventHandler.java
```diff
@@ -0,0 +1,74 @@
...
+        this.methodHandle = methodHandle;
+    }
+
+    @Override
+    public void dispatch(Object event) throws Exception {
+        try {
+            this.methodHandle.invokeExact(object, event);
+        } catch (Throwable e) {
+            // ew
+            throw new Exception(e);
```
octylFractal, line 48: Catch `Exception | Error e` first, re-throw that directly.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/util/eventbus/MethodHandleEventHandler.java
```diff
@@ -0,0 +1,82 @@
...
+     * @param priority the priority
+     * @param object The object to invoke it on
+     * @param methodHandle The handle to invoke
+     * @param methodName The name of the method (for equality checks)
+     */
+    protected MethodHandleEventHandler(Priority priority, Object object, MethodHandle methodHandle, String methodName) {
+        super(priority);
+
+        this.object = object;
+        this.methodHandle = methodHandle.asType(MethodType.methodType(void.class, Object.class, Object.class));
```
octylFractal, line 44: You should bind the method handle to the object, so you don't need a second parameter. Might be faster too.

octylFractal (changes requested):

octylFractal (approved):

PR #1882, opened 2021-08-30
Require 16 in API

octylFractal:
Minecraft 1.17 requires it, and 7.3.0 will drop everything but 1.17 support most likely.


me4502 (approved):
I agree with this;

Just not ditching pre-1.17 in 7.3.0

Would also prefer to wait for wiz to review too

wizjany (approved):

PR #1928, opened 2021-11-07
Add divide command

Lennart99:
This PR adds a relatively simple yet useful command:
**//divide [number] [pattern]**.

This command performs a similar task to the existing **//center [pattern]** command, but gives the user more control by allowing the player to specify a division factor. This command would divide the selected region by the provided quantity and set those blocks to the provided pattern.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -1199,6 +1199,73 @@ public class EditSession implements Extent, AutoCloseable {
     }
 
     /**
+     * Divides the region into N sections and places blocks at these points.
+     * If a point sits between two blocks on a certain axis, then two blocks
+     * will be placed to mark the point.
+     *
+     * @param pattern the replacement pattern
+     * @param pos1 the first position of the region
+     * @param pos2 the second position of the region
+     * @param number the number to divide by
+     * @return the number of blocks placed
+     * @throws MaxChangedBlocksException thrown if too many blocks are changed
+     */
+    public int divide(Pattern pattern, BlockVector3 pos1, BlockVector3 pos2, int number) throws MaxChangedBlocksException {
+        Set<BlockVector3> vset = new HashSet<>();
+
+        int x1 = pos1.getBlockX();
+        int y1 = pos1.getBlockY();
+        int z1 = pos1.getBlockZ();
+        int x2 = pos2.getBlockX();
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
```diff
@@ -67,6 +67,14 @@ default CuboidRegion getBoundingBox() {
      */
     Vector3 getCenter();
 
+    /**
+     * Divides the region into N sections and returns a Set with all unique the points in between those sections.
+     *
+     * @param number the number to divide by
+     * @return the resulting points
+     */
+    List<Vector3> getDivisionPoints(int number);
```
wizjany, line 76: this should have a `default` implementation or it's a breaking change

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
```diff
@@ -67,6 +67,14 @@ default CuboidRegion getBoundingBox() {
      */
     Vector3 getCenter();
 
+    /**
+     * Divides the region into N sections and returns a Set with all unique the points in between those sections.
```
wizjany, line 71: this description is extremely misleading - it divides the *diagonal of the bounding box* and returns a *list* of points (which might only be unique in fractional parts).

wizjany (changes requested):
In the current implementation this is actually worse just //line with a hypothetical equal spaced pattern. It's unimplemented for anything beside cuboids, which means that it will always just place points from the lowest x,y,z in a line to the highest x,y,z. This means you don't even have control over which direction this "line" goes (like you do with the actual //line command). The division of the region is actually a division of this line - any other axes are ignored. Additionally the way rounding is done means this will create multiple overlapping 2x2 cubes if the "division factor" is relatively high compared to the length of the diagonal.

On that note, I feel like this would be a lot more useful as a grid command instead which truly subdivides the space in all axes, or better yet, a grid pattern that allows non-random repetition of sub-patterns. The syntax for doing this as a pattern in a 3d space might be a bit weird though.

Lennart99:
I decided to re-write this feature and base the logic a bit more on the logic of //line. This simplified the logic and also optimized it quite a bit (it should be faster then //line now). This also means it won't generate 2x2 cubes anymore.
It however still is only implemented for cuboid regions, but it is now taking the correct positions (pos1 and 2 instead of min and max). I would like to also implement this for other region types, but I am not sure how this should work. 
Do you have any ideas?

PR #1930, opened 2021-11-10
Update paperweight + userdev bundle.

wizjany:


buildSrc/src/main/kotlin/AdapterConfig.kt
```diff
@@ -42,6 +29,6 @@ fun Project.applyPaperweightAdapterConfiguration(
     }
 
     configurations[REOBF_CONFIG].attributes {
-        attribute(Obfuscation.OBFUSCATION_ATTRIBUTE, objects.named(Obfuscation.REOBFUSCATED))
+        attribute(Obfuscation.OBFUSCATION_ATTRIBUTE, objects.named(Obfuscation.OBFUSCATED))
     }
```
octylFractal, line 33: I think we can remove this entirely.

octylFractal (changes requested):

octylFractal (approved):

PR #1939, opened 2021-11-14
1.18

octylFractal:
It's finally here, and we can't wait.

Drafted until 1.18 comes out. Use at your own risk.


wizjany:
todo: check if /tree warpedfungus/crimsonfungus/chorusplant are broken or what

wizjany (approved):

PR #1943, opened 2021-11-16
Remove finalize, use a Cleaner instead

### LINKED ISSUE #1927: Migrate away from finalize()
Currently we still use finalize() to power TracedEditSession, this will be [going away](https://openjdk.java.net/jeps/421) in the far future. We should move this to the new `Cleaner` API.

octylFractal:
This fixes #1927.

I also enabled the trace by default, since it's pretty useful. It can always be turned off if it's causing issues.


worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/QueryTool.java
```diff
@@ -53,9 +52,8 @@ public boolean canUse(Actor player) {
     public boolean actPrimary(Platform server, LocalConfiguration config, Player player, LocalSession session, Location clicked, @Nullable Direction face) {
 
         World world = (World) clicked.getExtent();
-        EditSession editSession = session.createEditSession(player);
         BlockVector3 blockPoint = clicked.toVector().toBlockPoint();
-        BaseBlock block = editSession.getFullBlock(blockPoint);
+        BaseBlock block = player.getExtent().getFullBlock(blockPoint);
```
wizjany, line 56:
```suggestion
        BaseBlock block = clicked.getExtent().getFullBlock(blockPoint);
```
unlikely a player will be able to click a block in some extent they aren't in, but just to be pedantic.

me4502 (approved):

Issue #1895, opened 2021-09-12
Unable to use JavaAdapter in CraftScripts


Snownee:
### WorldEdit Version

7.2.5

### Platform Version

Forge

### Confirmations

- [X] I am using the most recent Minecraft release.
- [X] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [X] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

Due to the current implementation of `MinecraftHidingClassShutter`, run the script below will show `Access to Java class "adapter1" is prohibited`

https://github.com/EngineHub/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/scripting/MinecraftHidingClassShutter.java#L30

Which means you can't extend any class by CraftScripts

### Expected Behavior

ability to use JavaAdapter in CraftScripts

### Reproduction Steps

Use this script:
```js
// $Id$
/*
 * Test CraftScript for WorldEdit
 * Copyright (C) 2021 Snownee
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program. If not, see <http://www.gnu.org/licenses/>.
*/

let we = com.sk89q.worldedit.WorldEdit.getInstance()
let factory = we.getBlockFactory()
let parser = new JavaAdapter(com.sk89q.worldedit.extension.factory.parser.DefaultBlockParser, {
	parseFromInput: function(input, context) {
		return null
	}
}, we)

```

octylFractal:
[...]

PR #1947, opened 2021-11-19
Re-add class shutter with tweaks.

wizjany:
We want to hide default-package classes, such as "com", "io", etc. which are common names
in Mojang obf classes but also common package names, without hiding generated default-package
classes such as described in #1895.

This reverts commit 6008fe73


worldedit-core/src/main/java/com/sk89q/worldedit/scripting/RhinoCraftScriptEngine.java
```diff
@@ -45,9 +45,11 @@ public int getTimeLimit() {
     }
 
     @Override
-    public Object evaluate(String script, String filename, Map<String, Object> args) throws Throwable {
+    public Object evaluate(String script, String filename, Map<String, Object> args)
+            throws ScriptException, Throwable {
```
octylFractal, line 49: Unnecessary? I know it came from the revert, just doesn't seem needed.

octylFractal (approved):

PR #1956, opened 2021-12-03
Include mapColor in BlockMaterial and add more delegations to Forge/Fabric BlockMaterials

DalekCraft2:
Maybe I'll do stuff with translucency in the JSON files and MCUtil later. I'unno.


DalekCraft2:
Actually, I just discovered that the MCUtils repository shows how to get some of these, so I am going to update this again.

me4502:
As we've deprecated the concept of bundled block data and added delegations separately, I'm going to close this PR. I'm also not sure if the map colour attribute is something we necessarily need/want at this point, as this PR doesn't seem to provide an actual usage of it.

PR #1959, opened 2021-12-05
Fix snapshots in 1.18

Brokkonaut:


worldedit-core/src/main/java/com/sk89q/worldedit/world/chunk/AnvilChunk18.java
```diff
@@ -0,0 +1,224 @@
...
+     * @param tag the tag to read
+     * @throws DataException on a data error
+     */
+    public AnvilChunk18(CompoundTag tag) throws DataException {
+        rootTag = tag;
+
+        rootX = NBTUtils.getChildTag(rootTag.getValue(), "xPos", IntTag.class).getValue();
+        rootZ = NBTUtils.getChildTag(rootTag.getValue(), "zPos", IntTag.class).getValue();
+
+        blocks = new HashMap<>();
```
octylFractal, line 68: Perhaps initialize this after `sections` and use the size of the list to initialize the size of the map.

octylFractal (changes requested):

octylFractal (approved):
Would like a second look-over from @me4502 or @wizjany.

me4502 (commented):
We've had reports that having an entities folder inside the world can break restore. Does this PR also fix that (I ask because I don't see anything related to it - and I assume that's semi important for 1.17-1.18 support)

Brokkonaut:
> We've had reports that having an entities folder inside the world can break restore. Does this PR also fix that (I ask because I don't see anything related to it - and I assume that's semi important for 1.17-1.18 support)

This should be fixed now too

me4502 (approved):

PR #1964, opened 2021-12-08
Create SECURITY.md

octylFractal:
We've not actually stated a formal policy, and in leu of email this should suffice.


wizjany (approved):

wizjany:
is it worth noting that the scope is "admin tool" and giving perms to untrusted users is *their* threat vector, not ours?

octylFractal:
Yea, I'll add that.

PR #1968, opened 2021-12-12
Merge 1.18.1 to master

me4502:
My local gradle is broken so building via TC


buildSrc/src/main/kotlin/PlatformConfig.kt
```diff
@@ -11,13 +11,14 @@ import org.gradle.kotlin.dsl.named
 import org.gradle.kotlin.dsl.register
 import kotlin.collections.set
 
-fun Project.applyPlatformAndCoreConfiguration() {
+fun Project.applyPlatformAndCoreConfiguration(javaRelease: Int = 16) {
```
octylFractal, line 14: I think this should be 17 now, yes?

buildSrc/src/main/kotlin/CommonJavaConfig.kt
```diff
@@ -11,7 +11,7 @@ import org.gradle.kotlin.dsl.dependencies
 import org.gradle.kotlin.dsl.get
 import org.gradle.kotlin.dsl.withType
 
-fun Project.applyCommonJavaConfiguration(sourcesJar: Boolean, banSlf4j: Boolean = true) {
+fun Project.applyCommonJavaConfiguration(sourcesJar: Boolean, javaRelease: Int = 16, banSlf4j: Boolean = true) {
```
octylFractal, line 14: 17 here too.

worldedit-bukkit/adapters/adapter-1.17.1/build.gradle.kts
```diff
@@ -1,3 +1,17 @@
-applyPaperweightAdapterConfiguration(
```
octylFractal, line 1: I would just drop this folder entirely, we're removing all legacy adapters with 7.3.0, and that means 1.17 too.

octylFractal (changes requested):

octylFractal (approved):

Issue #1983, opened 2021-12-22
Smooth is non-functional below y = 0


octylFractal:
### WorldEdit Version

7.2.9

### Platform Version

All Platforms

### Confirmations

- [X] I am using the most recent Minecraft release.
- [X] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [X] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

Smoothing is breaking below y = 0 for some reason. Needs a look at how the kernel is computed.

### Expected Behavior

Smooth should work the same above and below y = 0.

### Reproduction Steps

1. Go down.
2. Select
3. //smooth

octylFractal:
Fixed by #1986 / #1989.

rreeoo:
Is it fixed ?
I can't still use "/br smoorh" below y=0 prorerly in 1.18.1

PR #1989, opened 2021-12-28
Fix for #1983, Use Math.floor instead of int cast (master branch)

JOO200:
Same as #1986 for the mater branch.

Fixed //snowsmooth too.


me4502:
Not to be an annoyance - but would it be possible for this to please just include the master-specific changes so that we can retain the version/7.2.x commit for the master branch?

octylFractal (approved):

PR #1999, opened 2022-01-06
Deprecate MultiStageReorder and ReorderMode

me4502:
As MultiStage should be removed, and FAST/NONE do not make sense as a toggle, this system is being deprecated for removal in WE8.


worldedit-core/src/main/java/com/sk89q/worldedit/extent/reorder/MultiStageReorder.java
```diff
@@ -42,7 +42,10 @@
 
 /**
  * Re-orders blocks into several stages.
+ *
+ * @deprecated Use {@link com.sk89q.worldedit.extent.world.SideEffectExtent} with post edit instead.
```
octylFractal, line 46: `post edit` typo? Not sure what you wanted to refer to here. Maybe link to `#setPostEditSimulationEnabled`?

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/extent/reorder/MultiStageReorder.java
```diff
@@ -42,7 +42,10 @@
 
 /**
  * Re-orders blocks into several stages.
+ *
+ * @deprecated Use {@link com.sk89q.worldedit.extent.world.SideEffectExtent} with {@link com.sk89q.worldedit.extent.world.SideEffectExtent#setPostEditSimulationEnabled} instead.
```
octylFractal, line 46: Can you import the class so this isn't so long?

octylFractal (changes requested):

octylFractal (approved):

PR #2000, opened 2022-01-06
Add the UPDATE SideEffect to Forge

me4502:
Adds the UPDATE Mixin from Fabric to the Forge WorldEdit, and sets up Mixins for Forge.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -56,5 +56,5 @@ dependencies {
     implementation("net.fabricmc:sponge-mixin:$mixinVersion")
     implementation("org.enginehub.gradle:gradle-codecov-plugin:0.1.0")
     implementation("io.papermc.paperweight.userdev:io.papermc.paperweight.userdev.gradle.plugin:1.3.3")
-    implementation("org.spongepowered:mixingradle:0.7.+")
+    implementation("org.spongepowered:mixingradle:0.7.32")
 }
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -56,4 +56,5 @@ dependencies {
     implementation("net.fabricmc:sponge-mixin:$mixinVersion")
     implementation("org.enginehub.gradle:gradle-codecov-plugin:0.1.0")
     implementation("io.papermc.paperweight.userdev:io.papermc.paperweight.userdev.gradle.plugin:1.3.3")
+    implementation("org.spongepowered:mixingradle:0.7.+")
 }
--- a/worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgePlatform.java
+++ b/worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgePlatform.java
@@ -19,6 +19,7 @@
 
 package com.sk89q.worldedit.forge;
 
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.command.util.PermissionCondition;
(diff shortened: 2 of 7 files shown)


buildSrc/build.gradle.kts
```diff
@@ -56,4 +56,5 @@ dependencies {
     implementation("net.fabricmc:sponge-mixin:$mixinVersion")
     implementation("org.enginehub.gradle:gradle-codecov-plugin:0.1.0")
     implementation("io.papermc.paperweight.userdev:io.papermc.paperweight.userdev.gradle.plugin:1.3.3")
+    implementation("org.spongepowered:mixingradle:0.7.+")
```
octylFractal, line 59: No. Absolutely not. Use a real version.

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgePlatform.java
```diff
@@ -212,17 +215,23 @@ public String getPlatformVersion() {
...
     );
 
+    private static final Set<SideEffect> SUPPORTED_SIDE_EFFECTS = Sets.immutableEnumSet(
+        Iterables.concat(SUPPORTED_SIDE_EFFECTS_NO_MIXIN, Collections.singleton(SideEffect.UPDATE))
+    );
+
     @Override
     public Set<SideEffect> getSupportedSideEffects() {
-        return SUPPORTED_SIDE_EFFECTS;
+        return ExtendedChunk.class.isAssignableFrom(LevelChunk.class)
```
octylFractal, line 232: Does this even do anything? Can mixins fail to load randomly?

octylFractal (changes requested):

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/mixin/MixinLevelChunkSetBlockHook.java
```diff
@@ -0,0 +1,85 @@
...
+        method = "setBlockState",
+        slice = @Slice(
+            from = @At(value = "INVOKE", target = "Lnet/minecraft/world/level/block/state/BlockState;is(Lnet/minecraft/world/level/block/Block;)Z")
+        ),
+        at = @At(value = "INVOKE", target = "Lnet/minecraft/world/level/block/state/BlockState;onPlace(Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/level/block/state/BlockState;Z)V")
+    )
+    public void setBlockStateHook(BlockState target, Level world, BlockPos pos, BlockState old, boolean move) {
+        boolean localShouldUpdate;
+        MinecraftServer server = world.getServer();
+        if (server == null || Thread.currentThread() != server.getRunningThread()) {
```
zml2008, line 74: why not just `if (world.isClientSide/world.isRemote/whatever the mappings are)`?

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgePlatform.java
zml2008: Sometimes forge will change their transformer system during the update process without giving Mixin notice/time to update for their changes, and will temporarily remove the mixin library from their runtime until it gets updated.

me4502: It's also possible for another mod to explicitly disable our mixin - I feel a fallback is safest

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/mixin/MixinLevelChunkSetBlockHook.java
me4502: This was copied from Fabric - I am assuming there was a reason for it (probably in case something is dumb and calls updates on another thread)

octylFractal: Especially because the `shouldUpdate` flag isn't volatile, that would be a thread safety violation.

octylFractal (approved):

PR #2004, opened 2022-01-09
Show stack trace only when user has detailed error permission

Janikio:
This PR adds the functionality that error stack traces only get shown to players if they have the `worldedit.error.detailed` permission otherwise a general error message gets sent to the actor.

This feature was added as stack traces can expose the underlying platform and code base the server runs on. Bad actors could abuse this behavior to gather information and find matching vulnerabilities.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ErrorReporting.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ErrorReporting.java
@@ -30,10 +30,8 @@ import com.sk89q.worldedit.util.formatting.text.event.HoverEvent;
  */
 public class ErrorReporting {
     public static void trigger(Actor actor, Throwable error) {
-        boolean showDetailedError = actor.hasPermission("worldedit.error.detailed");
-
         actor.printError(TranslatableComponent.of("worldedit.command.error.report"));
-        if (showDetailedError) {
+        if (actor.hasPermission("worldedit.error.detailed")) {
             actor.print(
                 TextComponent.builder(error.getClass().getName() + ": " + error.getMessage())
                     .hoverEvent(HoverEvent.showText(TextComponent.of(Throwables.getStackTraceAsString(error))))
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ErrorReporting.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ErrorReporting.java
@@ -32,9 +32,8 @@ public class ErrorReporting {
     public static void trigger(Actor actor, Throwable error) {
         boolean showDetailedError = actor.hasPermission("worldedit.error.detailed");
 
-        actor.printError(TranslatableComponent.of("worldedit.command.error"));
+        actor.printError(TranslatableComponent.of("worldedit.command.error.report"));
         if (showDetailedError) {
-            actor.printError(TranslatableComponent.of("worldedit.command.error.report"));
             actor.print(
                 TextComponent.builder(error.getClass().getName() + ": " + error.getMessage())
                     .hoverEvent(HoverEvent.showText(TextComponent.of(Throwables.getStackTraceAsString(error))))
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ErrorReporting.java
```diff
@@ -30,12 +30,17 @@
...
-        actor.print(
-            TextComponent.builder(error.getClass().getName() + ": " + error.getMessage())
-                .hoverEvent(HoverEvent.showText(TextComponent.of(Throwables.getStackTraceAsString(error))))
-                .build()
-        );
+        boolean showDetailedError = actor.hasPermission("worldedit.error.detailed");
+
+        actor.printError(TranslatableComponent.of("worldedit.command.error"));
+        if (showDetailedError) {
+            actor.printError(TranslatableComponent.of("worldedit.command.error.report"));
```
octylFractal, line 37: I would rather not add a new message:
```suggestion
        actor.printError(TranslatableComponent.of("worldedit.command.error.report"));
        if (showDetailedError) {
```

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -263,6 +263,7 @@
     "worldedit.command.time-elapsed": "{0}s elapsed (history: {1} changed; {2} blocks/sec).",
     "worldedit.command.permissions": "You are not permitted to do that. Are you in the right mode?",
     "worldedit.command.player-only": "This command must be used with a player.",
+    "worldedit.command.error": "An unexpected error occurred trying to run your command",
```
octylFractal, line 266: In tandem with the above:
```suggestion
```

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ErrorReporting.java
```diff
@@ -30,12 +30,17 @@
  */
 public class ErrorReporting {
     public static void trigger(Actor actor, Throwable error) {
-        actor.printError(TranslatableComponent.of("worldedit.command.error.report"));
-        actor.print(
-            TextComponent.builder(error.getClass().getName() + ": " + error.getMessage())
-                .hoverEvent(HoverEvent.showText(TextComponent.of(Throwables.getStackTraceAsString(error))))
-                .build()
-        );
+        boolean showDetailedError = actor.hasPermission("worldedit.error.detailed");
```
octylFractal, line 33: This could probably be inlined. I don't think the separate variable adds much.

octylFractal (changes requested):

octylFractal:
P.S. -- if you're going to sign your commits, you should add your GPG key to GitHub so they can verify them properly.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/ErrorReporting.java
```diff
@@ -31,11 +31,13 @@
...
         actor.printError(TranslatableComponent.of("worldedit.command.error.report"));
-        actor.print(
-            TextComponent.builder(error.getClass().getName() + ": " + error.getMessage())
-                .hoverEvent(HoverEvent.showText(TextComponent.of(Throwables.getStackTraceAsString(error))))
-                .build()
-        );
+        if (actor.hasPermission("worldedit.error.detailed")) {
+            actor.print(
+                TextComponent.builder(error.getClass().getName() + ": " + error.getMessage())
+                    .hoverEvent(HoverEvent.showText(TextComponent.of(Throwables.getStackTraceAsString(error))))
```
octylFractal, line 37: Maybe we should only put the hover event behind the permission, as it was before. @wizjany & @me4502 thoughts?

wizjany: yea i think the message itself is still useful for the user tbh

me4502: Yeah IMO just doing prior behaviour without the perm node makes sense to me.

me4502:
Superseded by https://github.com/EngineHub/WorldEdit/pull/2122, due to long time without update

Issue #2020, opened 2022-01-28
`//regen` fails in Forge environment


gabizou:
### WorldEdit Version

7.2.9-SNAPSHOT (ab61a84) and 7.2.8

### Platform Version

Forge 1.18.1-39.0.63

### Confirmations

- [X] I am using the most recent Minecraft release.
- [X] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [X] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

Using //regen on a region that ends up generating a BlockEntity (presumably in the regenerated terrain) causes an error. Terrain that doesn't generate a BE doesn't throw the error

### Expected Behavior

Regen to work fine

### Reproduction Steps

1. Select region with a BE involved (go to a village, there's bounds to be BE's there)
2. //regen

### Anything Else?

Output looks like so:
```
[01:33:48] [Render thread/INFO]: [CHAT] Second position set to (-1427, -2, -4115) (622104).
[01:33:53] [Server thread/ERROR]: An unexpected error while handling a WorldEdit command
java.lang.NoSuchMethodError: 'net.minecraft.nbt.CompoundTag net.minecraft.world.level.block.entity.BlockEntity.save(net.minecraft.nbt.CompoundTag)'
	at com.sk89q.worldedit.forge.ForgeWorld.regenForWorld(ForgeWorld.java:431) ~[worldedit-mod-7.2.9-SNAPSHOT-dist.jar%2353!/:?]
	at com.sk89q.worldedit.forge.ForgeWorld.doRegen(ForgeWorld.java:343) ~[worldedit-mod-7.2.9-SNAPSHOT-dist.jar%2353!/:?]
	at com.sk89q.worldedit.forge.ForgeWorld.regenerate(ForgeWorld.java:303) ~[worldedit-mod-7.2.9-SNAPSHOT-dist.jar%2353!/:?]
	at com.sk89q.worldedit.command.RegionCommands.regenerate(RegionCommands.java:431) ~[worldedit-mod-7.2.9-SNAPSHOT-dist.jar%2353!/:?]
[...]

PR #2021, opened 2022-01-28
Fix regen on newer Forge due to removed method

me4502:
Looks like Forge removed the BlockEntity#save method for some reason, this switches to what we do on Fabric

Fixes https://github.com/EngineHub/WorldEdit/issues/2020


worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorld.java
```diff
@@ -427,9 +426,7 @@ private void regenForWorld(Region region, Extent extent, ServerLevel serverWorld
             BlockStateHolder<?> state = ForgeAdapter.adapt(chunk.getBlockState(pos));
             BlockEntity blockEntity = chunk.getBlockEntity(pos);
             if (blockEntity != null) {
-                net.minecraft.nbt.CompoundTag tag = new net.minecraft.nbt.CompoundTag();
-                blockEntity.save(tag);
-                state = state.toBaseBlock(NBTConverter.fromNative(tag));
+                state = state.toBaseBlock(NBTConverter.fromNative(blockEntity.saveWithId()));
```
wizjany, line 429: this method will drop ForgeData and ForgeCaps, should those be preserved? (the previous method did save them, as well as id and xyz, which matches to saveWithFullMetadata. iirc xyz isn't needed but we set the correct ones back in the output.)

me4502: Switched to that

octylFractal (approved):

PR #2035, opened 2022-02-21
Load config file before posting PlatformsRegisteredEvent

### LINKED ISSUE #1845: Command Logging Non-functional
### WorldEdit Version

v7.2.6-SNAPSHOT+5a143b8

### Platform Version

git-Paper-104

### Bug Description

Command Logging feature apparently doesn't create a log file, and therefore is completely non-functional.

### Expected Behavior

Commands are logged to a file.

### Reproduction Steps

1. Enable command logging in config.
2. Select a cuboid and run `//set stone`.
3. where `worldedit.log`

### Anything Else?

_No response_

dordsor21:
- The config file is accessed by some internal listeners of the PlatformsRegisteredEvent
 - Specifically command registration is done based on this event being fired, and this requires some config file access
 -  Fixes #1845

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -56,6 +56,6 @@ dependencies {
     implementation("net.fabricmc:fabric-loom:$loomVersion")
     implementation("net.fabricmc:sponge-mixin:$mixinVersion")
     implementation("org.enginehub.gradle:gradle-codecov-plugin:0.1.0")
-    implementation("io.papermc.paperweight.userdev:io.papermc.paperweight.userdev.gradle.plugin:1.3.4")
+    implementation("io.papermc.paperweight.userdev:io.papermc.paperweight.userdev.gradle.plugin:1.3.5")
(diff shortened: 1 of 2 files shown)


octylFractal:
I can't remember exactly what, but there were other issues with config ordering. Needs a deeper look to ensure this doesn't break other things (like legacy item loading).

dordsor21:
That's a good point, the LegacyMapper class does require `Capability.CONFIGURATION` to be available...

As far I can tell, this should be the only thing. The loading of platform capabilities should probably not be done all at once, as obviously loading the server is always highly likely to (and does) require certain capabilities to be available in a certain order/consecutively to handle everything properly.

I suppose a "load config" then a "parse config" would be required in order for the specific case of the legacy mapper. However, I see no reason why ResourceLoader and by extension the json files under `com/sk89q/worldedit/world/registry` can only be accessed once the configuration capability is available (which would fix the "loop" with LegacyMapper/configuration file).

Issue #2027, opened 2022-02-02
UPDATE is not triggered with applySideEffects


octylFractal:
### WorldEdit Version

7.2.x, 7.3.0-SNAPSHOT

### Platform Version

All Platforms

### Confirmations

- [X] I am using the most recent Minecraft release.
- [X] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [X] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

UPDATE side-effect does not get applied when using `applySideEffects`. For fast reorder to work properly, this _must_ occur.

### Expected Behavior

UPDATE is triggered on the block.

### Reproduction Steps

1. `//reorder fast`
2. Set a single sand block in the air.
3. It doesn't fall. Oops.

me4502:
Fixed in 30e1601489b3ed4a416a21de0595f444dad2c945

PR #2037, opened 2022-02-22
Apply update flag on later change

me4502:
Allows the UPDATE side effect to be applied later. It's default for now to not break the old adapters, but can be made non-default in 7.3.0

Fixes https://github.com/EngineHub/WorldEdit/issues/2027

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/wna/WorldNativeAccess.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/wna/WorldNativeAccess.java
@@ -146,6 +146,8 @@ public interface WorldNativeAccess<NC, NBS, NP> {
 
     void notifyNeighbors(NP pos, NBS oldState, NBS newState);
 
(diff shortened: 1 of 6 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/internal/wna/WorldNativeAccess.java
```diff
@@ -184,6 +187,10 @@ default void markAndNotifyBlock(NP pos, NC chunk, NBS oldState, NBS newState, Si
 
         // Seems used only for PoI updates
         onBlockStateChange(pos, oldState, blockState1);
+
+        if (sideEffectSet.shouldApply(SideEffect.UPDATE)) {
+            updateBlock(pos, oldState, newState);
+        }
```
octylFractal, line 193: Should be moved to before `LIGHTING` in `applySideEffects`, otherwise this'll get called on normal set-block too, which is unnecessary.

octylFractal (changes requested):

octylFractal (approved):

PR #2058, opened 2022-03-12
Fix passenger entity saving and loading

### LINKED ISSUE #1763: //copy and //stack do not copy entities riding status
### WorldEdit Version

7.2.5

### Platform Version

Fabric 0.11.3

### Bug Description

Trying to copy a region with entities riding minecarts. Both, minecart and mob get duplicated, but the mob is not riding the minecart anymore.

### Expected Behavior

When copying a mob riding a minecart the mob should also ride it in the copy.

### Reproduction Steps

1. Copy / Stack a region with -e argument

### Anything Else?

I noticed when duplicating layers for testing a cat-based creeper farm. The riding status is considered important here, because if not in minecart the cat will teleport if the player hurts.

wizjany:
Fixes #1763.

Original code from https://github.com/EngineHub/WorldEdit/issues/1763#issuecomment-1062429891

TODO:
* [x] Backport to 1.18/1.17 adapters (not doing older ones lol)
* [x] Port to Fabric/Forge(/Sponge?)
* [x] Test

### DIFF
--- a/worldedit-bukkit/adapters/adapter-1.17.1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_17_R1_2/PaperweightAdapter.java
+++ b/worldedit-bukkit/adapters/adapter-1.17.1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_17_R1_2/PaperweightAdapter.java
@@ -475,8 +475,8 @@ public final class PaperweightAdapter implements BukkitImplAdapter {
         }
 
         // Adapted from net.minecraft.world.entity.EntityType#loadEntityRecursive
-        if (tag.contains("Passengers", 9)) {
-            net.minecraft.nbt.ListTag nbttaglist = tag.getList("Passengers", 10);
+        if (tag.contains("Passengers", NBTConstants.TYPE_LIST)) {
+            net.minecraft.nbt.ListTag nbttaglist = tag.getList("Passengers", NBTConstants.TYPE_COMPOUND);
 
             for (int i = 0; i < nbttaglist.size(); ++i) {
                 removeUnwantedEntityTagsRecursively(nbttaglist.getCompound(i));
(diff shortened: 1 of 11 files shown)


worldedit-bukkit/adapters/adapter-1.17.1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_17_R1_2/PaperweightAdapter.java
```diff
@@ -437,27 +442,48 @@ public BaseEntity getEntity(org.bukkit.entity.Entity entity) {
...
 
+    // This removes all unwanted tags from the main entity and all its passengers
+    private void removeUnwantedEntityTagsRecursively(net.minecraft.nbt.CompoundTag tag) {
+        for (String name : Constants.NO_COPY_ENTITY_NBT_FIELDS) {
+            tag.remove(name);
+        }
+
+        // Adapted from net.minecraft.world.entity.EntityType#loadEntityRecursive
+        if (tag.contains("Passengers", 9)) {
+            net.minecraft.nbt.ListTag nbttaglist = tag.getList("Passengers", 10);
```
octylFractal, line 479: Can we use the NBT type constants?

worldedit-sponge/build.gradle.kts
```diff
@@ -71,7 +71,7 @@ tasks.named<ShadowJar>("shadowJar") {
         }
         include(dependency(":worldedit-core"))
 
-        relocate("org.antlr.v4", "com.sk89q.worldedit.sponge.antlr4")
+        relocate("org.antlr.v4", "com.sk89q.worldedit.antlr4")
```
octylFractal, line 74: Why change this?

octylFractal (changes requested):

worldedit-sponge/build.gradle.kts
wizjany: all the other platforms put it here

octylFractal (approved):

PR #1559, opened 2020-10-11
Add erosion brush

lewisjb:
Implements the erode brush specified in #1288

PR #2078, opened 2022-04-10
Add morph brush

me4502:
Supersedes https://github.com/EngineHub/WorldEdit/pull/1559

This is a morph brush, that allows applying combinations of morphological dilation and erosion. It also comes with two presets (erode, and dilate)


octylFractal (approved):
I don't see anything obviously wrong with this.

PR #2079, opened 2022-04-11
[1.16.5] Updated to Forge Permission API

MagnusHJensen:
My original plan was to also make a PR for 1.18.2, however the PermissionAPI has drastically changed, and I'm not too experienced in 1.18 yet.

I'm hoping this is the right way to create a PR for 1.16.

This is a simple change, that replaces the `canSendCommands` call with the permission api call.
This allows for permission mods to replace the PermissionAPI and have their own checks. Under the hood, the permission API call, if no permission value found, it calls the `ServerLifecycleHooks.getCurrentServer().getPlayerList().canSendCommands(player.getGameProfile())`

This was tested with FTBRanks for 1.16.5, where I gave a group the permissions for `worldedit.wand` and `worldedit.selection.pos`
If the 2nd permission wasn't given it popped up with the WE perm error message, which to me indicated that it worked, plus I could only run the `//wand` command and nothing else.

### DIFF
--- a/worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgePermissionsProvider.java
+++ b/worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgePermissionsProvider.java
@@ -22,6 +22,7 @@ package com.sk89q.worldedit.forge;
 import net.minecraft.entity.player.ServerPlayerEntity;
 import net.minecraft.world.GameType;
 import net.minecraftforge.fml.server.ServerLifecycleHooks;
+import net.minecraftforge.server.permission.DefaultPermissionLevel;
 import net.minecraftforge.server.permission.PermissionAPI;
 
 public interface ForgePermissionsProvider {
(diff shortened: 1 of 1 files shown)


MagnusHJensen:
I forgot to register all permission nodes to the VanillaPermissionHandler, so it defaulted to OP. 
If that was not done, not even OP's could use the commands! 

Should be good to go now!

wizjany:
if the forge perms api requires a permission to be registered then this is not going to work. not all perms are attached to commands so they won't ever get registered and then would always come up false, even though they should be true for op. you're going to need to restore the op/cansendcommands check.

MagnusHJensen:
> if the forge perms api requires a permission to be registered then this is not going to work. not all perms are attached to commands so they won't ever get registered and then would always come up false, even though they should be true for op. you're going to need to restore the op/cansendcommands check.

Okay, reintroduced the op/cansendcommands check, and remove registering any permissions nodes. So the permissionAPI call is only useful it it has been replaced by a seperate mod.

octylFractal (approved):

octylFractal:
Going to merge this, but we won't be making any new releases, so it'll still have to be built from scratch.

PR #2082, opened 2022-04-20
Fix int overflow when using fillr in negative y-coordinates

### LINKED ISSUE #973: //fillr with negativ y coordinates
WORLDEDIT-3223 - Reported by Jaegermeister

If you try to use //fillr with a position  y < 0
Error:
http://pastebin.com/pB64pq8w

using:
WorldEdit Build #3347 and newest Spigot Build

dordsor21:
- Also actually check the depth arg against the radius limit
 - Fixes #973

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
@@ -101,7 +101,6 @@ public class UtilityCommands {
         radius = Math.max(1, radius);
         we.checkMaxRadius(radius);
         depth = Math.max(1, depth);
-        we.checkMaxRadius(depth);
 
         BlockVector3 pos = session.getPlacementPosition(actor);
         int affected = editSession.fillXZ(pos, pattern, radius, depth, false);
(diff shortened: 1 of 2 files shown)


dordsor21:
If checking the depth size against max allowed radius should be done for fillr then it should also be done for fill, otherwise, the checkmaxradius in fillr should be removed

octylFractal:
I don't understand how this fixes a negative y-coord issue. It seems like you only changed the value used if `depth` isn't passed, but I can't connect that to fixing y-coord issues if you passed e.g. 1 instead.

dordsor21:
The depth is subtracted from the height of the player in the fillXZ method in editsession https://github.com/EngineHub/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java#L1007. Using Integer.MAX_VALUE causes an overflow, so changing the default depth fixes it.

dordsor21:
Separate to that (which would just be a bound to Integer.MIN_VALUE I suppose), should the depth be able to be larger than the max radius configured?

worldedit-core/src/main/java/com/sk89q/worldedit/command/UtilityCommands.java
```diff
@@ -101,6 +101,7 @@ public int fill(Actor actor, LocalSession session, EditSession editSession,
         radius = Math.max(1, radius);
         we.checkMaxRadius(radius);
         depth = Math.max(1, depth);
+        we.checkMaxRadius(depth);
```
me4502, line 104: So from looking at this, it doesn't actually appear that any of the changes in UtilityCommands.java are actually required here. And these do modify the behaviour of the commands.

Can this PR please be modified to only include the changes within EditSession.java?

dordsor21: Yeah, will do

dordsor21: All done!

PR #2084, opened 2022-04-23
Fix/asc desc loop

dordsor21:
Feel free to make a separate commit addressing differently, however, just PRing this as the same as I did to quickly patch FAWE


dordsor21:
Closing due to poor local git setup

PR #2085, opened 2022-04-23
Add a check to the "levels" argument supplied for //asc and //desc

dordsor21:
Feel free to commit separately with a different implementation, just keeping this the same as the quick FAWE patch. Can change downstream to whatever you guys implement as well.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/NavigationCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/NavigationCommands.java
@@ -29,11 +29,13 @@ import com.sk89q.worldedit.entity.Player;
 import com.sk89q.worldedit.util.Location;
 import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import com.sk89q.worldedit.util.formatting.text.TranslatableComponent;
+import com.sk89q.worldedit.world.World;
 import org.enginehub.piston.annotation.Command;
 import org.enginehub.piston.annotation.CommandContainer;
 import org.enginehub.piston.annotation.param.Arg;
 import org.enginehub.piston.annotation.param.Switch;
 
+import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/NavigationCommands.java
```diff
@@ -75,6 +77,8 @@ public void unstuck(Player player) throws WorldEditException {
     public void ascend(Player player,
                        @Arg(desc = "# of levels to ascend", def = "1")
                            int levels) throws WorldEditException {
+        World world = player.getWorld();
+        checkArgument(levels >= 1 && levels <= (world.getMaxY() - world.getMinY()), "1 <= levels <= world height");
```
octylFractal, line 81:
```suggestion
        checkArgument(1 <= levels && levels <= (world.getMaxY() - world.getMinY()), "1 <= levels <= world height");
```
This makes it look more like the error message, and I prefer to have bounds done this way.

worldedit-core/src/main/java/com/sk89q/worldedit/command/NavigationCommands.java
```diff
@@ -98,6 +102,8 @@ public void ascend(Player player,
     public void descend(Player player,
                         @Arg(desc = "# of levels to descend", def = "1")
                             int levels) throws WorldEditException {
+        World world = player.getWorld();
+        checkArgument(levels >= 1 && levels <= (world.getMaxY() - world.getMinY()), "1 <= levels <= world height");
```
octylFractal, line 106: Ditto here.
```suggestion
        checkArgument(1 <= levels && levels <= (world.getMaxY() - world.getMinY()), "1 <= levels <= world height");
```

octylFractal (changes requested):
Minor styling fix, and would you mind extracting this to a common method so if it changes both places get updated?

dordsor21:
For reference: IntellectualSites/FastAsyncWorldEdit#1718

PR #2103, opened 2022-05-22
Cache properties to reduce startup memory

me4502:
I don't personally believe this is the best solution here, but a better one would be breaking (Removing public constructor for properties and having a BooleanProperty.create function that returns a cached version if it exists). That'd be a WE8 change though (And one I believe we should do)

Very dirty test of loading the fabric server (with no other mods, vanilla blocks only) before and after showed a decrease in total allocation from 3228MB -> 2560MB

### DIFF
--- a/worldedit-bukkit/adapters/adapter-1.17.1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_17_R1_2/PaperweightAdapter.java
+++ b/worldedit-bukkit/adapters/adapter-1.17.1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_17_R1_2/PaperweightAdapter.java
@@ -500,6 +500,26 @@ public final class PaperweightAdapter implements BukkitImplAdapter {
     }
 
     @SuppressWarnings({ "unchecked", "rawtypes" })
+    private static final LoadingCache<net.minecraft.world.level.block.state.properties.Property, Property<?>> propertyCache = CacheBuilder.newBuilder().build(new CacheLoader<net.minecraft.world.level.block.state.properties.Property, Property<?>>() {
+        @Override
+        public Property<?> load(net.minecraft.world.level.block.state.properties.Property state) throws Exception {
+            if (state instanceof net.minecraft.world.level.block.state.properties.BooleanProperty) {
+                return new BooleanProperty(state.getName(), ImmutableList.copyOf(state.getPossibleValues()));
+            } else if (state instanceof DirectionProperty) {
+                return new DirectionalProperty(state.getName(),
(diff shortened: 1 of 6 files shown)


worldedit-bukkit/adapters/adapter-1.17.1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_17_R1_2/PaperweightAdapter.java
```diff
@@ -500,28 +500,34 @@ public Component getRichItemName(BaseItemStack itemStack) {
...
     @SuppressWarnings({ "unchecked", "rawtypes" })
-    @Override
-    public Map<String, ? extends Property<?>> getProperties(BlockType blockType) {
-        Map<String, Property<?>> properties = Maps.newTreeMap(String::compareTo);
-        Block block = getBlockFromType(blockType);
-        StateDefinition<Block, net.minecraft.world.level.block.state.BlockState> blockStateList =
-            block.getStateDefinition();
-        for (net.minecraft.world.level.block.state.properties.Property state : blockStateList.getProperties()) {
-            Property property;
+    private static final LoadingCache<net.minecraft.world.level.block.state.properties.Property, Property<?>> propertyCache = CacheBuilder.newBuilder().build(new CacheLoader<net.minecraft.world.level.block.state.properties.Property, Property<?>>() {
```
octylFractal, line 503: As a `private static final` this should be `PROPERTY_CACHE` IMO. Same for the others.

worldedit-bukkit/adapters/adapter-1.17.1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_17_R1_2/PaperweightAdapter.java
```diff
@@ -500,28 +500,34 @@ public Component getRichItemName(BaseItemStack itemStack) {
...
             } else {
                 throw new IllegalArgumentException("WorldEdit needs an update to support " + state.getClass().getSimpleName());
             }
+        }
+    });
 
+    @SuppressWarnings({ "rawtypes" })
+    @Override
+    public Map<String, ? extends Property<?>> getProperties(BlockType blockType) {
+        Map<String, Property<?>> properties = Maps.newTreeMap(String::compareTo);
```
octylFractal, line 525: Strings are natively comparable. We don't need the extra indirection. Also, we don't need `Maps`.
```suggestion
        Map<String, Property<?>> properties = new TreeMap<>();
```

octylFractal (changes requested):

octylFractal (approved):

PR #2110, opened 2022-06-05
Add a new blockBreakEffect method and deprecate World#playEffect

Owen1212055:
- Fixes the break effect used for super pickaxe using legacy ids
- Fixes block break effect not displaying on fabric/forge
- Allows you to now queue block effects of a blockstate rather than blocktype.

Please give feedback on the deprecation of ``World#playEffect`` and the old ``World#queueBlockBreakEffect`` method.
I was also split between making a separate method for the block break effect or making some kind of generic ``playEffect(Vector3 pos, int effect, T data)`` method that would do a bunch of instanceof checks to correctly adapt the data.
If a generic play effect method is preferred rather than a specific playBreakBlockEffect method, let me know!

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java
@@ -371,6 +371,14 @@ public class BukkitWorld extends AbstractWorld {
     }
 
     @Override
+    public boolean playBreakBlockEffect(Vector3 position, com.sk89q.worldedit.world.block.BlockState block) {
+        World world = getWorld();
+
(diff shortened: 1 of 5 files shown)


me4502:
My intention was for the replacement to this method to be generic, however I hadn't looked into it given how complex it seemed to be to implement now

I feel a bunch of instanceof checks would become extremely messy. How do other APIs implement this?

Owen1212055:
> My intention was for the replacement to this method to be generic, however I hadn't looked into it given how complex it seemed to be to implement now
> 
> I feel a bunch of instanceof checks would become extremely messy. How do other APIs implement this?

Well, i'm really only aware of the bukkit implementation, as it doesn't seem that the other platforms expose this at all. But other platforms such as bukkit check the individual keys which use the data field and cast it accordingly. There aren't many effects that use the data field, but it would require having a new Effect class instead of the int identifier. That can certainly be done if it's preferred.
image

PR #2113, opened 2022-06-08
Sort of fix #2008

### LINKED ISSUE #2008: Forest generation works bad in 1.18.1
### WorldEdit Version

7.2.9-SNAPSHOT+6045-3fa364b

### Platform Version

git-Paper-141 (MC: 1.18.1)

### Confirmations

### Bug Description

`/forestgen` generates new trees above existing ones.

image

### Expected Behavior

Before update to 1.18.1, in version 1.17.1, it worked correctly and doesn't place dirt blocks inside leaves. I expect that in the new version it should work the same way.

### Reproduction Steps

1. Go to plains;
2. Use `/forestgen` multiple times

### Anything Else?

_No response_

octylFractal:
This fix only applies to Forge/Fabric, and it still produces floating trees... but it's better than nothing


octylFractal:
Needs a better solution.

PR #2135, opened 2022-07-03
Fix Sponge rotation adapter

Yeregorix:
Pitch and yaw are currently inverted in worldedit-sponge. This makes most player rotation based actions totally broken: compass teleportation, `//expand` etc.


me4502:
This should be targetting `version/7.2.x` if it's a bugfix

Yeregorix:
My work was based on `version/7.2.x` but I just forgot to select this branch when opening the PR.

octylFractal (approved):

PR #2151, opened 2022-07-23
Refactor PlatformManager for performance

anonyco:
Two improvements sure to provide a nice speedup to WorldEdit:
1. Desynchronize `<PlatformManager>.queryCapability`, which is used in some pretty tight loops that would feel the benefits of less `synchronized`.
2. Remove the use of `EnumMap` from `PlatformManager.preferences`, which further optimizes `<PlatformManager>.queryCapability` because `EnumMap`s do incur some overhead, mostly from checking to make sure the `enum` you passed it is really of the same class as the `enum` of the `EnumMap`.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
@@ -70,7 +70,8 @@ public class PlatformManager {
     private final WorldEdit worldEdit;
     private final PlatformCommandManager platformCommandManager;
     private final List<Platform> platforms = new ArrayList<>();
-    private final Map<Capability, Platform> preferences = new EnumMap<>(Capability.class);
+    private final Capability[] preferences_keys = Capability.class.getEnumConstants();
+    private final Platform[] preferences_values = new Platform[ preferences_keys.length ];
     private @Nullable String firstSeenVersion;
     private final AtomicBoolean initialized = new AtomicBoolean();
     private final AtomicBoolean configured = new AtomicBoolean();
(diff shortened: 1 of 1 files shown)


me4502:
Refactors for performance that make the code substantially messier (the fact you didn't follow the style guide at all aside), without any proof of speed-up or evidence that it's necessary, won't be accepted.

Provide a microbenchmark and CPU profile that show why these changes make sense to do if you want to PR something like this.

I'm also not convinced it couldn't be done in a substantially cleaner manner, if it's true that this has a performance impact

anonyco:
I have significantly simplified these changes to the bare minimum: the removal of just the `synchronized`.

1. I was unaware that there was a style guide. I read all the markdown and txt files, including CONTRIBUTING.md. Could you refer me to where to find the style guide?

2. The results of my microbenchmark are below:

```
Before:
     Calculated BlockType.getRichName x10000 in 84128654ns
     Calculated BlockType.getRichName x10000 in 14269327ns
     Calculated BlockType.getRichName x10000 in 7806132ns

After:
     Calculated BlockType.getRichName x10000 in 60178043ns
     Calculated BlockType.getRichName x10000 in 7953745ns
     Calculated BlockType.getRichName x10000 in 4852244ns
```

Across the runs above, we see 40%-79% better performance for `BlockType.getRichName` with this small change.

3. I am very unfamiliar with the available Java tools and do not know "CPU profile." I am not asking you to waste your time explaining every step to me, rather I would really appreciate it if you could give me a piece of text or a name of a software/tool that I could look up and research to learn more about what the common way to do this is in the WorldEdit project.

Many thanks for your time and attention to this.

octylFractal (changes requested):
I would rather refactor the whole class to stop using `synchronized` at all, and instead take advantage of a `ReadWriteLock` (probably `ReentrantReadWriteLock` to start with). It'll increase performance everywhere and is not as potentially dangerous as unguarded reads from a thread-unsafe map. If you don't feel up to that, please close this PR and I will file an issue to do so at a later date.

me4502:
closing in favour of https://github.com/EngineHub/WorldEdit/pull/2571

PR #2155, opened 2022-07-24
Fix tasks not being scheduled on Sponge

Yeregorix:
Tasks on SpongeAPI 8 must be submitted to a scheduler.


me4502 (approved):
Thanks for this

PR #2171, opened 2022-08-13
Work around invalid legacy values in schematics

me4502:
Not 100% happy about this; but currently a lot of legacy schematics include entirely broken blocks (Eg, `1:5` rather than `1:0`) because they used broken or poor quality world editing tools to alter the world. These are invalid post-1.13, but we can default to the default state for these blocks to at least recover _some_ of the data that was meant to be here, rather than skipping it entirely.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/MCEditSchematicReader.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/MCEditSchematicReader.java
@@ -420,7 +420,12 @@ public class MCEditSchematicReader extends NBTSchematicReader {
     }
 
     private BlockState getBlockState(int id, int data) {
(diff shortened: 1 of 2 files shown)


octylFractal (changes requested):
Can we move this logic to the schematic loader itself?

me4502:
@octylFractal Given these blocks are "equivalent", I feel the mapper is the right place to do it. Plus, I feel it'd be weird if the blocks work in the schem but not via other commands

wizjany:
imo they *shouldn't* be valid in other commands. `//set 1:5` is a typo or some other mistake. we should treat it as such and error to the user. the only reason this makes sense in the context of schems is because users can get schematics from dumb sources that have invalid data and there's no easy way to fix the schem itself.

octylFractal:
Yes, I agree with wiz here.

octylFractal (approved):

PR #2172, opened 2022-08-15
Shade LinBus in -libs

me4502:
This is exposed in API so I put it in libs. This fixes running WE7.3


buildSrc/src/main/kotlin/LibsConfig.kt
```diff
@@ -87,6 +88,10 @@ fun Project.applyLibrariesConfiguration() {
         from({
             altConfigFiles("sources")
         })
+
+        // Yeet module-info's
+        exclude("module-info.java")
```
me4502, line 93: These cause duplications in the sources jar, hence excluding _.java_ rather than _.class_

octylFractal (approved):

PR #2173, opened 2022-08-15
Add parsing support for block NBT

me4502:
Adds block NBT parsing support. 

Test with `//set chest{'Lock':'cake'}`, then notice you can only open the chest with an item named `cake`

Test with `//set oak_sign[rotation=12]{'is_waxed':1,'front_text':{'messages':['{"text":"a"}','{"text":"a"}','{"text":"a"}','{"text":"a","color":"red"}']}}` for a waxed sign with 'a' on every line where the last line is red

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
@@ -41,6 +41,7 @@ import com.sk89q.worldedit.internal.util.DeprecationUtil;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.registry.state.Property;
 import com.sk89q.worldedit.util.HandSide;
+import com.sk89q.worldedit.util.concurrency.LazyReference;
 import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import com.sk89q.worldedit.util.formatting.text.TranslatableComponent;
 import com.sk89q.worldedit.world.World;
@@ -52,6 +53,9 @@ import com.sk89q.worldedit.world.block.FuzzyBlockState;
 import com.sk89q.worldedit.world.entity.EntityType;
 import com.sk89q.worldedit.world.entity.EntityTypes;
 import com.sk89q.worldedit.world.registry.LegacyMapper;
+import org.enginehub.linbus.format.snbt.LinStringIO;
+import org.enginehub.linbus.stream.exception.NbtParseException;
+import org.enginehub.linbus.tree.LinCompoundTag;
 
 import java.util.HashMap;
 import java.util.Locale;
@@ -254,9 +258,13 @@ public class DefaultBlockParser extends InputParser<BaseBlock> {
         if (blockAndExtraData.length == 0) {
             throw new NoMatchException(TranslatableComponent.of("worldedit.error.unknown-block", TextComponent.of(input)));
         }
-        blockAndExtraData[0] = woolMapper(blockAndExtraData[0]);
+        if (context.isTryingLegacy()) {
+            // Perform a legacy wool colour mapping
+            blockAndExtraData[0] = woolMapper(blockAndExtraData[0]);
+        }
 
         BlockState state = null;
+        LinCompoundTag blockNbtData = null;
 
         // Legacy matcher
         if (context.isTryingLegacy()) {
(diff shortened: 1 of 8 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -272,21 +280,41 @@ private BaseBlock parseLogic(String input, ParserContext context) throws InputPa
 
         if (state == null) {
             String typeString;
-            String stateString = null;
+
             int stateStart = blockAndExtraData[0].indexOf('[');
-            if (stateStart == -1) {
+            int nbtStart = blockAndExtraData[0].indexOf('{');
+            int typeEnd = stateStart == -1 ? nbtStart : stateStart;
```
octylFractal, line 286: Avoid second index if not needed? Maybe expand into an `if` statement to keep the name.
```suggestion
            int typeEnd = stateStart == -1 ? blockAndExtraData[0].indexOf('{') : stateStart;
```

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -272,21 +280,41 @@ private BaseBlock parseLogic(String input, ParserContext context) throws InputPa
...
-                typeString = blockAndExtraData[0].substring(0, stateStart);
+                typeString = blockAndExtraData[0].substring(0, typeEnd);
+            }
+
+            String stateString = null;
+            if (stateStart != -1) {
                 if (stateStart + 1 >= blockAndExtraData[0].length()) {
                     throw new InputParseException(TranslatableComponent.of("worldedit.error.parser.hanging-lbracket", TextComponent.of(stateStart)));
                 }
                 int stateEnd = blockAndExtraData[0].lastIndexOf(']');
```
octylFractal, line 299: Check behavior for this test case: `chest[double={whoops:"nbt ]"}`; it might pass, or fail with a weird error.

I suspect we might need to be a little more strict in what is accepted here. It might be worth writing an ANTLR grammar for this instead, since it's getting quite complex.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -272,21 +280,41 @@ private BaseBlock parseLogic(String input, ParserContext context) throws InputPa
...
-                stateString = blockAndExtraData[0].substring(stateStart + 1, blockAndExtraData[0].length() - 1);
+                stateString = blockAndExtraData[0].substring(stateStart + 1, stateEnd);
             }
+
+            String nbtString = null;
+            if (nbtStart != -1) {
+                if (nbtStart + 1 >= blockAndExtraData[0].length()) {
+                    throw new InputParseException(TranslatableComponent.of("worldedit.error.parser.hanging-lbrace", TextComponent.of(nbtStart)));
+                }
+                int nbtEnd = blockAndExtraData[0].lastIndexOf('}');
```
octylFractal, line 311: Similarly here, we might cut off NBT early if there's `}` in an NBT string, and someone doesn't put one at the end, resulting in a weird error.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -358,6 +389,25 @@ private BaseBlock parseLogic(String input, ParserContext context) throws InputPa
                     state = state.with(objProp, blockState.getValue());
                 }
             }
+
+            if (nbtString != null) {
+                LinCompoundTag otherTag;
+                try {
+                    System.out.println(nbtString);
```
octylFractal, line 396: Leftover debugging.
```suggestion
```

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
me4502: This is used later so it's not unneeded

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
me4502: `State is missing trailing ']'` is the error we get here, which appears to be the correct error

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
me4502: This will result in an invalid NBT error

octylFractal:
Closes #2375

octylFractal (approved):

PR #2180, opened 2022-08-28
Ignore structure void by default and use a flag to paste it

me4502:
Given Minecraft intends these to be a "blank" space for structure pasting, it makes sense for us to ignore these by default. I've added a flag to re-enable pasting these

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/ClipboardBrush.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/ClipboardBrush.java
@@ -41,25 +41,13 @@ public class ClipboardBrush implements Brush {
     private final Mask sourceMask;
 
     public ClipboardBrush(ClipboardHolder holder, boolean ignoreAirBlocks, boolean usingOrigin) {
-        this.holder = holder;
-        this.ignoreAirBlocks = ignoreAirBlocks;
-        this.ignoreStructureVoidBlocks = true;
-        this.usingOrigin = usingOrigin;
-        this.pasteBiomes = false;
-        this.pasteEntities = false;
-        this.sourceMask = null;
+        this(holder, ignoreAirBlocks, usingOrigin, false, false, null);
     }
 
     @Deprecated
     public ClipboardBrush(ClipboardHolder holder, boolean ignoreAirBlocks, boolean usingOrigin, boolean pasteEntities,
                           boolean pasteBiomes, Mask sourceMask) {
-        this.holder = holder;
-        this.ignoreAirBlocks = ignoreAirBlocks;
-        this.ignoreStructureVoidBlocks = true;
-        this.usingOrigin = usingOrigin;
-        this.pasteEntities = pasteEntities;
-        this.pasteBiomes = pasteBiomes;
-        this.sourceMask = sourceMask;
+        this(holder, ignoreAirBlocks, false, usingOrigin, pasteEntities, pasteBiomes, sourceMask);
     }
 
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/session/PasteBuilder.java
```diff
@@ -46,6 +48,7 @@ public class PasteBuilder {
 
     private BlockVector3 to = BlockVector3.ZERO;
     private boolean ignoreAirBlocks;
+    private boolean ignoreStructureVoidBlocks = true; // default to true as it's their intended purpose
```
octylFractal, line 51: I don't want to do this for API, at least not in 7.x. This changes existing behavior.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/ClipboardBrush.java
```diff
@@ -42,16 +43,30 @@ public class ClipboardBrush implements Brush {
...
         this.holder = holder;
         this.ignoreAirBlocks = ignoreAirBlocks;
+        this.ignoreStructureVoidBlocks = true;
+        this.usingOrigin = usingOrigin;
+        this.pasteEntities = pasteEntities;
+        this.pasteBiomes = pasteBiomes;
+        this.sourceMask = sourceMask;
+    }
+
+    public ClipboardBrush(ClipboardHolder holder, boolean ignoreAirBlocks, boolean ignoreStructureVoidBlocks,
```
wizjany, line 65: the amount of params here is getting ugly but can we at least collapse these into properly overloaded constructors (`this(holder, ignoreAir...);`) so we reduce the repetition and re-setting of fields.

worldedit-core/src/main/java/com/sk89q/worldedit/session/PasteBuilder.java
```diff
@@ -101,6 +104,21 @@ public PasteBuilder ignoreAirBlocks(boolean ignoreAirBlocks) {
         return this;
     }
 
+    /**
+     * Set whether structure void blocks in the source are skipped over when pasting.
+     *
+     * <p>
+     * This defaults to true, to better align to how Minecraft intends these blocks to function.
```
wizjany, line 111: docs should be updated to a) mention the correct default and b) mention the intent to change in a major-version update.
it's best if people explicitly set this to whatever they want to avoid defaulting to something they might not.

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/ClipboardBrush.java
```diff
@@ -42,16 +43,30 @@ public class ClipboardBrush implements Brush {
     public ClipboardBrush(ClipboardHolder holder, boolean ignoreAirBlocks, boolean usingOrigin) {
         this.holder = holder;
         this.ignoreAirBlocks = ignoreAirBlocks;
+        this.ignoreStructureVoidBlocks = true;
```
wizjany, line 46: afaict, this is being defaulted to `true` for anyone creating their own clipboardbrush, but `false` for anyone creating a paste via pastebuilder. is there a reason for this mismatch?

wizjany (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/brush/ClipboardBrush.java
octylFractal: We could even introduce a new `record` for these parameters, and have the canonical constructor be the record-taking one.

octylFractal (approved):

wizjany (approved):

PR #2187, opened 2022-09-12
Update README.md (Minor Changes)

vishwakrishnan02:
Main changes made: Added table of contents, additional download links and instructions for other Java IDEs, basic instructions on Git, and changes to the welcome section (additional details).


octylFractal:
Most of this seems out of scope. We're not here to list every option and teach everything. Realistically some parts of the README should even be removed. For those reasons this PR will not be accepted.

PR #2189, opened 2022-09-18
Add API to disable history tracking in EditSessions

me4502:
Adds API to get whether history is tracked, or enable/disable it. A command could be added to handle this in the future too.

Many API usecases never touch history at all, so it's basically time spent that could be avoided

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/ChangeSetExtent.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/ChangeSetExtent.java
@@ -46,6 +46,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 public class ChangeSetExtent extends AbstractDelegateExtent {
 
     private final ChangeSet changeSet;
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extent/ChangeSetExtent.java
```diff
@@ -54,30 +55,63 @@ public class ChangeSetExtent extends AbstractDelegateExtent {
...
      */
     public ChangeSetExtent(Extent extent, ChangeSet changeSet) {
+        this(extent, changeSet, true);
+    }
+
+    /**
+     * Create a new instance.
+     *
+     * @param extent the extent
+     * @param changeSet the change set
```
octylFractal, line 65:
```suggestion
     * @param changeSet the change set
     * @param enabled if the extent is enabled
```

octylFractal (changes requested):

octylFractal (approved):

PR #2212, opened 2022-11-08
SchematicsManager: Handling and caching list of known schematics, providing cli suggestions

### LINKED ISSUE #2095: Add tab completion for loading schematics
### The Problem

Currently i can only use /schematic list to see whats in there but then have to type it in manually

### A Solution

Add tab completetion for all available schmatic shown in the list command

### Alternatives

_No response_

### Anything Else?

_No response_

seijikun:
I implemented a `SchematicsManager` that handles the listing / caching of known schematic files.
The `SchematicsManager` has multiple possible backends.
The preferred backend is:

`FileWatcherSchematicsBackend`:
This backend uses Java's `WatchService` to listen for changes in the `schematics` directory. This rids us of any file listing IO.

`PollingSchematicsBackend`:
This is the fallback backend for the previous mentioned that is only used when the previous failed to initialize (for whatever reason that might be). It caches the result and returns the previous result as long as it's not older than `10` seconds. This is basically an Eventually Consistent Cache.

I then went on to implement a `SchematicConverter` to parse Schematics from arguments and provide suggestions.

This fixes #2095


worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/Schematic.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package com.sk89q.worldedit.util.schematic;
+
+import java.nio.file.Path;
+
+/**
+ * Record representing one Schematic file.
+ */
+public record Schematic(Path path) {
```
me4502, line 27: I haven't had a complete chance to review the PR yet sorry, but it definitely looks great - thanks for putting this together :) 

One part I'm not super sure about is calling the record that represents a schematic path just `Schematic`. Given we might theoretically have data types that represent the data stored in a schematic file in the future, or that users might get confused about what this data actually represents, it might make more sense to give it a name that implies it's purely the path. For example, `SchematicLocator`, `SchematicDescriptor`, `SchematicPath`, etc. I somewhat feel `SchematicPath` might make the most sense, given it feels like a specialisation of the `Path` type for schematics

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/Schematic.java
```diff
@@ -0,0 +1,53 @@
...
+     * Get this Schematic's path.
+     * @return This Schematic's path.
+     */
+    public Path getPath() {
+        return path;
+    }
+
+    @Override
+    public String toString() {
+        return path.toString();
```
me4502, line 39: I don't believe toString or hashCode is strictly necessary for records. As the toString method returns different to the default this one makes sense to keep, but the hashCode method is identical to what a record produces natively

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/Schematic.java
```diff
@@ -0,0 +1,53 @@
...
+ * Record representing one Schematic file.
+ */
+public record Schematic(Path path) {
+
+    /**
+     * Get this Schematic's path.
+     * @return This Schematic's path.
+     */
+    public Path getPath() {
+        return path;
```
me4502, line 34: Records generally don't have getters, as they generate the internal `path()` getter for the path object in the constructor

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/Schematic.java
```diff
@@ -0,0 +1,53 @@
...
+        return path.toString();
+    }
+
+    @Override
+    public int hashCode() {
+        return path.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
me4502, line 48: This equals method shouldn't be necessary, as the default equals method for records is identical to this one

seijikun:
This change in action:
(The suggestions for `//schematics load ...` and `//schematics delete ...`)

https://user-images.githubusercontent.com/820845/200692406-920a1447-10e2-4685-9d22-23ccc53685d6.mp4

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/FileWatcherSchematicsBackend.java
```diff
@@ -0,0 +1,98 @@
...
+        directoryWatcher.start(event -> {
+            lock.writeLock().lock();
+            if (event instanceof RecursiveDirectoryWatcher.FileCreatedEvent) {
+                schematics.add(new SchematicPath(event.getPath()));
+            } else if (event instanceof RecursiveDirectoryWatcher.FileDeletedEvent) {
+                schematics.remove(new SchematicPath(event.getPath()));
+            }
+            lock.writeLock().unlock();
+            if (event instanceof RecursiveDirectoryWatcher.FileCreatedEvent) {
+                LOGGER.info("New Schematic found: " + event.getPath());
```
me4502, line 73: I assume these info logger points are debug code?

me4502 (commented):
I'm happy with this PR outside of the noted log points, but I'd also like someone else to have a go over it too :)

me4502 (approved):

JOO200:
I noticed that the implementation doesn't handle symlinks correctly.
With the setting `files.allow-symbolic-links` the symlinks will be followed.

Another issue with symlinks: Using recursive symlinks doesn't work fine:
1. Create a folder in schematics (e.g. `mkdir dummy`)
2. Create a symlink that points to one directory upwards (e.g. `cd dummy && ln -fs ../ test`)
With this setup WorldEdit will follow the schematic path:
```
[12:45:30 INFO]: New Schematic found: /home/user/dev/server/paper/plugins/WorldEdit/schematics/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test/dummy/test
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/file/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,226 @@
...
+        return new RecursiveDirectoryWatcher(root, watchService);
+    }
+
+    private void registerFolderWatchAndScanInitially(Path root) throws IOException {
+        WatchKey watchKey = root.register(watchService, StandardWatchEventKinds.ENTRY_CREATE, StandardWatchEventKinds.ENTRY_DELETE);
+        LOGGER.debug("Watch registered: " + root);
+        watchRootMap.put(watchKey, root);
+        eventConsumer.accept(new DirectoryCreatedEvent(root));
+        for (Path path : Files.newDirectoryStream(root)) {
+            if (Files.isDirectory(path)) {
```
octylFractal, line 129: Should respect the allow symlinks configuration flag.

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/file/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,226 @@
...
+                            LOGGER.warn("RecursiveDirectoryWatcher Seems like we can't keep up with updates");
+                            continue;
+                        }
+                        // make sure to work with an absolute path
+                        Path path = (Path) event.context();
+                        Path parentPath = watchRootMap.get(watchKey);
+                        path = parentPath.resolve(path);
+
+                        if (kind.equals(StandardWatchEventKinds.ENTRY_CREATE)) {
+                            if (Files.isDirectory(path)) { // new subfolder created, create watch for it
```
octylFractal, line 171: Should respect the allow symlinks configuration flag.

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/file/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,226 @@
...
+                WatchKey watchKey;
+                while (true) {
+                    try {
+                        watchKey = watchService.take();
+                    } catch (InterruptedException e) { break; }
+
+                    for (WatchEvent<?> event : watchKey.pollEvents()) {
+                        WatchEvent.Kind<?> kind = event.kind();
+                        if (kind.equals(StandardWatchEventKinds.OVERFLOW)) {
+                            LOGGER.warn("RecursiveDirectoryWatcher Seems like we can't keep up with updates");
```
octylFractal, line 162: The class name is included in the logs.
```suggestion
                            LOGGER.warn("Seems like we can't keep up with updates");
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/file/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,226 @@
...
+            } catch (ClosedWatchServiceException ignored) { }
+            LOGGER.debug("RecursiveDirectoryWatcher::EventConsumer exited");
+        });
+        watchThread.setName("RecursiveDirectoryWatcher");
+        watchThread.start();
+    }
+
+    /**
+     * Stop this RecursiveDirectoryWatcher instance and wait for it to be completely shut down.
+     * @warning RecursiveDirectoryWatcher is not reusable!
```
octylFractal, line 211: Same question about the `@warning` here.

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/file/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,226 @@
...
+     */
+    public void stop() {
```
octylFractal, line 213: Why doesn't this just implement `(Auto)Closeable`?

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/file/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,226 @@
...
+        try {
+            watchService.close();
+        } catch (IOException e) { e.printStackTrace(); }
```
octylFractal, line 216: Should be using the logger.

worldedit-core/src/main/java/com/sk89q/worldedit/util/io/file/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,226 @@
...
+        if (watchThread != null) {
+            try {
+                watchThread.join();
+            } catch (InterruptedException e) { e.printStackTrace(); }
```
octylFractal, line 220: Should be using the logger.

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/DummySchematicsBackend.java
```diff
@@ -0,0 +1,48 @@
...
+package com.sk89q.worldedit.util.schematic.backends;
+
+import com.sk89q.worldedit.util.schematic.SchematicPath;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * DummyBackend for the SchematicsManager.
+ * This is only used in the error-case. For example when creating the schematics folder failed for whatever reason.
```
octylFractal, line 29:
```suggestion
 * A backend that never lists any files.
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/FileWatcherSchematicsBackend.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * SchematicsBackend making use of the RecursiveDirectoryWatcher.
+ * This backend initially scans all schematics in the folder tree and then registers for file change events to
+ * avoid manually polling / rescanning the folder structure for every query.
```
octylFractal, line 38:
```suggestion
 * A backend that efficiently scans for file changes using {@link RecursiveDirectoryWatcher}.
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/FileWatcherSchematicsBackend.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public void init() {
+        directoryWatcher.start(event -> {
+            lock.writeLock().lock();
+            if (event instanceof RecursiveDirectoryWatcher.FileCreatedEvent) {
+                schematics.add(new SchematicPath(event.getPath()));
+            } else if (event instanceof RecursiveDirectoryWatcher.FileDeletedEvent) {
+                schematics.remove(new SchematicPath(event.getPath()));
+            }
+            lock.writeLock().unlock();
```
octylFractal, line 71:
```suggestion
            try {
                if (event instanceof RecursiveDirectoryWatcher.FileCreatedEvent) {
                    schematics.add(new SchematicPath(event.getPath()));
                } else if (event instanceof RecursiveDirectoryWatcher.FileDeletedEvent) {
                    schematics.remove(new SchematicPath(event.getPath()));
                }
            } finally {
                lock.writeLock().unlock();
            }
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/FileWatcherSchematicsBackend.java
```diff
@@ -0,0 +1,97 @@
...
+    public void uninit() {
+        directoryWatcher.stop();
+    }
+
+    @Override
+    public List<SchematicPath> getList() {
+        lock.readLock().lock();
+        List<SchematicPath> result = new ArrayList<>(schematics);
+        lock.readLock().unlock();
+        return result;
```
octylFractal, line 90:
```suggestion
        try {
            return List.copyOf(schematics);
        } finally {
            lock.readLock().unlock();
        }
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/FileWatcherSchematicsBackend.java
```diff
@@ -0,0 +1,97 @@
...
+        });
+    }
+
+    @Override
+    public void uninit() {
+        directoryWatcher.stop();
+    }
+
+    @Override
+    public List<SchematicPath> getList() {
```
octylFractal, line 86: I'm honestly not sure why this is a `List`? It's copying a `HashSet`, so there's absolutely no important ordering here. We should just be returning a `Set<SchematicPath>` here.

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/FileWatcherSchematicsBackend.java
```diff
@@ -0,0 +1,97 @@
...
+        lock.readLock().lock();
+        List<SchematicPath> result = new ArrayList<>(schematics);
+        lock.readLock().unlock();
+        return result;
+    }
+
+    @Override
+    public void update() {
+        // Nothing to do here, we probably already know :)
```
octylFractal, line 95: It might be worth updating the list anyways, I would hate for there to be a weird inconsistency when we could do so.

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -399,6 +403,7 @@ private static class SchematicSaveTask extends SchematicOutputTask<Void> {
         public Void call() throws Exception {
             try {
                 writeToOutputStream(new FileOutputStream(file));
+                WorldEdit.getInstance().getSchematicsManager().update();
```
octylFractal, line 406: Why aren't we passing `file` down here? There's no need to do a full rescan when we could just forcibly add the appropriate file entry.

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/PollingSchematicsBackend.java
```diff
@@ -0,0 +1,108 @@
...
+import java.nio.file.Path;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * SchematicsBackend implementation that scans the folder tree, then caches the result for a certain amount of time.
+ * This essentially is an eventually consistent cache that is used as fallback.
```
octylFractal, line 38:
```suggestion
 * A backend that scans the folder tree, then caches the result for a certain amount of time.
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/PollingSchematicsBackend.java
```diff
@@ -0,0 +1,108 @@
...
+    public void uninit() {
+    }
+
+    @Override
+    public synchronized List<SchematicPath> getList() {
+        // udpate internal cache if requried (determined by age)
+        Duration age = Duration.between(lastUpdateTs, Instant.now());
+        if (age.compareTo(MAX_RESULT_AGE) >= 0) {
+            runRescan();
+        }
```
octylFractal, line 100: I wonder if we shouldn't be doing this more proactively, using a scheduled executor service. Maybe if there's significant lag on this for someone, we can revisit.

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/PollingSchematicsBackend.java
```diff
@@ -0,0 +1,108 @@
...
+                }
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return pathList;
+    }
+
+    private void runRescan() {
+        LOGGER.debug("Rescanning Schematics");
```
octylFractal, line 81:
```suggestion
        LOGGER.debug("Rescanning schematics");
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/PollingSchematicsBackend.java
```diff
@@ -0,0 +1,108 @@
...
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(root)) {
+            for (Path path : stream) {
+                if (Files.isDirectory(path)) {
+                    pathList.addAll(scanFolder(path));
+                } else {
+                    pathList.add(new SchematicPath(path));
+                }
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
octylFractal, line 75: Use the logger.

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/PollingSchematicsBackend.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public static PollingSchematicsBackend create(Path schematicsFolder) {
+        return new PollingSchematicsBackend(schematicsFolder);
+    }
+
+    private List<SchematicPath> scanFolder(Path root) {
+        List<SchematicPath> pathList = new ArrayList<>();
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(root)) {
+            for (Path path : stream) {
+                if (Files.isDirectory(path)) {
```
octylFractal, line 68: Needs to respect allow symlinks configuration flag.

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/PollingSchematicsBackend.java
```diff
@@ -0,0 +1,108 @@
...
+     * @param schematicsFolder Root folder for schematics.
+     * @return A new PollingSchematicsBackend instance.
+     */
+    public static PollingSchematicsBackend create(Path schematicsFolder) {
+        return new PollingSchematicsBackend(schematicsFolder);
+    }
+
+    private List<SchematicPath> scanFolder(Path root) {
+        List<SchematicPath> pathList = new ArrayList<>();
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(root)) {
```
octylFractal, line 66: Honestly, maybe we should use `Files#walkFileTree` here? It'll likely be more efficient.

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/SchematicsBackend.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package com.sk89q.worldedit.util.schematic.backends;
+
+import com.sk89q.worldedit.util.schematic.SchematicPath;
+
+import java.util.List;
+
+/**
+ * SchematicsManager backend interface.
```
octylFractal, line 27:
```suggestion
 * {@link SchematicsManager} backend interface.
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/SchematicsBackend.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public interface SchematicsBackend {
+
+    /**
+     * Initialize the SchematicsManager backend.
```
octylFractal, line 32:
```suggestion
     * Initialize the backend.
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/SchematicsBackend.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    void init();
+
+    /**
+     * Uninitialize the SchematicsManager backend.
```
octylFractal, line 37:
```suggestion
     * Uninitialize the backend.
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/SchematicsBackend.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    void uninit();
+
+    /**
+     * Get the list of known schematics.
+     *
+     * @return List of known schematics.
```
octylFractal, line 44:
```suggestion
     * {@return the list of known schematics}
```

worldedit-core/src/test/java/com/sk89q/worldedit/util/schematic/SchematicPathTest.java
```diff
@@ -0,0 +1,57 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests {@link SchematicPath}.
+ *
+ * @see <a href="https://github.com/EngineHub/WorldEdit/pull/2212#discussion_r1016580671">
+ *     Trusting me4502 is good, controlling is better! :)
+ * </a>
+ */
+public class SchematicPathTest {
```
octylFractal, line 38: I don't think this test is necessary. If `Path` or `record` hashcode/equality is broken, there are bigger issues.

worldedit-core/src/main/java/com/sk89q/worldedit/util/schematic/backends/PollingSchematicsBackend.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public synchronized List<SchematicPath> getList() {
+        // udpate internal cache if requried (determined by age)
+        Duration age = Duration.between(lastUpdateTs, Instant.now());
+        if (age.compareTo(MAX_RESULT_AGE) >= 0) {
+            runRescan();
+        }
+        return new ArrayList<>(schematics);
```
octylFractal, line 101:
```suggestion
        return List.copyOf(schematics);
```

octylFractal (changes requested):
I like this a lot, but there are some edgecases to handle and cleanup to do.

Can we also move the `com.sk89q.worldedit.util.schematic` stuff to `com.sk89q.worldedit.internal.schematic`? I don't think this should be exposed as API. The `RecursiveDirectoryWatcher` should probably also be moved to `com.sk89q.worldedit.internal.util`.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/schematic/backends/FileWatcherSchematicsBackend.java
```diff
@@ -0,0 +1,97 @@
...
+                    schematics.remove(event.path());
+                }
+            } finally {
+                lock.writeLock().unlock();
+            }
+            if (event instanceof RecursiveDirectoryWatcher.FileCreatedEvent) {
+                LOGGER.debug("New Schematic found: " + event.path());
+            } else if (event instanceof RecursiveDirectoryWatcher.FileDeletedEvent) {
+                LOGGER.debug("Schematic deleted: " + event.path());
+            }
```
wizjany, line 74: why can't this just be inside the try statement, the instanceof is already being checked once.

wizjany (changes requested):

UntouchedOdin0:
Has this unfortunately been abandoned now? It seemed like a really good idea.

me4502:
Closing in favour of https://github.com/EngineHub/WorldEdit/pull/2542

PR #2213, opened 2022-11-11
Attempt to fix #2125

### LINKED ISSUE #2125: Unhandled exception executing tab-completer
### WorldEdit Version

7.2.11-beta-01

### Platform Version

Paper 1.19-31

### Confirmations

### Bug Description

I'm getting a tab-completer error when I'm typing the following command `/brush cyl "oak_slab[waterlogged=true]"`. I first noticed the error on a Purpur fork in 1.18.2 with WE 7.2.10, but proceeded to test on Paper 1.19-31 with WE 7.2.11 Beta 01. 

`org.bukkit.command.CommandException: Unhandled exception executing tab-completer for '/brush cyl "oak_slab[waterlogged=true' in com.sk89q.bukkit.util.DynamicPluginCommand(brush)`

The full error can be found here: https://paste.enginehub.org/DCFGW5MJp

### Expected Behavior

Did not expect to receive an error, but either auto-complete suggestions, or nothing at all.

### Reproduction Steps

1. Type the following `/brush cyl "oak_slab[waterlogged=true`, if the command is complete then move the cursor before the closing `]`
2. See the error in console

### Anything Else?

_No response_

seijikun:
I think the issue is that the block property completer adds two suggestions:

- One ending with `,`, to start a new property `name=value` pair.
- One ending with `]` and a **trailing space**, to signal that thats the end of the command

`onlyOnLastQuotedWord()` picks up on the trailing space, and attempts to limit the suggestion to the part behind the trailing space it. There is no part behind it, though - it's trailing after all. My fix was to simply ignore the trailing spaces.
That should be looked over by someone who knows the internals and possible sideeffects from this change better than I do, though!

Fixes #2125

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/command/CommandUtil.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/command/CommandUtil.java
@@ -227,7 +227,9 @@ public class CommandUtil {
             return suggestion;
         }
         String substr = suggestion.getSubstring();
-        int sp = substr.lastIndexOf(' ');
+        // Ignore if suggestion ends with a " ", since that signals the end of
+        // the completed command.
+        int sp = substr.trim().lastIndexOf(' ');
         if (sp < 0) {
             return suggestion;
         }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/command/CommandUtil.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/command/CommandUtil.java
@@ -227,8 +227,7 @@ public class CommandUtil {
             return suggestion;
         }
         String substr = suggestion.getSubstring();
-        // Ignore if suggestion ends with a " ", since that signals the end of
-        // the completed command.
+        // Check if there is a space inside the substring, and suggest starting from there instead.
         int sp = substr.trim().lastIndexOf(' ');
         if (sp < 0) {
             return suggestion;


worldedit-core/src/main/java/com/sk89q/worldedit/internal/command/CommandUtil.java
```diff
@@ -229,7 +229,9 @@ private static Substring onlyOnLastQuotedWord(Substring lastArg, Substring sugge
             return suggestion;
         }
         String substr = suggestion.getSubstring();
-        int sp = substr.lastIndexOf(' ');
+        // Ignore if suggestion ends with a " ", since that signals the end of
+        // the completed command.
+        int sp = substr.trim().lastIndexOf(' ');
```
me4502, line 234: This change makes the code no longer make sense. It’s no longer a check for if it ends with a space, as it no longer can due to being trimmed

seijikun: Oh okay. I probably misunderstood what this mechanism is doing, then...
I thought its purpose was remapping a suggestion such as: `oak slab` to `slab`.
Or in other words: Only suggest starting from the last space within the argument.

At least I noticed that suggestions always start from the last space, no matter whether I'm still in the same argument:

https://user-images.githubusercontent.com/820845/201366233-4611e503-e227-49e0-981f-2934d57bf408.mp4

It's also the way I had to implement it in my plugin, because the Bukkit API only allows you to suggest starting from the last space.

octylFractal: I actually think this is correct, though the comment needs adjusting.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/command/CommandUtil.java
```diff
@@ -229,7 +229,9 @@ private static Substring onlyOnLastQuotedWord(Substring lastArg, Substring sugge
             return suggestion;
         }
         String substr = suggestion.getSubstring();
-        int sp = substr.lastIndexOf(' ');
+        // Ignore if suggestion ends with a " ", since that signals the end of
+        // the completed command.
```
octylFractal, line 233:
```suggestion
        // Check if there is a space inside the substring, and suggest starting from there instead.
```

octylFractal (changes requested):

octylFractal:
Can you also rebase this onto `version/7.2.x` instead? As a bug fix, it can go there.

octylFractal (approved):

PR #2217, opened 2022-11-20
Update WorldEdit for Sponge to API 9 (1.18.2)

Lignium:
This is my attempt to port WorldEdit for Sponge to version 1.18.2 (API 9).

- [x] Fix syntax errors
- [x] Fix Gradle build
- [x] Fix `SpongeTransmogrifier`

### DIFF
--- a/worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorld.java
+++ b/worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorld.java
@@ -49,7 +49,8 @@ import com.sk89q.worldedit.world.item.ItemTypes;
 import com.sk89q.worldedit.world.weather.WeatherType;
 import com.sk89q.worldedit.world.weather.WeatherTypes;
 import net.minecraft.core.BlockPos;
-import net.minecraft.data.worldgen.Features;
+import net.minecraft.data.worldgen.features.EndFeatures;
+import net.minecraft.data.worldgen.features.TreeFeatures;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.level.LevelReader;
 import net.minecraft.world.level.levelgen.feature.ConfiguredFeature;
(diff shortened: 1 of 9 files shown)


worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/internal/SpongeWorldNativeAccess.java
```diff
@@ -105,8 +105,8 @@ public boolean updateTileEntity(BlockPos position, com.sk89q.jnbt.CompoundTag ta
         if (tileEntity == null) {
             return false;
         }
-        tileEntity.setLevelAndPosition(getWorld(), position);
```
Lignium, line 108: What is the point of calling `setLevelAndPosition` here?

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorld.java
```diff
@@ -285,43 +286,43 @@ public boolean regenerate(Region region, Extent extent, RegenOptions options) {
...
+                return TreeFeatures.SWAMP_OAK.value();
             case ACACIA:
-                return Features.ACACIA;
+                return TreeFeatures.ACACIA.value();
             case DARK_OAK:
-                return Features.DARK_OAK;
+                return TreeFeatures.DARK_OAK.value();
             case TALL_BIRCH:
-                return Features.BIRCH_TALL;
+                return TreeFeatures.SUPER_BIRCH_BEES.value(); // TODO Bees???
```
Lignium, line 315: I cannot find the exact equivalent.

worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/internal/SpongeWorldNativeAccess.java
me4502: IIRC that data was missing from the BlockEntity instance

Lignium: > IIRC that data was missing from the BlockEntity instance

In this game version, position cannot be set, only level

me4502: Is the position data available in the BlockEntity though? In the old version, that data was missing and had to be added

me4502 (commented):
This PR looks mostly good outside of the pending question; however I'm unsure how we're wanting to handle merging Sponge API 9. Is this a version that's considered "released" or near releasing? As the version/7.2.x branch is our "releases" branch, anything that goes in here should be considered something we're happy to have as a newly cut release

PR #2218, opened 2022-11-23
Fix some exceptions from SpongeWorld

Lignium:
The `BlockEntityType` registry should be requested from the `Game`.


me4502 (approved):
Has this changed within Sponge? The code as-is worked fine upon release of API 8

Lignium:
> Has this changed within Sponge? The code as-is worked fine upon release of API 8 (IIRC we actually had to use the world's registry at that point?)

Perhaps it was in snapshots, but not in the release.
From SpongeAPI 8.1.0 sources:
```java
@RegistryScopes(scopes = RegistryScope.GAME)
public final class BlockEntityTypes { ... }
```

Issue #2132, opened 2022-06-30
Set entire column with biome brush


ItzRonaldMLG:
### The Problem

I have searched hours but i dont think this is a feature:
I tried to change the biome, but its in 3D, meaning it wont change from top to bottom of the world.
Is there a way to use a brush to both yaxis aswell up and down simultaneously? like ''-100:100''?

### A Solution

something like dividing with ":"

octylFractal:
I'm not really sure what you're asking for, since AFAIK there isn't a biome brush, so there isn't anything to apply this to?

octylFractal:
Apparently I forgot about our biome brush. I don't think we need anything but a `-c` flag to set the whole column with the brush though, unless there is a compelling use-case for only specific y-offsets.

ItzRonaldMLG:
Atm im using ''/brush cylinder $deep_ocean 5 100'', but the 100 will only apply from block to 100 blocks above it, and -100 will of course only do the block and 100 below.
So if you stand in the water you are in biome ''deep ocean'' but if you fly or just even jump around, it will still be the previous biome.

If this explains it better. 
Or maybe use ''vert'' or ''expand-vert'' to apply it from bottom to top?

octylFractal:
That syntax doesn't exist? Are you using a fork despite checking the box saying you aren't?

ItzRonaldMLG:
> That syntax doesn't exist? Are you using a fork despite checking the box saying you aren't?

I cant use it in the command ''/brush cylinder $birch_forest 30 vert'' or expand-vert does not work no

octylFractal:
No, it is not. Please don't describe syntax from other plugins. Leaving this open for the feature of adding columns to the biome brush, but in the future do not create issues when not using WorldEdit.

me4502:
This was closed in https://github.com/EngineHub/WorldEdit/commit/856aaa62670b4cd2f2cc2fbb721bee16e4a49098

PR #2235, opened 2022-12-28
Add a -c flag to the biome brush to set entire column

me4502:
Adds a `-c` flag to the biome brush, to apply to the entire column

Adds https://github.com/EngineHub/WorldEdit/issues/2132


octylFractal (changes requested):
I don't like the design of this, because it doesn't _actually set the whole column_, instead hacking around it with a large shape. If this brush is used near the top or bottom of a world, it'll produce the wrong results. The factories should be taking exact minY/maxY positions to be the full column, maybe even as a separate factory type.

PR #2237, opened 2022-12-28
Add method to check initialize state in PlatformManager, deny access to registries when uninitialized

JOO200:
This should fix a bug, where worldedit types such as BlockTypes.AIR will be null when some plugin (or mod) try to access WorldEdit too early (e.g. access BlockTypes.AIR in onLoad).

I added a method to get the current initialization state in PlatformManager for that. To check that state I added some check to the UnitTests.

The static code block in the Types classes are not the cleanest approach but adding some code in Registry classes might be a breaking change because some other plugins may use them.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
@@ -298,8 +298,22 @@ public class PlatformManager {
         return queryCapability(Capability.CONFIGURATION).getConfiguration();
     }
 
+    /**
+     * Get the current supported {@link SideEffect}s.
+     *
+     * @return the supported side effects
+     * @throws NoCapablePlatformException thrown if no platform is capable
+     */
     public Collection<SideEffect> getSupportedSideEffects() {
         return queryCapability(Capability.WORLD_EDITING).getSupportedSideEffects();
+    }
+
+    /**
+     * Get the initialized state of the Platform.
+     * {@return if the platform manager is initialized}
+     */
+    public boolean isInitialized() {
+        return initialized.get();
     }
 
     /**
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/registry/Registry.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/registry/Registry.java
@@ -19,6 +19,8 @@
 
 package com.sk89q.worldedit.registry;
 
+import com.sk89q.worldedit.WorldEdit;
+
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -34,9 +36,15 @@ import static java.util.Objects.requireNonNull;
 public class Registry<V extends Keyed> implements Iterable<V> {
     private final Map<String, V> map = new HashMap<>();
     private final String name;
+    private final boolean checkInitialized;
 
     public Registry(final String name) {
+        this(name, false);
+    }
+
+    public Registry(final String name, final boolean checkInitialized) {
         this.name = name;
+        this.checkInitialized = checkInitialized;
     }
 
     public String getName() {
@@ -46,6 +54,10 @@ public class Registry<V extends Keyed> implements Iterable<V> {
     @Nullable
     public V get(final String key) {
         checkState(key.equals(key.toLowerCase(Locale.ROOT)), "key must be lowercase: %s", key);
+        if (this.checkInitialized) {
+            checkState(WorldEdit.getInstance().getPlatformManager().isInitialized(),
+                    "WorldEdit is not enabled yet.");
+        }
         return this.map.get(key);
     }
 
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/registry/Registry.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/registry/Registry.java
@@ -56,7 +56,7 @@ public class Registry<V extends Keyed> implements Iterable<V> {
         checkState(key.equals(key.toLowerCase(Locale.ROOT)), "key must be lowercase: %s", key);
         if (this.checkInitialized) {
             checkState(WorldEdit.getInstance().getPlatformManager().isInitialized(),
-                    "WorldEdit is not enabled yet.");
+                    "WorldEdit is not initialized yet.");
         }
         return this.map.get(key);
     }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/registry/NamespacedRegistry.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/registry/NamespacedRegistry.java
@@ -19,6 +19,8 @@
 
 package com.sk89q.worldedit.registry;
 
+import com.sk89q.worldedit.WorldEdit;
+
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
@@ -36,8 +38,16 @@ public final class NamespacedRegistry<V extends Keyed> extends Registry<V> {
         this(name, MINECRAFT_NAMESPACE);
     }
 
+    public NamespacedRegistry(final String name, final boolean checkInitialized) {
+        this(name, MINECRAFT_NAMESPACE, checkInitialized);
+    }
+
     public NamespacedRegistry(final String name, final String defaultNamespace) {
-        super(name);
+        this(name, defaultNamespace, false);
+    }
+
+    public NamespacedRegistry(final String name, final String defaultNamespace, final boolean checkInitialized) {
+        super(name, checkInitialized);
         this.defaultNamespace = defaultNamespace;
     }
 
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/biome/BiomeType.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/biome/BiomeType.java
@@ -29,7 +29,7 @@ import com.sk89q.worldedit.registry.NamespacedRegistry;
  */
 public class BiomeType implements Keyed, BiomePattern {
 
-    public static final NamespacedRegistry<BiomeType> REGISTRY = new NamespacedRegistry<>("biome type");
+    public static final NamespacedRegistry<BiomeType> REGISTRY = new NamespacedRegistry<>("biome type", true);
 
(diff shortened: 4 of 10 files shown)


me4502:
IMO this is a decent fix until WE8 where we can break the registry accesses given it at least makes the issue less cryptic. Has this been tested with a plugin trying to access it too early? (and with a plugin accessing it at a normal time etc, to make sure it doesn't break normal behaviour?)

JOO200:
I tested it with a simple Bukkit plugin which depends on WorldEdit:
```java
public class CrashPlugin extends JavaPlugin {
    @Override
    public void onLoad() {
        // getLogger().info(BlockTypes.AIR.getId());
    }

    @Override
    public void onEnable() {
        getLogger().info(BlockTypes.AIR.getId());
    }
}
```
If i try to access BlockTypes.AIR in onLoad the following error will be displayed:
```
[11:58:22 INFO]: [WorldEditCrasher] Loading WorldEditCrasher vSERVER-SNAPSHOT
[11:58:22 ERROR]: [org.bukkit.craftbukkit.v1_19_R2.CraftServer] null initializing WorldEditCrasher vSERVER-SNAPSHOT (Is it up to date?)
java.lang.ExceptionInInitializerError: null
	at de.terraconia.worldedit.CrashPlugin.onLoad(CrashPlugin.java:9) ~[WorldEditCrasher-SERVER-SNAPSHOT.jar:?]
	at org.bukkit.craftbukkit.v1_19_R2.CraftServer.loadPlugins(CraftServer.java:442) ~[paper-1.19.3.jar:git-Paper-362]
	at net.minecraft.server.dedicated.DedicatedServer.initServer(DedicatedServer.java:273) ~[paper-1.19.3.jar:git-Paper-362]
	at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1101) ~[paper-1.19.3.jar:git-Paper-362]
	at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:316) ~[paper-1.19.3.jar:git-Paper-362]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
Caused by: java.lang.IllegalStateException: Tried to access WorldEdit while not initialized.
	at com.sk89q.worldedit.world.block.BlockTypes.<clinit>(BlockTypes.java:975) ~[worldedit-bukkit-7.2.14-SNAPSHOT-dist.jar:?]
	... 6 more
[11:58:22 INFO]: Server permissions file permissions.yml is empty, ignoring it
[11:58:22 INFO]: [WorldEdit] Enabling WorldEdit v7.2.14-SNAPSHOT+84ebc65
[11:58:22 INFO]: Registering commands with com.sk89q.worldedit.bukkit.BukkitServerInterface
[11:58:22 INFO]: WEPIF: Using the Bukkit Permissions API.
[11:58:22 INFO]: Using com.sk89q.worldedit.bukkit.adapter.impl.v1_19_R2.PaperweightAdapter as the Bukkit adapter
[11:58:22 ERROR]: Error occurred while enabling WorldEdit v7.2.14-SNAPSHOT+84ebc65 (Is it up to date?)
java.lang.NoClassDefFoundError: Could not initialize class com.sk89q.worldedit.world.block.BlockTypes
	at com.sk89q.worldedit.LocalConfiguration.getDefaultDisallowedBlocks(LocalConfiguration.java:98) ~[worldedit-bukkit-7.2.14-SNAPSHOT-dist.jar:?]
	at com.sk89q.worldedit.util.YAMLConfiguration.load(YAMLConfiguration.java:84) ~[worldedit-bukkit-7.2.14-SNAPSHOT-dist.jar:?]
	at com.sk89q.worldedit.bukkit.BukkitConfiguration.load(BukkitConfiguration.java:48) ~[worldedit-bukkit-7.2.14-SNAPSHOT-dist.jar:?]
	at com.sk89q.worldedit.bukkit.WorldEditPlugin.setupPreWorldData(WorldEditPlugin.java:197) ~[worldedit-bukkit-7.2.14-SNAPSHOT-dist.jar:?]
	at com.sk89q.worldedit.bukkit.WorldEditPlugin.onEnable(WorldEditPlugin.java:177) ~[worldedit-bukkit-7.2.14-SNAPSHOT-dist.jar:?]
	at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:264) ~[paper-api-1.19.3-R0.1-SNAPSHOT.jar:?]
	at org.bukkit.plugin.java.JavaPluginLoader.enablePlugin(JavaPluginLoader.java:371) ~[paper-api-1.19.3-R0.1-SNAPSHOT.jar:?]
	at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:544) ~[paper-api-1.19.3-R0.1-SNAPSHOT.jar:?]
	at org.bukkit.craftbukkit.v1_19_R2.CraftServer.enablePlugin(CraftServer.java:578) ~[paper-1.19.3.jar:git-Paper-362]
	at org.bukkit.craftbukkit.v1_19_R2.CraftServer.enablePlugins(CraftServer.java:492) ~[paper-1.19.3.jar:git-Paper-362]
	at net.minecraft.server.dedicated.DedicatedServer.initServer(DedicatedServer.java:274) ~[paper-1.19.3.jar:git-Paper-362]
	at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1101) ~[paper-1.19.3.jar:git-Paper-362]
	at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:316) ~[paper-1.19.3.jar:git-Paper-362]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
```

Unfortunatly WorldEdit will not be able to recover from this state and WorldEdit (and other plugins such as WorldGuard) will not start correctly. I don't think there is an easy fix for that without other breaking changes.

Accessing BlockTypes.AIR in onEnable works fine and there is no Exception:
```
[12:00:49 INFO]: [WorldEditCrasher] Enabling WorldEditCrasher vSERVER-SNAPSHOT
[12:00:49 INFO]: [WorldEditCrasher] minecraft:air
```

me4502 (approved):

octylFractal (approved):
I wonder if it might be more prudent to put this in the `Registry` class initializer, so we don't miss it on any of these types of classes.

JOO200:
I considered Registry as public api, adding such a change to that ctor is partially a breaking change.

wizjany:
maybe we can add an `AbstractWorldEditRegistry`-kinda intermediate class with this logic

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
```diff
@@ -298,10 +298,26 @@ public LocalConfiguration getConfiguration() {
...
+     *
+     * @return the supported side effects
+     */
     public Collection<SideEffect> getSupportedSideEffects() {
         return queryCapability(Capability.WORLD_EDITING).getSupportedSideEffects();
     }
 
+    /**
+     * Get the initialized state of the Platform.
+     * @return whether the platform is initialized
```
octylFractal, line 315:
```suggestion
     * {@return if the platform manager is initialized}
```

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
```diff
@@ -298,10 +298,26 @@ public LocalConfiguration getConfiguration() {
         return queryCapability(Capability.CONFIGURATION).getConfiguration();
     }
 
+    /**
+     * Get the current supported {@link SideEffect}s.
+     *
+     * <p>If no platform has been registered yet, then a fallback value
+     * will be returned.</p>
```
octylFractal, line 305: Isn't this wrong? Won't this _throw_ if no platforms are registered?

worldedit-core/src/main/java/com/sk89q/worldedit/world/biome/BiomeTypes.java
```diff
@@ -133,6 +135,12 @@ public final class BiomeTypes {
     private BiomeTypes() {
     }
 
+    static {
+        if (!WorldEdit.getInstance().getPlatformManager().isInitialized()) {
+            throw new IllegalStateException("Tried to access WorldEdit while not initialized.");
+        }
```
octylFractal, line 141: Can we move this whole thing to a `PlatformManager#checkInitialized()`? That way we're not repeating the message across every instance, in case we edit it in the future.

octylFractal (changes requested):
On a more thorough reading I found a few changes I would like.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
JOO200: yes, that was wrong. I fixed the wrong javadoc comment

worldedit-core/src/main/java/com/sk89q/worldedit/world/biome/BiomeTypes.java
JOO200: I added some logic to Registry to allow that check in the Registry.

JOO200:
I moved the logic to the Registry class and made an overloaded constructor with a boolean argument for the check.

Other way is to add some WorldEditRegistry implementation (between NamespacedRegistry and Registry). Feedback is appreciated.

worldedit-core/src/main/java/com/sk89q/worldedit/registry/Registry.java
```diff
@@ -46,6 +54,10 @@ public String getName() {
     @Nullable
     public V get(final String key) {
         checkState(key.equals(key.toLowerCase(Locale.ROOT)), "key must be lowercase: %s", key);
+        if (this.checkInitialized) {
+            checkState(WorldEdit.getInstance().getPlatformManager().isInitialized(),
+                    "WorldEdit is not enabled yet.");
```
octylFractal, line 59:
```suggestion
                    "WorldEdit is not initialized yet.");
```
I would prefer consistent wording here, especially because "enabled" is a Bukkit-specific term.

octylFractal (changes requested):

octylFractal (approved):

PR #2239, opened 2022-12-31
Add a feature generator and allow undoing of feature placement

me4502:
This PR adds feature generation via a command. The same limitations of the vanilla feature gen apply, where it has strict requirements about placement that are rather opaque and not easily avoided.

This PR also allows the feature gen to be reverted with undo, which means from a baseline featureset perspective we're ahead of the vanilla version of this. As it's passed through an EditSession, we also get all masking/perf settings/history/etc that come with WorldEdit for free, also for trees which we didn't have prior (outside of Bukkit).

Also includes a brush variant

TODO:
* [x] Fabric
* [x] Forge
* [x] Bukkit
* [ ] Sponge(?)
* [x] Ensure this works properly with obfuscation

https://user-images.githubusercontent.com/546754/210136260-07e1aa98-3bc0-4ec2-b59c-62c3637a30e8.mp4

https://user-images.githubusercontent.com/546754/210137698-ae849d18-225e-4c6f-9c2e-8953f70a0ca2.mp4


worldedit-bukkit/adapters/adapter-1.19.3/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_19_R2/PaperweightServerLevelDelegateProxy.java
```diff
@@ -0,0 +1,120 @@
...
+                    // getBlockEntity
+                    return getBlockEntity(blockPos);
+                }
+            }
+            case "a", "setBlock", "removeBlock" -> {
+                if (args.length >= 2 && args[0] instanceof BlockPos blockPos && args[1] instanceof BlockState blockState) {
+                    // setBlock
+                    return setBlock(blockPos, blockState);
+                } else if (args.length >= 2 && args[0] instanceof BlockPos blockPos && args[1] instanceof Boolean bl) {
+                    // removeBlock (and also matches destroyBlock)
```
octylFractal, line 110: ... on purpose, or accidentally? if on purpose, shouldn't `destroyBlock` be in the `case`?

worldedit-bukkit/adapters/adapter-1.19.3/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_19_R2/PaperweightServerLevelDelegateProxy.java
```diff
@@ -0,0 +1,120 @@
...
+        try {
+            return editSession.setBlock(BlockVector3.at(blockPos.getX(), blockPos.getY(), blockPos.getZ()), BlockTypes.AIR.getDefaultState());
+        } catch (MaxChangedBlocksException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+        switch (method.getName()) {
```
octylFractal, line 92: Idea:
> initialize a `Map<Method, Function<Object[], Object>>`, use `Refraction.pickName` to choose the correct names, then just do a map lookup

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -245,6 +245,8 @@
     "worldedit.sphere.created": "{0} blocks have been created.",
     "worldedit.forestgen.created": "{0} trees created.",
     "worldedit.pumpkins.created": "{0} pumpkin patches created.",
+    "worldedit.feature.created": "Feature created.",
+    "worldedit.feature.failed": "Failed to generate feature. Is it a valid spot for it?",
```
octylFractal, line 249:
```suggestion
    "worldedit.feature.failed": "Failed to generate feature. Is there a valid spot for it?",
```

octylFractal (changes requested):

worldedit-bukkit/adapters/adapter-1.19.3/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_19_R2/PaperweightServerLevelDelegateProxy.java
me4502: I'd missed it for the de-obfuscated names. For obfuscated names, they are both `"a"`

worldedit-bukkit/adapters/adapter-1.19.3/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_19_R2/PaperweightServerLevelDelegateProxy.java
me4502: IMO this would make more sense to do once we have unified mappings, as currently that class is within WE-Bukkit & isn't really that much cleaner

DereC4:
thanks! Didn't realize my issue from high school got made into a real feature!

PR #2240, opened 2023-01-01
Improving the Save Feature: Suggestions for Enhanced Functionality and Performance

miguelgargallo:
# Improving the Save Feature: Suggestions for Enhanced Functionality and Performance

### The suggested changes to the save feature could be considered essential for the following reasons:

- Allow user to specify custom save location
- Allow user to specify which changes to save
- Implement dirty check to improve performance

Here are the suggested changes to the save feature, enumerated and presented in the format of an enumerated list:

Allow the user to specify the file path where the save file should be saved, instead of always saving to the default location.

Allow the user to specify which modifications should be saved, by adding a list of "Change" objects as a parameter to the save() method. The Change class could represent an individual change made to the world. This would enable the user to save only a specific set of changes, rather than all changes.

Optimize the save process by adding a "dirty check" method. This method could compare the current state of the world to the saved state, and only perform a save if the two states are different. This would prevent unnecessary save operations and improve the performance of the code.

Custom save locations allow the user to specify where they would like to save their files, which can be useful for organization and convenience. This is especially helpful when the user has limited access to the default save location.

The ability to selectively save changes gives the user more control over their saved data. They can choose to save changes incrementally rather than all at once, and can even revert to previous versions of the world. This can be useful for undoing accidental changes or for saving only the most important changes.

A dirty check is a process that improves the performance and efficiency of the code by avoiding unnecessary save operations. It checks whether there have been any changes made to the data that need to be saved, and only performs the save operation if necessary. This can save resources and prevent the code from slowing down due to excessive saving.

Overall, custom save locations and the ability to selectively save changes give the user more control and organization over their saved data, while a dirty check helps to ensure that the save feature does not consume too many resources or become a bottleneck in the code. This can improve the user experience and make the save feature faster and more efficient.

### DIFF
--- a/worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorld.java
+++ b/worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIWorld.java
@@ -2,6 +2,7 @@
  * WorldEdit, a Minecraft world manipulation toolkit
  * Copyright (C) sk89q <http://www.sk89q.com>
  * Copyright (C) WorldEdit team and contributors
+ * Copyright (C) miguelgargallo to Optimize and improve the code
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


octylFractal:
This PR has multiple formatting issues and doesn't change any implementation of this interface. Are you intending to write the implementation changes? If not, we will likely close this PR. Proposals to change things without corresponding implementation changes should be made as issues, not PRs.

me4502:
Also don’t just add yourself to the copyright header

miguelgargallo:
> Also don’t just add yourself to the copyright header

Understood, Happy New Year

PR #2245, opened 2023-01-09
EditSession.drawSpline: use vectors at block center for spline nodes

cthbleachbit:
The spline generator was fed integral block node vectors, where center of a block at coordinate (X,Y,Z) is actually in (X+0.5,Y+0.5,Z+0.5). When generating a spline with (X,Y,Z)s, the curve sometimes doesn't go through the intended block and may make weird turns at end of the curve.

Placing spline nodes at the center causes the curve to actually hit the center of every block selected.

Here is a problematic curve. Nodes are marked in circles. Note that the entire curve is at the very least supposed to look symmetric along the secondary diagonal direction, but upper-left corner looks significantly messed up compared to the lower-right corner.
截图 2023-01-09 01-05-13

The curve generated along the same rig with same nodes after the fix:
截图 2023-01-09 01-24-38

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -2588,7 +2588,7 @@ public class EditSession implements Extent, AutoCloseable {
         Interpolation interpol = new KochanekBartelsInterpolation();
 
         for (BlockVector3 nodevector : nodevectors) {
-            Node n = new Node(nodevector.toVector3());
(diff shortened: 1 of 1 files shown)


me4502 (approved):
This makes sense to me, although I'm not familiar with the spline node code here at all so I'll also let someone else review.

cthbleachbit:
It's been a while - any updates on this?

octylFractal:
I've been tired so haven't been able to review any PRs recently. I am not sure when that's going to let up, but it will be reviewed eventually.

octylFractal (approved):
I'm not totally sure about this, but I think it makes sense. Generally offsetting by a half-block is necessary for rounding to work properly everywhere...

PR #2248, opened 2023-01-15
Few cleanups over the code and use more Java 9+ features

me4502:
Did a bunch of code cleanups by looking through files. Mostly migrating to Java 9+ features. `.toList()` rather than `.collect(Collectors.toList())` is apparently a lot more performant (not that we use it in hot spots anyway).

Also deleted two unused classes in the internal package


worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/DeprecationUtil.java
```diff
@@ -44,7 +43,7 @@ private DeprecationUtil() {
      */
     public static void checkDelegatingOverride(Class<?> implementingClass) {
         // pull the information about the caller
-        StackTraceElement caller = Throwables.lazyStackTrace(new Throwable()).get(1);
```
SirYwell, line 47: This and the change in `LogManagerCompat` could make use of `StackWalker` instead

worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIBlockRegistry.java
```diff
@@ -33,30 +33,28 @@
...
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 public class CLIBlockRegistry extends BundledBlockRegistry {
 
     private Property<?> createProperty(String type, String key, List<String> values) {
         switch (type) {
-            case "int": {
```
SirYwell, line 43: This could be a switch expression aswell

worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/CommandsManagerRegistration.java
```diff
@@ -32,7 +32,6 @@
 import java.util.Map;
 
 @Deprecated
-@SuppressWarnings("deprecation")
```
octylFractal, line 35: IIRC there was a warning in `javac` suppressed by this. Is this no longer necessary?

worldedit-core/src/main/java/com/sk89q/jnbt/CompoundTag.java
```diff
@@ -61,7 +61,7 @@ public boolean containsKey(String key) {
         return linTag.value().containsKey(key);
     }
 
-    @SuppressWarnings({ "unchecked", "rawtypes" })
```
octylFractal, line 64: IIRC I needed this for either IntelliJ or javac.

worldedit-core/src/main/java/com/sk89q/worldedit/command/util/SuggestionHelper.java
```diff
@@ -88,26 +88,23 @@ public static Stream<String> getBlockPropertySuggestions(String blockType, Strin
...
-                        default:
-                            return matchingProps.stream().map(p -> lastValidInput + p.getName() + "=");
-                    }
+                            .map(Map.Entry::getValue).toList();
+                    return switch (matchingProps.size()) {
+                        case 0 -> propertyMap.keySet().stream().filter(p -> !matchedProperties.contains(p)).map(prop ->
+                            lastValidInput + prop + "=");
+                        case 1 -> matchingProps.get(0).getValues().stream().map(val ->
+                            lastValidInput + matchingProps.get(0).getName() + "="
+                                + val.toString().toLowerCase(Locale.ROOT));
```
octylFractal, line 97: Can you format these lambdas a little better, it's kinda ugly. Maybe:
```suggestion
                        case 0 -> propertyMap.keySet().stream()
                            .filter(p -> !matchedProperties.contains(p))
                            .map(p -> lastValidInput + p + "=");
                        case 1 -> matchingProps.get(0).getValues().stream()
                            .map(val ->
                                lastValidInput + matchingProps.get(0).getName() + "="
                                    + val.toString().toLowerCase(Locale.ROOT)
                            );
```

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -163,58 +163,56 @@ private String woolMapper(String string) {
...
                         ));
+                    } else {
+                        WorldEdit.logger.debug("Unknown property " + parts[0] + " for block " + type.getId());
                     }
-
-                    blockStates.put(propertyKey, value);
-                } catch (InputParseException e) {
-                    throw e; // Pass-through
-                } catch (Exception e) {
+                    return Maps.newHashMap();
```
octylFractal, line 189: Change to `Map.of()` while we're here?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -163,58 +163,56 @@ private String woolMapper(String string) {
     private static Map<Property<?>, Object> parseProperties(BlockType type, String[] stateProperties, ParserContext context) throws InputParseException {
         Map<Property<?>, Object> blockStates = new HashMap<>();
```
octylFractal, line 164: Maybe change this to use `ImmutableMap`?

worldedit-core/src/main/java/com/sk89q/worldedit/internal/Constants.java
```diff
@@ -35,11 +33,11 @@ private Constants() {
     public static final List<String> NO_COPY_ENTITY_NBT_FIELDS;
 
     static {
-        NO_COPY_ENTITY_NBT_FIELDS = Collections.unmodifiableList(Arrays.asList(
-                "UUIDLeast", "UUIDMost", "UUID", // Bukkit and Vanilla
-                "WorldUUIDLeast", "WorldUUIDMost", // Bukkit and Vanilla
-                "PersistentIDMSB", "PersistentIDLSB" // Forge
-        ));
+        NO_COPY_ENTITY_NBT_FIELDS = List.of(
```
octylFractal, line 36: I wonder if we should change this to a `Set` and use `tag.getAllKeys().removeAll(Constants.NO_COPY_ENTITY_NBT_FIELDS);` at usage sites.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/DeprecationUtil.java
octylFractal: Yes, we should 100% be doing that instead.

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorldEdit.java
```diff
@@ -172,7 +172,6 @@ private void setupPlatform() {
     }
 
     // TODO clean this up once Forge adds a proper API for this
```
octylFractal, line 174: You might not need the `TODO` here as well. I think it was related to the deprecations.

octylFractal (changes requested):

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorldEdit.java
me4502: I believe this TODO was for a proper registry event wasn't it? So that we'd be aware if a registry change happened later or something

worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/CommandsManagerRegistration.java
me4502: None appear to show - but AFAIK classes marked with `@Deprecated` do not show deprecation warnings anyway?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
me4502: I've made it wrap on return - can't use a builder as this does a `containsKey` check which builders don't support

worldedit-core/src/main/java/com/sk89q/worldedit/internal/Constants.java
me4502: I feel this would be better for a followup PR

worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorldEdit.java
octylFractal: I would just remove the comment for now.

octylFractal (approved):

PR #2249, opened 2023-01-15
Allow finding schematic format by InputStream

me4502:
Adds methods to find the format of a schematic by an InputStream, rather than a file. Given all the loading etc functions use InputStream, this makes sense to include.

This allows situations that don't have a proper file to still determine the format, such as situations where a schematic is downloaded from the internet or loaded from a zip file.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormat.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormat.java
@@ -20,12 +20,12 @@
 package com.sk89q.worldedit.extent.clipboard.io;
 
 import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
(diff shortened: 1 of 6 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/ClipboardFormat.java
```diff
@@ -68,7 +70,23 @@ public interface ClipboardFormat {
...
+        }
+    }
+
+    /**
+     * Return whether the given stream is of this format.
+     *
+     * @param inputStream The stream
+     * @return true if the given stream is of this format
+     */
+    default boolean isFormat(InputStream inputStream) {
```
octylFractal, line 87: This method should not close the stream in implementors, and it should be documented that the caller is responsible for closing it.

octylFractal (changes requested):

octylFractal (approved):

PR #2253, opened 2023-01-26
Allow custom biomes in bukkit

Brokkonaut:
More and more servers use data packs with custom biomes, so i think it would be nice if worldedit supports them.


me4502:
Due to the changes in this PR, it would make sense for it to target master rather than the release branch.

Also, [this PR here](https://github.com/EngineHub/WorldEdit/pull/2239/files#diff-e41f7785f633e49ffffcf35ef6c4fac6bf0dfdff23801637c1043b47edfaf39bR853) already adds a way to add registry values from within adapters on the Bukkit WE platform. It'd likely make sense for this PR to use that method added in that PR. It's just waiting on octy to review it

Brokkonaut:
I would prefer to have it in 7.2.x, because it does not change any api or so, it just allows that feature on the bukkit platform that is already available on other platforms. but if that is not possible, i can retarget it.

for the way how the data from the registry is loaded, yes i can change that

octylFractal (approved):
Looks fine, should be merged after #2239

octylFractal:
Oh, that's master only. Hmmm. I guess we can resolve the conflict manually.

PR #2275, opened 2023-03-06
Add optional argument to //toggleplace

TomyLobo:
By invoking either `//toggleplace pos1` or `//toggleplace player`, the player can select either pos1 or player explicitly.
Without the argument, the command behaves as it always has and toggles between the two states


TomyLobo:
Turned this into a draft PR, since these will probably get separated out into a `//placement` command in #2280, so adding the new syntax only to then remove it makes no sense.

TomyLobo:
rebased

TomyLobo:
deleted the branch and closed the PR. See #2280.

PR #2276, opened 2023-03-08
Add clipboard support to //deform and //brush deform

TomyLobo:
This adds a -l flag to `//deform` and `//brush deform` which, when used, will fetch blocks from the clipboard instead of the world.
Additionally, coordinate transformations that would normally be based on the selection position+size are based on the clipboard's position+size when using that mode.

Example usage:
image
The shape on the right is the result of copying the square shape on the left and deforming the selection with the command on screen.

image
This is the following brush using the same clipboard contents as before:
`//brush deform -l sphere 6 x=-x;y=0;z=-z`


worldedit-core/src/main/java/com/sk89q/worldedit/function/factory/Deform.java
```diff
@@ -64,6 +70,10 @@ public Deform(Extent destination, Region region, String expression) {
     }
 
     public Deform(Extent destination, Region region, String expression, Mode mode) {
+        this(destination, region, expression, mode, false);
+    }
+
+    public Deform(Extent destination, Region region, String expression, Mode mode, boolean useClipboard) {
```
me4502, line 76: I sort of feel a sourceExtent & sourceRegion would be more useful here as something to pass in rather than useClipboard. Would also allow deform brushes to "bake in" the clipboard so you can have multiple different ones

TomyLobo:
rebased

worldedit-core/src/main/java/com/sk89q/worldedit/function/factory/Deform.java
TomyLobo: Sorry, I only just saw your comment.
We'd have to offer both, because some people would want the current clipboard, some wouldn't.
Same with the -o flag.
That'd probably a separate PR.

wizjany: baking in is the default for most things and should probably be consistent for ux. (see clipboard brush, `#sel` vs `#dsel` masks, etc)

TomyLobo: ok I wasn't aware `//brush clipboard` already baked in the clipboard.
I use brushes so rarely :)
guess I'll have to bake that in before merging it then

TomyLobo: `-o`, on the other hand, is not so simple.
1. Do we read placement type at `//brush` time or at click time?
2. Do we read `pos1` at `//brush` time or at click time?
3. Do we read the clipboard at `//brush` time or at click time?
4. Do we read `player` position at `//brush` time or at click time?

1. Probably at `//brush` time 
2. That'd be the difference between `#sel` and `#dsel`, so we'd probably want to offer both
3. We said we want to bake it in at `//brush` time, but if we offer both in 2., offering both here as well actually saves code
4. I have no idea tbh

TomyLobo:
rebased this on top of most of the //placement PR in order to be able to use the new Placement system for the Deform brush.
Also marked as draft because the Current code is a WIP

TomyLobo:
Moved the (contentious and not yet rebased) `//brush deform` changes into a separate `deform-brush-clipboard` branch and rebased again.
I'll unmark this as a draft once I'm done testing if it still works.

I might rebase `deform-brush-clipboard` later, but that's somewhat of a failed implementation and probably should be redone from scratch anyway.

worldedit-core/src/main/java/com/sk89q/worldedit/math/transform/SimpleTransform.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+package com.sk89q.worldedit.math.transform;
+
+import com.sk89q.worldedit.math.Vector3;
+
+/**
+ * A more light-weight {@link Transform} than {@link AffineTransform}, supporting only translation and scaling.
+ */
+public class SimpleTransform implements Transform {
```
me4502, line 27: Is SimpleTransform able to please be a record? it might allow the JVM to optimise `isIdentity` to a computed constant

TomyLobo: actually already had this turned into a record locally :)

TomyLobo:
Checks are passing, deform-brush-clipboard pushed
deform-brush-clipboard is untested.
deform-clipboard (this PR) is tested and working well, except for one issue:
for flat clipboards (i.e. clipboards which are only 1 thick on any axis) non-flat selections map some points outside the clipboard.
A workaround for the user is to set the flat axis to 0.

Two alternatives were considered:
option B: adjust only when inverting
advantage: `//deform -l 1` fills the entire selection even if the clipboard only is 1 thick
disadvantage(?): any coordinate on the axis in which the clipboard only is 1 thick will yield the same raw coordinate, meaning it's impossible to go outside the box on that axis anymore, with or without `-l`
disadvantage: the transform's implied matrix is singular if the clipboard only is 1 thick
disadvantage: the resulting transform is technically no longer really its own inverse (though `transform.inverse().inverse() == transform` still holds true)

option C: add 0.5 to the region/clipboard's min/max
advantage: no singularities, continuous, no special cases
disadvantage: breaks people's existing expressions for `//generate`,  `//generatebiome`, `//deform` and `//brush deform`
disadvantage: makes the relationship between block coordinates and -1..1 coordinates more complex
disadvantage: coordinates never reach -1 or 1, breaking expressions that rely on that

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -2419,31 +2500,47 @@ public int deformRegion(final Region region, final Vector3 zero, final Vector3 u
...
 
         final DoubleArrayList<BlockVector3, BaseBlock> queue = new DoubleArrayList<>(false);
 
+        final Transform outputTransformInverse = outputTransform.inverse();
         for (BlockVector3 position : region) {
-            // offset, scale
-            final Vector3 scaled = position.toVector3().subtract(zero).divide(unit);
-
             // transform
+            final Vector3 scaled = outputTransformInverse.apply(position.toVector3());
```
me4502, line 2534: is `scaled` the correct name for this field now, if it's also translated? would `transformedPosition` be a more descriptive name?

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -2419,31 +2500,47 @@ public int deformRegion(final Region region, final Vector3 zero, final Vector3 u
...
+            // deform
             expression.evaluate(new double[]{ scaled.x(), scaled.y(), scaled.z() }, timeout);
 
-            final BlockVector3 sourcePosition = environment.toWorld(x.value(), y.value(), z.value());
+            // untransform, round-nearest
+            final BlockVector3 sourcePosition = inputTransform.apply(Vector3.at(x.value(), y.value(), z.value())).add(0.5, 0.5, 0.5).toBlockPoint();
 
-            // read block from world
-            final BaseBlock material = world.getFullBlock(sourcePosition);
+            // read block from world/clipboard
```
me4502, line 2542: this should probably just say extent, given there could theoretically be other extents

TomyLobo:
Now that #2582 is merged, I rebased and force-pushed this and #2580
Will test now, but I don't expect any issues, unless the heat is getting to me more than I thought :)

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
TomyLobo: It's called `inputPosition` now and after the expression is evaluated, I introduced an intermediate variable called `outputPosition` to match it.
Resolving conversation

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
TomyLobo: it says "read block from source extent (e.g. world/clipboard)" now.
Resolving conversation

TomyLobo:
I'm done testing and enabled auto-merge.
So once someone reviews this, it's in :)

me4502 (commented):
This looks good to me, but I'd prefer if someone else reviewed this too as I'm less familiar with the expression/deform system

TomyLobo:
just [force-pushed](https://github.com/EngineHub/WorldEdit/compare/ef8b77b0f3ab470a8981a0ebde525744fef7ebfa..11a442591cd4b04f68bdb7db466c780e0b11cd9d) to change a javadoc description from input/output to source/target

TomyLobo:
[force-pushed](https://github.com/EngineHub/WorldEdit/compare/11a442591cd4b04f68bdb7db466c780e0b11cd9d..2e0ecdb02be70ecaa0911aca1901409276783d7e) to rename SimpleTransform to ScaleAndTranslateTransform, as suggested by @wizjany.

TomyLobo:
Any objections to merging this?

octylFractal:
I'd like to give it a review, but I can't immediately because I'm working.

TomyLobo:
this is still done and just pending a review :)

me4502 (approved):

TomyLobo:
Rebased

worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
```diff
@@ -23,24 +23,28 @@
...
 
 public class WorldEditExpressionEnvironment implements ExpressionEnvironment {
 
-    private final Vector3 unit;
-    private final Vector3 zero2;
+    private final Transform transform;
     private Vector3 current = Vector3.ZERO;
     private final Extent extent;
 
+    @Deprecated
```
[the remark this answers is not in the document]
TomyLobo: Done, as well as all the other @Deprecated annotations I added

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -2267,6 +2270,7 @@ public List<Countable<BlockState>> getBlockDistribution(Region region, boolean s
      * @throws ExpressionException if there is a problem with the expression
      * @throws MaxChangedBlocksException if the maximum block change limit is exceeded
      */
+    @Deprecated
```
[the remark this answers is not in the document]
TomyLobo: Adjusted all @deprecated javadocs that link to an overload with a Transform and that are on methods that take zero/unit

worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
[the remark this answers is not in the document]
octylFractal: This links to the wrong method now?

worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
```diff
@@ -23,24 +23,31 @@
...
 public class WorldEditExpressionEnvironment implements ExpressionEnvironment {
 
-    private final Vector3 unit;
-    private final Vector3 zero2;
+    private final Transform transform;
     private Vector3 current = Vector3.ZERO;
     private final Extent extent;
 
+    /**
+     * @deprecated Use {@link EditSession#makeBiomeShape(Region, Transform, BiomeType, String, boolean, int)} and pass a {@link ScaleAndTranslateTransform}.
```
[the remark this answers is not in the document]
octylFractal: Not confident that resolves properly, please check the output.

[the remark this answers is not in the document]
me4502: I can confirm this resolves properly- should be good to commit as-is

PR #2278, opened 2023-03-09
Add the trim command

### LINKED ISSUE #1629: Allow trimming a cuboid selection based on a given predicate
### Is your feature request related to a problem? Please describe.

Currently, the manipulation of selections based on their contents is quite limited. It is hard to use the wand tool or get the manual coordinates of a built structure to select only that structure, without extraneous air around the selection.

### Describe the solution you'd like

Given a cuboid selection, a new command should be added (`//trim <mask>`) to trim that selection. The selection's bounds would change to the minimum and maximum positions that match that mask.

### Describe alternatives you've considered

One alternative that defeats the point is manually changing the selection.

haykam821:
Fixes #1629

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
@@ -41,6 +41,9 @@ import com.sk89q.worldedit.extension.platform.permission.ActorSelectorLimits;
 import com.sk89q.worldedit.extent.clipboard.Clipboard;
 import com.sk89q.worldedit.function.block.BlockDistributionCounter;
 import com.sk89q.worldedit.function.mask.Mask;
+import com.sk89q.worldedit.function.mask.MaskIntersection;
+import com.sk89q.worldedit.function.mask.Masks;
+import com.sk89q.worldedit.function.mask.RegionMask;
 import com.sk89q.worldedit.function.operation.Operations;
 import com.sk89q.worldedit.function.visitor.RegionVisitor;
 import com.sk89q.worldedit.internal.annotation.Chunk3d;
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
```diff
@@ -504,6 +505,133 @@ private BlockVector3[] getChangesForEachDir(int amount, boolean onlyHorizontal,
...
+
+                    if ((skipContains || originalRegion.contains(vec)) && mask.test(vec)) {
+                        maxZ = z;
+                        break outer;
+                    }
+                }
+            }
+        }
+
+        session.setRegionSelector(world, new CuboidRegionSelector(
```
me4502, line 626: This command should probably do something similar to https://github.com/EngineHub/WorldEdit/commit/ba94660e27a90f39d2c14e2593cacf5c4b00f8b1, to ensure that `//sel extend` is retained when using this command

worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
```diff
@@ -504,6 +505,133 @@ private BlockVector3[] getChangesForEachDir(int amount, boolean onlyHorizontal,
...
+        desc = "Minimize the selection to encompass matching blocks"
+    )
+    @Logging(REGION)
+    @CommandPermissions("worldedit.selection.trim")
+    public void trim(Actor actor, World world, LocalSession session,
+                     @Arg(desc = "Mask of blocks to keep within the selection", def = "#existing")
+                         Mask mask) throws WorldEditException {
+        // Avoid checking blocks outside the original region but within the cuboid region
+        Region originalRegion = session.getSelection(world);
+        boolean skipContains = originalRegion instanceof CuboidRegion;
```
me4502, line 519: Technically instead of doing this skipContains check in all of the below loops, you can conditionally wrap the mask with a RegionMask if it's not a cuboid region. This would bake it into the mask & simplify the code a fair bit

me4502 (approved):

octylFractal (approved):

PR #2280, opened 2023-03-14
Add more placement types and a //placement command to explicitly set placement type

TomyLobo:
Allows the following modes:
- `//toggleplace` - toggle between pos1 and player
- `//placement pos1` - set to pos1
- `//placement player` - set to player
- `//placement min` - set to selection minimum point
- `//placement max` - set to selection maximum point
- `//placement world` - set to world origin
- `//placement here` - set current player coordinates

All modes can be followed by a multiplier+offset in the same format as `//shift`

All modes have been tested with a regular player in the IDE.
I haven't tested it with a non-`Locatable` `Actor`, but it should work.

Should be ready to review, but the docs need to be changed to reflect the new command.
I have no idea where the docs are maintained, though.


worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -426,15 +426,28 @@ public void gmask(Actor actor, LocalSession session,
         desc = "Switch between your position and pos1 for placement"
     )
     public void togglePlace(Actor actor, LocalSession session) {
-        if (!(actor instanceof Locatable)) {
+        if (session.getPlacement() == Placement.POS1) {
+            placement(actor, session, Placement.PLAYER);
+        } else {
+            placement(actor, session, Placement.POS1);
+        }
```
TomyLobo, line 433: is calling another command handler like that ok?

TomyLobo:
rebased

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
TomyLobo: There's a helper method now.

TomyLobo:
rebased

TomyLobo:
Changed the syntax to `//placement <placementType> [<multiplier>] [<offset>]` to mimic `//move`, `//stack`, `//shift`, `//expand` and `//contract`.
Also rebased this thoroughly to clean up the history.

TomyLobo:
Fixed checkstyle error and tested all `PlacementType`s.
Windows builds are failing for some unrelated reason.

TomyLobo:
I need the underlying code changes from this to give `//brush deform` clipboard support.
Can this PR be reviewed and pulled soon, so that other PR doesn't list half of this PR's changes?

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
```diff
@@ -559,34 +561,40 @@ public boolean toggleSuperPickAxe() {
...
+    public Placement getPlacement() {
+        return this.placement;
+    }
 
-        return selector.getPrimaryPosition();
+    public void setPlacement(Placement placement) {
+        this.placement = placement;
     }
 
+    @Deprecated
```
wizjany, line 575: could we have a javadoc `@deprecated` tag here explaining what to use instead so people know where to migrate

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
```diff
@@ -559,34 +561,40 @@ public boolean toggleSuperPickAxe() {
...
     public void setPlaceAtPos1(boolean placeAtPos1) {
-        this.placeAtPos1 = placeAtPos1;
+        this.placement = new Placement(placeAtPos1 ? PlacementType.POS1 : PlacementType.PLAYER, BlockVector3.ZERO);
     }
 
+    @Deprecated
```
wizjany, line 580: as above

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
```diff
@@ -559,34 +561,40 @@ public boolean toggleSuperPickAxe() {
...
     }
 
     /**
-     * Toggle placement position.
+     * Toggle placement position between POS1 and PLAYER.
      *
-     * @return whether "place at position 1" is now enabled
+     * @return whether placement is now at position 1
      */
+    @Deprecated
```
wizjany, line 590: as above. also maybe mention that the default will be pos1 if it's outside the domain (i.e. between POS1 and PLAYER)

worldedit-core/src/main/java/com/sk89q/worldedit/session/PlacementType.java
```diff
@@ -0,0 +1,110 @@
...
+    private final String translationKeyWithOffset;
+
+    PlacementType(String translationKey, String translationKeyWithOffset) {
+        this.translationKey = translationKey;
+        this.translationKeyWithOffset = translationKeyWithOffset;
+    }
+
+    public abstract BlockVector3 getPlacementPosition(RegionSelector selector, Actor actor) throws IncompleteRegionException;
+
+    public boolean canUseActor(Actor actor) {
```
wizjany, line 99: feels like it would make more sense if this (and its derivatives) was named `canActorUse` or more verbosely `canBeUsedBy(Actor)` or something else that makes the subject and object more obvious.

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -69,7 +69,17 @@
...
     "worldedit.toggleplace.player": "Now placing at the block you stand in.",
+    "worldedit.toggleplace.player-offset": "Now placing at an offset of ({0}, {1}, {2}) from the block you stand in.",
+    "worldedit.toggleplace.min": "Now placing at the minimum of the current selection.",
+    "worldedit.toggleplace.min-offset": "Now placing at an offset of ({0}, {1}, {2}) from the minimum of the current selection.",
+    "worldedit.toggleplace.max": "Now placing at the maximum of the current selection.",
+    "worldedit.toggleplace.max-offset": "Now placing at an offset of ({0}, {1}, {2}) from the maximum of the current selection.",
+    "worldedit.toggleplace.world": "Now placing at world origin.",
+    "worldedit.toggleplace.world-offset": "Now placing at ({0}, {1}, {2}).",
+    "worldedit.toggleplace.here": "Now placing at the block you stand in *right now*.",
+    "worldedit.toggleplace.here-offset": "Now placing at an offset of ({0}, {1}, {2}) from the block you stand in *right now*.",
```
wizjany, line 82: this is actually a lot worse than my last nitpick on language. I genuinely find some of these confusing as to whether they are static locations or dynamic given the use of qualifiers like "current", "right now", or none.

i think part of this is that all of the absolute(/static) locations use different terminology when they realistically can all just tell the player the absolute coordinates being used as the placement position. that is, even though `world(-offset)` and `here-(offset)` might parse and calculate differently, they are all static coordinate locations (as far as i can tell), so that should really be all the player cares about. i wouldn't even bother telling them the offset. in all four cases just give them the absolute x/y/z position.
now, all the relative/dynamic locations can contrast that by giving the reference point (player, min/max, etc) and optional offset. i think that would make these a lot more clear.

wizjany (changes requested):
also re: docs, they are generated by the DocumentationPrinter though we generally only do it for version bumps, i.e. we will tag the docs repo for 7.2 and move to 7.3 when that's released.

worldedit-core/src/main/resources/lang/strings.json
TomyLobo: `worldedit.toggleplace.here[-offset]` aren't used in practice, since the special case HERE is replaced by WORLD before it ever hits the `printInfo` line in `placementImpl`.
These strings are here, just in case someone _else_ needs to use `printInfo` on the result of `PlacementType.HERE.getTranslationKey[Offset]` for some reason.

worldedit-core/src/main/java/com/sk89q/worldedit/session/PlacementType.java
TomyLobo: will be fixed in my next push

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
TomyLobo: Will be added in my next push

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
TomyLobo: Will be added in my next push

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
TomyLobo: Will be added in my next push, including an explanation for what happens if it was outside the previous domain.

TomyLobo:
All changes since @wizjany's review:
 https://github.com/EngineHub/WorldEdit/compare/0ad9a245ac5eb940ad54605386562b3623cd907a..toggleplace-arg-enhancement

worldedit-core/src/main/resources/lang/strings.json
TomyLobo: In truth, I didn't think it's worth making a new converter that accepts either `PlacementType` and the string "here".
That's a lot of extra code for just a tiny gain in code quality.
Making `setPlacement` reject it is a good idea though. I'll get on it.
I'll make `getTranslationKey[Offset]` throw as well while I'm at it.

TomyLobo: since those are gone now, I'm gonna resolve the conversation.

me4502 (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/session/PlacementType.java
```diff
@@ -0,0 +1,120 @@
...
+        @Override
+        public BlockVector3 getPlacementPosition(RegionSelector selector, Actor actor) throws IncompleteRegionException {
+            return BlockVector3.ZERO;
+        }
+    },
+
+    PLAYER("worldedit.toggleplace.player", "worldedit.toggleplace.player-offset") {
+        @Override
+        public BlockVector3 getPlacementPosition(RegionSelector selector, Actor actor) throws IncompleteRegionException
+        {
```
me4502, line 41: Is this `{` and the one on line 58 intentionally on a newline? It's not tripping checkstyle which confuses me, so is there a specific reason they're here?

TomyLobo: totally not intentional. might have been autoformat...
I'll fix it and look for other places where that happened in my PRs

PR #2288, opened 2023-03-26
(Forge 1.19.4) Fixed clients without WorldEdit not being able to join servers with WorldEdit installed

RedstoneDubstep:
During the 1.19.4 port of Forge, the absent version marker NetworkRegistry.ABSENT, which used to be of type String, has been changed to now be a ChannelData instance. This causes problems due to .equals() checks with the version marker and incoming versions, which worked in Forge for 1.19.3, now always failing due to two different objects being compared. In this specific case, it leads to clients without WorldEdit installed not being able to join Forge servers with WorldEdit installed due to the validateLenient method checking inappropriately whether the version is the ABSENT string, and thus the client connection getting rejected due to mismatching channel versions.
This PR fixes this issue by adding a `.version()` call to the affected check, ensuring that the incoming version is compared to the actual "Absent" version string instead of the ChannelData object.

### DIFF
--- a/worldedit-forge/src/main/java/com/sk89q/worldedit/forge/net/handler/PacketHandlerUtil.java
+++ b/worldedit-forge/src/main/java/com/sk89q/worldedit/forge/net/handler/PacketHandlerUtil.java
@@ -42,7 +42,7 @@ final class PacketHandlerUtil {
     private static Predicate<String> validateLenient(String protocolVersion) {
         return remoteVersion ->
             protocolVersion.equals(remoteVersion)
-                || NetworkRegistry.ABSENT.equals(remoteVersion)
+                || NetworkRegistry.ABSENT.version().equals(remoteVersion)
                 || NetworkRegistry.ACCEPTVANILLA.equals(remoteVersion);
     }
 }


worldedit-forge/src/main/java/com/sk89q/worldedit/forge/net/handler/PacketHandlerUtil.java
```diff
@@ -42,7 +42,7 @@ static NetworkRegistry.ChannelBuilder buildLenientHandler(String id, int protoco
     private static Predicate<String> validateLenient(String protocolVersion) {
         return remoteVersion ->
             protocolVersion.equals(remoteVersion)
-                || NetworkRegistry.ABSENT.equals(remoteVersion)
+                || NetworkRegistry.ABSENT.version().equals(remoteVersion)
                 || NetworkRegistry.ACCEPTVANILLA.equals(remoteVersion);
```
me4502, line 46: Does this one also need the same change?

RedstoneDubstep: No, since the `ACCEPTVANILLA` version constant is still a `String` object, as seen here in [NetworkRegistry.java](https://github.com/MinecraftForge/MinecraftForge/blob/1.19.x/src/main/java/net/minecraftforge/network/NetworkRegistry.java#L50).

me4502 (approved):

me4502:
Thanks for fixing this :)

JOO200:
shouldn't this be merged into the v7.2.x branch too?

Issue #2188, opened 2022-09-17
//sel extend forgets it was set after arbitrary time


e-motiv:
### WorldEdit Version

7.2.12

### Platform Version
Forge-1.19.2-43.1.25

### Confirmations

- [X] I am using the most recent Minecraft release.
- [X] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [X] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

1. I type 
    /tool farwand
    //sel extend
2. Perfect! Selections extend after right clicking
3. After a while, without dc'ing, right clicking doesn't extend anymore, but becomes a normal cuboid selection.

I tried to link this with other commands, but seems arbitrary for now.
This happens like 100 times per hour. Really frustrating, though I do appreciate your work, developers. Just saying.

me4502:
If you’ve disconnected from the server for >5 minutes this will happen. That’s so that worldedit isn’t storing data for potentially limitless players in memory over time

wizjany:
or if this is single player, immediately after leaving the world. either way this likely isn't happening randomly. you can always set your default selector if you want it to persist after the session being dropped. all documented behavior.

e-motiv:
Thanks for the comments both of you, but...

> @me4502: If you’ve disconnected from the server for >5 minutes this will happen. That’s so that worldedit isn’t storing data for potentially limitless players in memory over time

It happens without disconnecting or being AFK. Sometimes even after only 2 minutes actively editing the world.


> @wizjany: or if this is single player, immediately after leaving the world. 

It happens on single player indeed but also on multiplayer (though as an admin and only player online) WITHOUT leaving the world or disconnecting or being afk.


> @wizjany: either way this likely isn't happening randomly. 

Being a developer myself,  I understand this is probably not random, but so far it feels like it and I couldn't figure out when exactly this happens. One does a lot of things when editing the world, but I haven't notice any reproducibility yet.


> @wizjany: you can always set your default selector if you want it to persist after the session being dropped. all documented behavior.

[...]

PR #2289, opened 2023-03-27
Keep extending cuboid selector after //paste -s/n

TomyLobo:
When using `//paste -s` or `//paste -n`, the selector type would revert back to `//sel cuboid`.
This fixes uses the extending cuboid selector if it's the currently selected on and the regular cuboid selector otherwise.
Fixes an issue that might be the one reported in #2188.

There is one more relevant place in worldedit-cli that *looked* like it needed to be changed as well, so I also changed that.
I have no experience with worldedit-cli, so I hope that's ok.

Built locally and tested successfully.
Merges with both 7.2.x and master without conflicts.
worldedit-cli parts untested.

### DIFF
--- a/worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIExtraCommands.java
+++ b/worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIExtraCommands.java
@@ -23,6 +23,7 @@ import com.sk89q.worldedit.LocalSession;
 import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.extension.platform.Actor;
 import com.sk89q.worldedit.regions.selector.CuboidRegionSelector;
+import com.sk89q.worldedit.regions.selector.ExtendingCuboidRegionSelector;
 import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import com.sk89q.worldedit.util.task.Task;
 import com.sk89q.worldedit.world.World;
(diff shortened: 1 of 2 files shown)


TomyLobo:
Here's a diff and the full patched source file for worldedit-cli, just in case:
[CLIExtraCommands.zip](https://github.com/EngineHub/WorldEdit/files/11074089/CLIExtraCommands.zip)
```diff
diff --git a/worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIExtraCommands.java b/worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIExtraCommands.java
index c3f74f2dc..2ded87ac3 100644
--- a/worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIExtraCommands.java
+++ b/worldedit-cli/src/main/java/com/sk89q/worldedit/cli/CLIExtraCommands.java
@@ -23,6 +23,7 @@
 import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.extension.platform.Actor;
 import com.sk89q.worldedit.regions.selector.CuboidRegionSelector;
+import com.sk89q.worldedit.regions.selector.ExtendingCuboidRegionSelector;
 import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import com.sk89q.worldedit.util.task.Task;
 import com.sk89q.worldedit.world.World;
@@ -38,9 +39,13 @@ public class CLIExtraCommands {
         desc = "Select the entire world"
     )
     public void selectWorld(Actor actor, World world, LocalSession session) {
-        session.setRegionSelector(world, new CuboidRegionSelector(
-            world, world.getMinimumPoint(), world.getMaximumPoint()
-        ));
+        final CuboidRegionSelector selector;
+        if (session.getRegionSelector(world) instanceof ExtendingCuboidRegionSelector) {
+            selector = new ExtendingCuboidRegionSelector(world, world.getMinimumPoint(), world.getMaximumPoint());
+        } else {
+            selector = new CuboidRegionSelector(world, world.getMinimumPoint(), world.getMaximumPoint());
+        }
+        session.setRegionSelector(world, selector);
         actor.printInfo(TextComponent.of("Selected the entire world."));
     }
 
```

me4502 (approved):
IMO this looks good. I do wonder if we should do something to make this cleaner in the future, and not miss it when needed. Whether it be a static `create` function on the Cuboid selector or something else

PR #2292, opened 2023-04-02
Update McRegionChunkStore.java

ykgamma:
- Fixes an issue where the LinCompoundTag passed to ChunkStoreHelper::getChunk is not the expected root compound tag, but one that wraps it. See comment of LinRootEntry::toLinTag


me4502:
Thanks, I've tested and it's not working before this, and works with this change applied.

me4502 (approved):

PR #2294, opened 2023-04-05
Replace enum value access by namespaced keys

MartijnMuijsers:
Match with namespaced keys for `Material`, `EntityType` and `Biome`, instead of their Bukkit enum names.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
@@ -332,10 +332,7 @@ public class BukkitAdapter {
      */
     public static Material adapt(ItemType itemType) {
         checkNotNull(itemType);
-        if (!itemType.getId().startsWith("minecraft:")) {
-            throw new IllegalArgumentException("Bukkit only supports Minecraft items");
-        }
-        return Material.getMaterial(itemType.getId().substring(10).toUpperCase(Locale.ROOT));
+        return Material.matchMaterial(itemType.getId());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


wizjany:
can you elaborate on the motivation for this change? a quick look at that method seems to indicate this would actually be a regression on some hybrid platforms, and is otherwise functionally identical to what we're already doing.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
```diff
@@ -363,6 +357,7 @@ public static GameMode adapt(org.bukkit.GameMode gameMode) {
         return GameModes.get(gameMode.name().toLowerCase(Locale.ROOT));
     }
 
+    private static final Biome[] biomeValuesCache = Biome.values();
```
me4502, line 360: What's the purpose of caching these? Given the lookup occurs within a cached call I can't see it being _that_ major in terms of performance, especially given there aren't tens of thousands of biomes

me4502:
I definitely feel moving to keys here makes sense, the key APIs weren't available for these options back when the registry enumeration was added originally. Better to use actual MC IDs rather than hack around with the enums

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
```diff
@@ -246,10 +245,11 @@ private void initializeRegistries() {
...
         for (org.bukkit.entity.EntityType entityType : org.bukkit.entity.EntityType.values()) {
-            String mcid = entityType.getName();
-            if (mcid != null) {
-                String lowerCaseMcId = mcid.toLowerCase(Locale.ROOT);
-                EntityType.REGISTRY.register("minecraft:" + lowerCaseMcId, new EntityType("minecraft:" + lowerCaseMcId));
+            try {
+                String entityTypeKey = entityType.getKey().toString();
+                EntityType.REGISTRY.register(entityTypeKey, new EntityType(entityTypeKey));
+            } catch (IllegalArgumentException e) {
+                // EntityType.UNKNOWN has no key
```
me4502, line 252: For the two instances of this where the enum is iterated, could it instead just check that it's not the UNKNOWN value rather than let it throw an exception? Exceptions are vastly slower than doing a comparison a few times, and look a lot messier. Blindly catching an exception here also might cause us to miss actual errors in the future too

me4502:
Also - as this is basically a "bug fix" and isn't adding any API or new features, this should likely target the `version/7.2.x` branch

me4502:
Closing this as it's superseded by https://github.com/EngineHub/WorldEdit/pull/2330

PR #2305, opened 2023-04-16
Use an annotation to register the methods in the Functions class

TomyLobo:
There's a lot less code now and it's also easier to add new methods.


me4502 (approved):
This makes a lot of sense to me, only concern would be if we for some reason wanted to register an arg as a different type to the expression runtime than what the Java type was, but idk if that's actually likely

PR #2317, opened 2023-05-17
Add Console/CommandBlock support to Fabric/Forge/Sponge

me4502:
Adds console & command block support to Forge, Fabric, & Sponge. Is semi awkward for console due to the weird world names (Fabric uses `world_minecraft:overworld`???)

Targeted 7.2.x as this is platform parity rather than a "new" feature


wizjany:
does it make sense to hoist the config to core now that it's (mostly) cross-platform? also potentially make an interface in core that notates those platform impls as commandblock/console senders. would be useful for automatically determining if things like //schem load should block instead of being async. also would be useful to avoid issues like enginehub/worldguard#1724

me4502:
I'll hoist the config option, and make a common one for command blocks. IMO we already have a good class to designate "console" in these PRs, as we don't actually add these as console, they're just "non-player sender that isn't an already handled type". So `AbstractNonPlayerActor` already covers it enough IMO

PR #2319, opened 2023-05-21
Add a memoizing mask utility

me4502:
Decided to add a simple mask memoizer utility. Realised it'd likely be useful for the Trim command PR, as well as a few other places throughout the codebase.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/MaskMemoizer.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/MaskMemoizer.java
@@ -1,3 +1,22 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/MaskMemoizer.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public boolean test(BlockVector3 vector) {
+        return this.cache.computeIfAbsent(vector, mask::test);
+    }
+
+    public void clear() {
+        this.cache.clear();
+    }
+
+    @Nullable
```
octylFractal, line 54: Should not be marked nullable?

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/Masks.java
```diff
@@ -55,6 +55,28 @@ public static Mask2D alwaysTrue2D() {
         return ALWAYS_TRUE;
     }
 
+    /**
+     * Memoize the given mask.
```
octylFractal, line 59: Might be good to include the comment about single-use here and below.

octylFractal (approved):

PR #2325, opened 2023-05-25
Improve ArbitraryShape caching

TomyLobo:
Every point is now only evaluated once.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/ArbitraryShape.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/ArbitraryShape.java
@@ -121,13 +121,11 @@ public abstract class ArbitraryShape {
         int y = position.getBlockY();
         int z = position.getBlockZ();
 
-        final BaseBlock defaultMaterial = pattern.applyBlock(position);
-
         if (!hollow) {
-            return getMaterial(x, y, z, defaultMaterial);
+            return getMaterial(x, y, z, pattern.applyBlock(position));
         }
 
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/ArbitraryShape.java
```diff
@@ -37,21 +39,27 @@ public abstract class ArbitraryShape {
...
+     * <li>Optional.of(material) = inside</li>
+     * </ul>
+     *
+     * <p>You're <i>technically</i> not *supposed* to set variables of type <code>Optional</code> to null, but if we did
+     * that, we'd need to use <code>Optional&lt;Optional&lt;BaseBlock&gt;&gt;</code> and pre-fill the cache with
+     * <code>Optional.empty()</code>, so we just use null here and rely on Java to pre-fill the array with
+     * <code>null</code>s for us.
      */
-    private final byte[] cache;
+    private final Optional<BaseBlock>[] cache;
```
me4502, line 62: Has any performance testing of this been done? This might substantially increase the memory usage & the CPU time (due to memory allocations & indirection) of the caching system here, potentially outweighing the underlying calculations

TomyLobo: This is now an `Object[]`, resolving.

me4502:
Out of curiosity how does this impact command durations? Even just a before & after command time elapsed with the `debug`/`profile` config option enabled (changes name per platform)

PR #2330, opened 2023-06-04
Use registry lookups for Bukkit adapters

me4502:
When a lot of this code was written, Bukkit did not support Registry lookups. As these are now supported and various bugs have been fixed, we can greatly clean this up.

I also standardised the isValidMob check to be like other platforms

Targets master as this breaks support <1.17

Supersedes https://github.com/EngineHub/WorldEdit/pull/2294 - resolves all GitHub comments and makes better use of Bukkit API (no iterations of enums etc)


MartijnMuijsers:
Nice work! Thank you.

PR #2332, opened 2023-06-11
feat: implement recently-added tree types to tree generator

dordsor21:
I believe this is what the trees should be like

image

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/TreeGenerator.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/TreeGenerator.java
@@ -108,6 +108,9 @@ public final class TreeGenerator {
                 return editSession.getWorld().generateTree(this, editSession, pos.subtract(0, 1, 0));
             }
         },
+        MANGROVE("Mangrove tree", "mangrove"),
+        TALL_MANGROVE("Tall mangrove tree", "tall_mangrove"),
+        CHERRY("Cherry blossom", "cherry"),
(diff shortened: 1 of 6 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/util/TreeGenerator.java
```diff
@@ -108,6 +108,9 @@ public boolean generate(EditSession editSession, BlockVector3 pos) throws MaxCha
                 return editSession.getWorld().generateTree(this, editSession, pos.subtract(0, 1, 0));
             }
         },
+        MANGROVE("Mangrove tree", "mangrove"),
```
me4502, line 111: This doesn't map these anywhere, so these won't do anything. They'll need to be mapped to tree features in the platforms

dordsor21: Ah yeah mb, I was copying over from FAWE changes and we lazily just look up the enums via Bukkit's methods

me4502:
Looks like tests are failing

dordsor21:
Yeah, that'll be because of the bukkit module using 1.17 API I suppose, so the TreeType enum doesn't include cherry or mangrove trees. Not sure if/when you guys would want to switch to a later paper API?

dordsor21:
👍 Won't be able to do any test until Thurs or Fri

me4502:
Thanks, just bumped it so should pass now

Issue #2370, opened 2023-07-24
No longer loading sign texts from old schematics


Brokkonaut:
### WorldEdit Version

7.2.16-SNAPSHOT (build 6489)

### Platform Version

git-Paper-"b533905" (MC: 1.20.1)

### Confirmations

- [X] I am using the most recent Minecraft release.
- [X] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [X] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

WorldEdit is no longer loading texts in signs in old schematics, this was working in 1.19.4 but no longer in 1.20

### Expected Behavior

The text on those signs should be loaded

### Reproduction Steps

1. Copy the schematic from https://webinterface.cubeside.de/temp/shop_mi_016.schematic to your schematics folder
2. Load, Paste
3. No text on the sign in 1.20

emilyy-dev:
Duplicate of #2369

octylFractal:
I don't think so, I think this is the other way around. We should aim to load older schematics even if the format changes.

wizjany:
old schematics are manually upgraded via the compat handlers at the moment. those only move things to 1.13-compatible formats. moving 1.13->1.20 and beyond would be another step. i tried doing it with DFU when i originally implemented that but it was pretty buggy at the time (well, it still is tbh). paper's dataconverters might alleviate that though, but idk if there's any interest in fixing that atm. this is kinda just how it goes when mojang changes data formats.

Brokkonaut:
I tried this to load schematics: https://github.com/Brokkonaut/WorldEdit/commit/c3adf1b3c1a83791ab0b7fe943d9e08b5050d2f1 - this converts the signs correctly and at least for some old schematics we have I could not find any issues.

I used Constants.DATA_VERSION_MC_1_13_2, because the compatibility handlers already update it to 1.13 and with -1 it just caused errors.

Would it be ok, if I PR it?

wizjany:
[...]

PR #2373, opened 2023-07-27
Use DFU for updating MCEdit schematics

Brokkonaut:
See https://github.com/EngineHub/WorldEdit/issues/2370


me4502:
How heavily has this been tested? I'd inclined to merge this if it does appear to work better, although would ideally prefer to try testing whatever issues led to them being skipped in the first place (if @wizjany remembers what they were)

wizjany:
I think some of the issues were due to running the legacy fixers (i.e. from DV=-1) as opposed to using our inbuilt compathandlers in tandem with DFU from DV=1.13, as this PR does. That said DFU has always been finicky and it may be worth doing some error handling and throwing out tags that cause exceptions just in case?

me4502:
We could also add DFU as a side effect that can be turned off in //perf as a follow up too, if there are some instances where it's problematic to have it enabled

me4502:
Closed in https://github.com/EngineHub/WorldEdit/commit/392393552874bc8d9ceecfe984a33e41ba55bbe4, thanks - I ended up rebasing this against master rather than putting it into 7.2.x due to potential risk of the change.

PR #2379, opened 2023-08-03
Add Folia basic support;

HarvelsX:
Adds adapters for schedulers. Required for Bukkit and Folia schedulers.
Created schedulers can be used in WorldGuard and in other plugins;

+ [ ] Tested in detail


octylFractal (commented):
I'm unsure this covers all the changes we'd need to be properly folia-supportive, but one thing for sure is I don't think we want to expose `SchedulerAdapter` + these `Entity` methods as full API. I think those belong specifically in the Bukkit section. I'm not sure we'd even like to have them be public there, but that's a discussion I'd need to have with the rest of the dev team.

Yomamaeatstoes:
When testing whether this project would build in Ubuntu I did encounter a build failure. I pasted it in mclogs
https://mclo.gs/PXagKd8. When I tried building it in Terminal I did encounter this issue as well. I suppose they could be apart of the same issue https://mclo.gs/FGdD5KX

octylFractal:
Yes, the new API must be accepted.

PR #2391, opened 2023-09-03
Apply a few micro-optimisations

me4502:
This applies a few optimisations to a few places that showed up in tracing mode of a profiler. It does appear to have made things a little faster, but it's hard to tell in tracing mode.

Either way, this basically just adds a bunch of fast exits when something isn't required, which is probably a good idea anyway

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
@@ -117,12 +117,15 @@ public class BlockTransformExtent extends AbstractDelegateExtent {
         checkNotNull(block);
         checkNotNull(transform);
 
+        if (transform.isIdentity()) {
+            return block;
+        }
+
         B result = block;
         List<? extends Property<?>> properties = block.getBlockType().getProperties();
 
         for (Property<?> property : properties) {
-            if (property instanceof DirectionalProperty) {
-                DirectionalProperty dirProp = (DirectionalProperty) property;
+            if (property instanceof DirectionalProperty dirProp) {
                 Direction value = (Direction) block.getState(property);
                 if (value != null) {
                     Vector3 newValue = getNewStateValue(dirProp.getValues(), transform, value.toVector());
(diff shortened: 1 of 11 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
```diff
@@ -117,36 +117,32 @@ public static <B extends BlockStateHolder<B>> B transform(B block, Transform tra
...
-                        case "z":
-                            value = Direction.NORTH;
-                            break;
-                        default:
-                            break;
+                        case "x" -> value = Direction.EAST;
+                        case "y" -> value = Direction.UP;
+                        case "z" -> value = Direction.NORTH;
+                        default -> {
+                        }
```
octylFractal, line 145: Should be switch expression, e. g.
```java
var value = switch(...) {...}
```
`null` can be returned from the default branch.

worldedit-core/src/main/java/com/sk89q/worldedit/util/LocatedBlock.java
```diff
@@ -22,46 +22,37 @@
...
 /**
  * Represents a block located at some position.
  */
-public final class LocatedBlock {
-
-    private final BlockVector3 location;
-    private final BaseBlock block;
+public record LocatedBlock(BlockVector3 location, BaseBlock block) {
 
     public LocatedBlock(BlockVector3 location, BaseBlock block) {
```
octylFractal, line 32: Can drop argument list here. 
```suggestion
     public LocatedBlock {
```

worldedit-core/src/main/java/com/sk89q/worldedit/util/LocatedBlock.java
```diff
@@ -22,46 +22,37 @@
...
         this.location = checkNotNull(location);
         this.block = checkNotNull(block);
```
octylFractal, line 34:
```suggestion
         checkNotNull(location);
         checkNotNull(block);
```

octylFractal (changes requested):

octylFractal (approved):

PR #2410, opened 2023-10-14
Fix doctools with platform refactor

me4502:
This updates doctools to work with the platform refactor

### DIFF
--- a/worldedit-core/doctools/src/main/kotlin/com/sk89q/worldedit/internal/util/DocumentationPlatform.kt
+++ b/worldedit-core/doctools/src/main/kotlin/com/sk89q/worldedit/internal/util/DocumentationPlatform.kt
@@ -64,11 +64,11 @@ class DocumentationPlatform : AbstractPlatform() {
 
     override fun getCapabilities(): MutableMap<Capability, Preference> {
         val capabilities: MutableMap<Capability, Preference> = EnumMap(Capability::class.java)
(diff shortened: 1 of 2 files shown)


worldedit-core/doctools/src/main/kotlin/com/sk89q/worldedit/internal/util/DocumentationPlatform.kt
```diff
@@ -55,11 +58,20 @@ class DocumentationPlatform : AbstractPlatform() {
...
-    override fun registerCommands(commandManager: CommandManager?) = error("Documentation does not provide this")
+    override fun registerCommands(commandManager: CommandManager?) {}
 
-    override fun setGameHooksEnabled(enabled: Boolean) = error("Documentation does not provide this")
+    override fun setGameHooksEnabled(enabled: Boolean) {}
 
-    override fun getCapabilities() = error("Documentation does not provide this")
+    override fun getCapabilities(): MutableMap<Capability, Preference> {
+        val capabilities: MutableMap<Capability, Preference> = EnumMap(Capability::class.java)
+        capabilities[Capability.CONFIGURATION] = Preference.NORMAL
```
octylFractal, line 67: I don't think it really matters but I think all of these should be `PREFER_OTHERS`

octylFractal (approved):

Issue #2415, opened 2023-10-15
Apply max-radius to the '//line' command.


fadfhkdfga:
### The Problem

Since max-radius is not applied to the '//line' command, a problem occurs that can cause anyone to crash the server.

### A Solution

Please ensure that the 'max-radius' setting is also applied to the '//line' command.
//line anyblock 10 << here

PR #2417, opened 2023-10-16
Limit //line thickness by max-radius

me4502:
Fixes https://github.com/EngineHub/WorldEdit/issues/2415.

I doubt this actually fixes the issue mentioned in the issue, but logically this is a radius and therefore does make sense to have limited by this setting.


wizjany:
probably worth adding to //curve as well

PR #2447, opened 2023-12-11
Deprecate the concept of bundled block data

me4502:
This PR deprecates the concept of bundled block data, and removes it from all non-CLI platforms.

To fully remove it from CLI, we'll have to likely add this data into the current CLI data format, which is out of scope for this PR and a bit of a larger task (new version; also needs to include items etc)

This also, like with the item registry changes, allows us to skip loading of the bundled data on startup. If a plugin is using it for some reason, it'll still load, but by default it won't be loaded into server memory.

### DIFF
--- a/worldedit-bukkit/adapters/adapter-1.20.2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_20_R2/PaperweightBlockMaterial.java
+++ b/worldedit-bukkit/adapters/adapter-1.20.2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_20_R2/PaperweightBlockMaterial.java
@@ -1,0 +1,145 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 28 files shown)


worldedit-bukkit/adapters/adapter-1.20.2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_20_R2/PaperweightBlockMaterial.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Override
+    public boolean isAir() {
+        return block.isAir();
+    }
+
+    @Override
+    public boolean isFullCube() {
+        VoxelShape vs = block.getCollisionShape(EmptyBlockGetter.INSTANCE, BlockPos.ZERO);
+        return !vs.isEmpty() && vs.bounds().equals(FULL_CUBE);
```
PierreSchwang, line 51: Might be worth using `Block.isShapeFullBlock(block.getShape(EmptyBlockGetter.INSTANCE, BlockPos.ZERO))` instead of comparing the shape manually? That'd utilize moonrise patches and optimizations (including [additional] caching). That's also what minecraft uses for it's own checks like "can snow falling onto this block spawn a snow layer"

me4502: thanks, i've applied this change.

worldedit-bukkit/adapters/adapter-1.21.4/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_21_4/PaperweightBlockMaterial.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Override
+    public boolean isTranslucent() {
+        return !block.canOcclude();
+    }
+
+    @Override
+    public boolean hasContainer() {
+        return block.getBlock() instanceof EntityBlock entityBlock
+                && entityBlock.newBlockEntity(BlockPos.ZERO, block) instanceof Clearable;
```
wizjany, line 137: Is this valid? iirc (most?) BEs validate on creation.

wizjany (approved):

PR #2453, opened 2023-12-24
Improve error handling for unsupported schematic files

me4502:
This PR does two things:

* Unwraps an internal `Schematic` tag, to improve cross-version compatibility when reading schematic files (or ones that violate spec)
* Prints the unsupported version error to the player, rather than as an IOException. This also means it now supports translation.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SchematicLoadException.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SchematicLoadException.java
@@ -24,7 +24,7 @@ import com.sk89q.worldedit.util.formatting.text.Component;
 /**
  * Raised when a known exception occurs during schematic load.
  */
-public class SchematicLoadException extends RuntimeException {
+public final class SchematicLoadException extends RuntimeException {
 
     private final Component message;
 
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SchematicLoadException.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SchematicLoadException.java
@@ -1,0 +1,43 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SchematicLoadException.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package com.sk89q.worldedit.extent.clipboard.io;
+
+import com.sk89q.worldedit.util.formatting.text.Component;
+
+/**
+ * Raised when a known exception occurs during schematic load.
+ */
+public class SchematicLoadException extends RuntimeException {
```
octylFractal, line 27: Can we mark this as non-API, i.e. with a Javadoc comment? Or maybe move it to some internal package?

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SpongeSchematicReader.java
```diff
@@ -154,6 +155,12 @@ private CompoundTag getBaseTag() throws IOException {
         // Check
         Map<String, Tag> schematic = schematicTag.getValue();
 
+        // Handle newer versions in a cleaner way
```
octylFractal, line 158: I would like to be more explicit about the leniency here.
```suggestion
        // Be lenient about the specific nesting level of the Schematic tag
        // Also allows checking the version from newer versions of the specification
```

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SpongeSchematicReader.java
```diff
@@ -124,7 +125,7 @@ public Clipboard read() throws IOException {
             BlockArrayClipboard clip = readVersion1(schematicTag);
             return readVersion2(clip, schematicTag);
         }
-        throw new IOException("This schematic version is currently not supported");
+        throw new SchematicLoadException(TranslatableComponent.of("worldedit.schematic.load.unsupported-version"));
```
octylFractal, line 128: While we're here, can we add the version to the message for clarity?

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SchematicLoadException.java
me4502: Is there a reason to do that? Given adding schematic formats is API, I can see throwing this exception as pretty useful from an API perspective

octylFractal: Hmm, I guess it's fine. Maybe we can just mark it `final` then? I don't know if I want to open it up for extending that we might need to consider later.

octylFractal (approved):

PR #2460, opened 2024-01-17
fix hang on shutdown for fabric due to hanging user threads

### LINKED ISSUE #2459: Fabric with worldedit hangs when stopping during shutdown
### WorldEdit Version

7.3.0-beta-03

### Platform Version

Fabric Loader 0.15.3

### Confirmations

### Bug Description

The server hangs during shutdown in some cases. 
The was tracked down to the thread "WorldEdit Task Executor - 0" left running, whichin turn is due to the WorldEdit.executorService not being stopped. This particular task executor service is not run using daemon threads and thus require an explicit shutdown to terminate. If not stopped it will prevent the jvm from initiating the shutdown.

### Expected Behavior

Sending /stop to the server should terminate the process in a reasonable time.

### Reproduction Steps

1. Make a selection
2. //copy
3. //schem save somefilename
4. /stop

Observe that the fabric server does not termiante. 
If checking the remaining server status with jstack the thread "WorldEdit Task Executor - 0" is still running and not a daemon thread.

### Anything Else?

I am preparing a PR for this, hopefully later today.

petersv5:
The JVM initiates shutdown when all non-daemon java threads have terminated. When saving a schematic two threads are created that are never stopped, thus preventing the server for finishing the shutdown.

One is the "WorldEdit Task Executor - 0", shich is easily fixed.

The second is the TimerThread created by FutureProgressListener. In this patch the timer is made to use a deamon thread instead. If critical scheduled events on this timer persost after a server shutdown is otherwise complete maybe this needs to be changed to an executor service? Otherwise scheduled events may get lost by the server shutting down.

This fixes #2459

### DIFF
--- a/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
+++ b/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
@@ -90,6 +90,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -286,6 +287,7 @@ public class FabricWorldEdit implements ModInitializer {
         WorldEdit worldEdit = WorldEdit.getInstance();
         worldEdit.getSessionManager().unload();
         WorldEdit.getInstance().getEventBus().post(new PlatformUnreadyEvent(platform));
+        WorldEdit.getInstance().getExecutorService().shutdown();
     }
 
     private boolean skipEvents() {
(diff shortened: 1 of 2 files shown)


worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
```diff
@@ -286,6 +287,7 @@ private void onStopServer(MinecraftServer minecraftServer) {
         WorldEdit worldEdit = WorldEdit.getInstance();
         worldEdit.getSessionManager().unload();
         WorldEdit.getInstance().getEventBus().post(new PlatformUnreadyEvent(platform));
+        WorldEdit.getInstance().getExecutorService().shutdown();
```
me4502, line 290: If this is something that must be shut down on Fabric, I feel that we should probably also shut this down on other platforms.

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
```diff
@@ -90,6 +90,7 @@
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
```
me4502, line 93: This import also seems to be unnecessary, it should be removed

me4502:
i'll close this for now in favour of https://github.com/EngineHub/WorldEdit/pull/2570 which has been merged. i'm uncertain whether making the Timer a daemon thread could cause some unknown issues, but shutting down the executor is fairly safe.

PR #2462, opened 2024-01-28
WorldEdit-Folia

Yomamaeatstoes:
Hello, I have been using for the past month. I have tested it and it seems others have as well.

If you have any requested changes let me know.


README.md
```diff
@@ -84,3 +84,8 @@ YourKit supports open source projects with innovative and intelligent tools for
 YourKit is the creator of [YourKit Java Profiler](https://www.yourkit.com/java/profiler/),
 [YourKit .NET Profiler](https://www.yourkit.com/.net/profiler/),
 and [YourKit YouMonitor](https://www.yourkit.com/youmonitor/).
+
+# Folia Inquisitors 
+[<img src="https://github.com/Folia-Inquisitors.png" width=80 alt="Folia-Inquisitors">](https://github.com/orgs/Folia-Inquisitors/repositories)
+[<img src="https://github.com/TechnicallyCoded.png" width=80 alt="TechnicallyCoded">](https://github.com/TechnicallyCoded)
+[<img src="https://github.com/Yomamaeatstoes.png" width=80 alt="Yomamaeatstoes">](https://github.com/Yomamaeatstoes)
```
octylFractal, line 91: Why was this added? I don't see why we should be specifically calling these groups/people out. Unless you plan to assist in maintaining Folia support, in which case it would be good to mention that here, e.g. "Thanks to <xyz> for maintaining Folia support in WorldEdit". But if you do so, we will assign Folia issues to you.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java
```diff
@@ -462,25 +462,9 @@ public com.sk89q.worldedit.world.block.BlockState getBlock(BlockVector3 position
...
-        if (WorldEditPlugin.getInstance().getLocalConfiguration().unsupportedVersionEditing) {
-            Block bukkitBlock = getWorld().getBlockAt(position.getBlockX(), position.getBlockY(), position.getBlockZ());
-            bukkitBlock.setBlockData(BukkitAdapter.adapt(block), sideEffects.doesApplyAny());
-            return true;
-        } else {
-            throw new RuntimeException(new UnsupportedVersionEditException());
-        }
+        Block bukkitBlock = getWorld().getBlockAt(position.getBlockX(), position.getBlockY(), position.getBlockZ());
+        bukkitBlock.setBlockData(BukkitAdapter.adapt(block), sideEffects.doesApplyAny());
+        return true;
```
octylFractal, line 467: This completely breaks setting NBT and side-effect support in all Bukkit-based servers. The solution here _must_ not remove the calls to `worldNativeAccess`.

octylFractal (changes requested):
This is almost certainly not enough fixes, as entities will likely need adjusting to Folia as well.

See #2379 for some more changes that may be needed, but likely more vetting will be needed before we declare full Folia support.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java
Yomamaeatstoes: Alright my team will be looking into this. Thanks for the feedback.

Yomamaeatstoes: I already have a proposed idea.

README.md
Yomamaeatstoes: Thanks for the feedback. I have talked to one of the members in the organization who so far plans on doing a pull request. Once that is done it should be reflected in the readme.

PR #2469, opened 2024-02-17
Slightly optimise and improve debugability of state creation code

me4502:
This PR does some minor optimisations to memory allocation in the state creation code to reduce GC churn, moves the watchdog tick to reduce ability for poorly coded mods to trigger the watchdog, and adds a debug line to output what block is taking a ridiculously long amount of time.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockState.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockState.java
@@ -133,8 +133,9 @@ public class BlockState implements BlockStateHolder<BlockState> {
                 watchdog.tick();
             }
         }
-        if (System.currentTimeMillis() - startTime > 60000) {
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockState.java
```diff
@@ -79,22 +79,30 @@ public void setInternalId(BlockState blockState, int internalId) {
...
 
         if (!properties.isEmpty()) {
-            List<List<Object>> separatedValues = Lists.newArrayList();
+            // Create a list of lists of values, with a copy of the underlying lists
+            List<List<Object>> separatedValues = Lists.newArrayListWithCapacity(properties.size());
             for (Property<?> prop : properties) {
-                List<Object> vals = Lists.newArrayList();
-                vals.addAll(prop.getValues());
-                separatedValues.add(vals);
+                separatedValues.add(Lists.newArrayList(prop.getValues()));
```
octylFractal, line 96: Can we make this immutable instead?

octylFractal (approved):

PR #2478, opened 2024-03-02
Folia support

Euphillya:
Yes, another PR for Folia...

I'm using [Energy](https://github.com/Euphillya/Energie) to make WorldEdit compatible for Folia while keeping Spigot and older versions of Minecraft available. 

## Will Energie be kept up to date without breaking Spigot? 

If Folia makes a change, yes I'll update it because I need it for my Skyllia plugin (Skyblock for Folia/Spigot).

## Are any of the current features no longer working?

I tested Paper and Folia on my side and nothing seemed to be broken, but checks are requested just in case.

## Is it mergeable?

At the moment, I'd say there isn't enough feedback to merge it and mark it: SUPPORT FOLIA! 

## Will Energie still be available?

I'm using Jitpack.io, because I can't afford hosting for only 1 repo.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -32,4 +32,8 @@ jobs:
       with:
         name: logs for ${{ matrix.os }}
         path: '**/*.log'
-
+    - name: Upload WorldEdit jar
+      uses: actions/upload-artifact@v2
+      with:
+        name: WorldEdit-Folia-Snapshot
+        path: worldedit-bukkit/build/libs/*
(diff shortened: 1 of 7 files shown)


me4502:
This is almost definitely not enough to make it actually work properly with folia. The entire edit layer is ignoring the concept of regions, and entirely unmodified by this PR.

Also, adding a random unknown dependency is not something we'd likely want to do. That feels like something more worthwhile to contribute to a project such as PaperLib that we already use

Euphillya:
Currently, basic commands like pos1/pos2 /set replace seem to work. But WE has so much functionality that they can be crashed.

Next, for Energie, I fully understand your point of view. Especially since there's no guarantee that I won't abandon it in the future, and we'll have to find a way to maintain it!

I'll see about doing something else, like submitting to PaperLib so there won't be any need for an additional dependency.

PR #2498, opened 2024-04-01
Return of the -Werror

octylFractal:
Renames, removals, and more to fix all warnings in `javac` and `javadoc` invocations.

return_of_the_werror


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java
```diff
@@ -272,8 +275,23 @@ public boolean generateTree(TreeGenerator.TreeType type, EditSession editSession
         if (bukkitType == TreeType.CHORUS_PLANT) {
             pt = pt.add(0, 1, 0); // bukkit skips the feature gen which does this offset normally, so we have to add it back
         }
-        return type != null && world.generateTree(BukkitAdapter.adapt(world, pt), bukkitType,
-                new EditSessionBlockChangeDelegate(editSession));
+        return type != null && world.generateTree(
+            BukkitAdapter.adapt(world, pt),
+            ThreadLocalRandom.current(),
+            bukkitType,
+            block -> {
```
me4502, line 282: What was the reason for inlining this rather than using the block change delegate?

octylFractal: The new method doesn't take one of those, AFAIK?

me4502: Fair, just assumed we'd just update the delegate to match the new function signature rather than inline, but I'm happy either way

PR #2502, opened 2024-04-03
Add a #clipboard mask, to mask to blocks that match the clipboard

me4502:
This is the second part of `//ungen`. It adds a `#clipboard` mask that checks if the block matches the clipboard. In the future once "Masks 2.0" is complete, arguments to add offset can be added here too. For now it requires clipboard space to match world space (which is fine if you do `//copy`, `//regen -c`, or use a schematic from the same world, etc).

Functionality of `//ungen` can be replicated by simply doing:

* `//regen -c`
* `//replace #clipboard air`


worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/ClipboardMaskParser.java
```diff
@@ -0,0 +1,55 @@
...
+import com.sk89q.worldedit.function.mask.Mask;
+import com.sk89q.worldedit.function.mask.MatchMask;
+import com.sk89q.worldedit.internal.registry.SimpleInputParser;
+import com.sk89q.worldedit.util.formatting.text.TranslatableComponent;
+
+import java.util.List;
+
+public class ClipboardMaskParser extends SimpleInputParser<Mask> {
+
+    private final List<String> aliases = ImmutableList.of("#clipboard");
```
octylFractal, line 36: `static`?

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/MatchMask.java
```diff
@@ -0,0 +1,65 @@
...
+
+package com.sk89q.worldedit.function.mask;
+
+import com.sk89q.worldedit.extent.Extent;
+import com.sk89q.worldedit.math.BlockVector3;
+
+/**
+ * A mask that returns true if the two given extents have the same block at the given position, with offset support.
+ */
+public class MatchMask extends AbstractMask {
```
octylFractal, line 28: `final`?

octylFractal (approved):

PR #2507, opened 2024-04-07
Update for Java 21

octylFractal:
- Fix some new warnings
- Removed sponge from build, as it simply isn't maintained well currently. May be removed entirely in the future
- Remove 1.18 support from Bukkit as it breaks under 21 for some reason
- Fix a bug in `ForwardSeekableInputStream`

This PR is not to be merged until 1.20.5.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -46,16 +46,28 @@ val mixinVersion: String = properties.getProperty("mixin.version")
 dependencies {
     implementation(gradleApi())
     implementation("gradle.plugin.org.cadixdev.gradle:licenser:0.6.1")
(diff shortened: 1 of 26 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/storage/McRegionReader.java
```diff
@@ -134,6 +134,7 @@ public synchronized InputStream getChunkInputStream(BlockVector2 position) throw
 
         stream.seek((long) sectorNumber * SECTOR_BYTES);
         int length = dataStream.readInt();
+        System.err.println("Length: " + length);
```
me4502, line 137: Debug code?

octylFractal:
Closing in favor of #2512

PR #2510, opened 2024-04-11
Update to 1.20.5

DrexHD:
This draft PR is intended to track the progress of updating this mod _(fabric module)_ to 1.20.5
(Based on `ot/feature/jdk-21-hype`)


octylFractal:
Unfortunately, I've already done most of this locally and in certain ways that avoid breaking our API like this PR does. So I'm going to close it in favor of that, which should be up soon.

DrexHD:
Sure. Is `FabricAdapter` considered API or what breaking changes are you referring to?

wizjany:
https://worldedit.enginehub.org/en/latest/api/internal-apis/

PR #2514, opened 2024-04-18
Deprecate bundled item data

me4502:
This PR deprecates the concept of bundled item data, as well as the concept of "item material". This was added but never actually used. It's not very mod-compatible, and also requires storing increasingly large numbers of JSON files within the repo and jar files.

Due to the fact that WorldEdit never uses these at all, and that the bundled item registry is loaded when first accessed, this means we can remove the call to actually load it from WorldEdit. If something does happen to be using this deprecated API, it'll trigger it to be loaded upon first access.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitItemRegistry.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitItemRegistry.java
@@ -24,7 +24,7 @@ import com.sk89q.worldedit.util.formatting.text.Component;
 import com.sk89q.worldedit.world.item.ItemType;
 import com.sk89q.worldedit.world.registry.BundledItemRegistry;
 
-@SuppressWarnings("deprecation")
+@SuppressWarnings("removal")
 class BukkitItemRegistry extends BundledItemRegistry {
     @Override
     public Component getRichName(ItemType itemType) {
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitItemRegistry.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitItemRegistry.java
@@ -24,6 +24,7 @@ import com.sk89q.worldedit.util.formatting.text.Component;
 import com.sk89q.worldedit.world.item.ItemType;
 import com.sk89q.worldedit.world.registry.BundledItemRegistry;
 
+@SuppressWarnings("deprecation")
 class BukkitItemRegistry extends BundledItemRegistry {
     @Override
     public Component getRichName(ItemType itemType) {
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/item/ItemType.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/item/ItemType.java
@@ -111,7 +111,7 @@ public class ItemType implements Keyed {
      * @return The material
      * @deprecated Deprecated without alternative
      */
-    @Deprecated
+    @Deprecated(forRemoval = true)
     @SuppressWarnings("removal")
     public ItemMaterial getMaterial() {
         return itemMaterial.getValue();
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/item/ItemType.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/item/ItemType.java
@@ -51,7 +51,7 @@ public class ItemType implements Keyed {
         WorldEdit.getInstance().getPlatformManager().queryCapability(Capability.GAME_HOOKS)
             .getRegistries().getItemRegistry().getRichName(this)
     );
-    @SuppressWarnings("this-escape")
+    @SuppressWarnings("removal")
     private final LazyReference<ItemMaterial> itemMaterial = LazyReference.from(() ->
         WorldEdit.getInstance().getPlatformManager().queryCapability(Capability.GAME_HOOKS)
             .getRegistries().getItemRegistry().getMaterial(this)
@@ -109,7 +109,10 @@ public class ItemType implements Keyed {
      * Get the material for this ItemType.
      *
      * @return The material
+     * @deprecated Deprecated without alternative
      */
+    @Deprecated
+    @SuppressWarnings("removal")
     public ItemMaterial getMaterial() {
         return itemMaterial.getValue();
     }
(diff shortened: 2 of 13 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledItemRegistry.java
```diff
@@ -73,6 +76,8 @@ public String getName(ItemType itemType) {
 
     @Nullable
     @Override
+    @Deprecated
+    @SuppressWarnings("removal")
```
octylFractal, line 80:
```suggestion
    @Deprecated(forRemoval = true)
```

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledItemData.java
```diff
@@ -104,8 +108,11 @@ public ItemEntry findById(String id) {
      *
      * @param id the string ID
      * @return the material's properties, or null
+     * @deprecated Deprecated without alternative
      */
     @Nullable
+    @Deprecated
+    @SuppressWarnings("removal")
```
octylFractal, line 115:
```suggestion
    @Deprecated(forRemoval = true)
```

worldedit-core/src/main/java/com/sk89q/worldedit/world/item/ItemType.java
```diff
@@ -107,7 +108,10 @@ public BlockType getBlockType() {
      * Get the material for this ItemType.
      *
      * @return The material
+     * @deprecated Deprecated without alternative
      */
+    @Deprecated
+    @SuppressWarnings("removal")
```
octylFractal, line 114:
```suggestion
    @Deprecated(forRemoval = true)
```

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitItemRegistry.java
```diff
@@ -24,6 +24,7 @@
 import com.sk89q.worldedit.world.item.ItemType;
 import com.sk89q.worldedit.world.registry.BundledItemRegistry;
 
+@SuppressWarnings("deprecation")
```
octylFractal, line 27:
```suggestion
@Deprecated
```

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledItemData.java
```diff
@@ -46,7 +46,10 @@
  * <p>The data is read from a JSON file that is bundled with WorldEdit. If
  * reading fails (which occurs when this class is first instantiated), then
  * the methods will return {@code null}s for all items.</p>
+ *
+ * @deprecated Deprecated without replacement.
  */
+@Deprecated
```
octylFractal, line 52:
```suggestion
@Deprecated(forRemoval = true)
```

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledItemRegistry.java
```diff
@@ -30,7 +30,10 @@
 /**
  * A item registry that uses {@link BundledItemRegistry} to serve information
  * about items.
+ *
+ * @deprecated Use the platform Item Registries
  */
+@Deprecated
```
octylFractal, line 36:
```suggestion
@Deprecated(forRemoval = true)
```

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/ItemRegistry.java
```diff
@@ -63,7 +63,10 @@ default String getName(ItemType itemType) {
      *
      * @param itemType the item
      * @return the material, or null if the material information is not known
+     * @deprecated ItemMaterial is deprecated with no alternative.
      */
+    @Deprecated(forRemoval = true)
     @Nullable
+    @SuppressWarnings("removal")
```
octylFractal, line 70: Is this really needed?

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/PassthroughItemMaterial.java
```diff
@@ -23,6 +23,8 @@
 
 import static com.sk89q.worldedit.util.GuavaUtil.firstNonNull;
 
+@Deprecated(forRemoval = true)
+@SuppressWarnings("removal")
```
octylFractal, line 27: Is this really needed?

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/SimpleItemMaterial.java
```diff
@@ -19,5 +19,7 @@
 
 package com.sk89q.worldedit.world.registry;
 
+@Deprecated(forRemoval = true)
+@SuppressWarnings("removal")
```
octylFractal, line 23: Is this really needed?

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitItemRegistry.java
me4502: This class isn't deprecated

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/ItemRegistry.java
me4502: Yes, any reference to something deprecated for removal is now erroring the build due to producing a warning, even if the current block is deprecated too

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BundledRegistries.java
```diff
@@ -65,6 +65,7 @@ static URL loadRegistry(String name) throws IOException {
     }
 
     private final BundledBlockRegistry blockRegistry = new BundledBlockRegistry();
+    @SuppressWarnings("removal")
```
octylFractal, line 68: Shouldn't we deprecate this field / the related getters in this class?

me4502: we can deprecate the field, but we cannot deprecate the getter currently because that propagates it to all classes that extend this, and would lead to non-deprecated methods getting marked as deprecated.

we currently can't remove this from the inheritance tree, so it's not really an option to deprecate this currently

octylFractal (approved):

PR #2542, opened 2024-06-01
Schematic caching system for suggestions

me4502:
This is a continuation of https://github.com/EngineHub/WorldEdit/pull/2212, as the PR is abandoned by its original author


worldedit-core/src/main/java/com/sk89q/worldedit/WorldEdit.java
```diff
@@ -396,8 +407,10 @@ private File getSafeFileWithExtension(File dir, String filename, String extensio
         return new File(dir, filename);
     }
 
+    private static final java.util.regex.Pattern SAFE_FILENAME_REGEX = java.util.regex.Pattern.compile("^[A-Za-z0-9_\\- \\./\\\\'\\$@~!%\\^\\*\\(\\)\\[\\]\\+\\{\\},\\?]+\\.[A-Za-z0-9]+$");
+
     private boolean checkFilename(String filename) {
-        return filename.matches("^[A-Za-z0-9_\\- \\./\\\\'\\$@~!%\\^\\*\\(\\)\\[\\]\\+\\{\\},\\?]+\\.[A-Za-z0-9]+$");
+        return SAFE_FILENAME_REGEX.matcher(filename).matches();
```
octylFractal, line 413: This set of characters doesn't make sense, it's not safe for windows (contains `?`, `*`, doesn't ban COM and others), and POSIX will generally take anything. It also excludes extended Unicode characters for some reason, despite those being perfectly fine.

Can we just drop checking this aside from "you must have an extension", i.e. `^.+\..+$`?

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,234 @@
...
+     * @throws IOException If creating the watcher failed, e.g. due to root not existing.
+     */
+    public static Optional<RecursiveDirectoryWatcher> create(Path root) throws IOException {
+        try {
+            WatchService watchService = root.getFileSystem().newWatchService();
+            return Optional.of(new RecursiveDirectoryWatcher(root, watchService));
+        } catch (UnsupportedOperationException ignored) {
+            return Optional.empty();
+        }
+    }
```
octylFractal, line 108: Let's not bother with the `Optional` path here. We're not currently dealing with non-default `Path` providers, and I don't expect to any time soon. Quote from the Javadocs of `newWatchService`:
> This exception is not thrown by FileSystems created by the default provider.

So, let's remove the `Optional` and just throw.

```suggestion
    public static RecursiveDirectoryWatcher create(Path root) throws IOException {
        WatchService watchService;
        try {
            watchService = root.getFileSystem().newWatchService();
        } catch (UnsupportedOperationException e) {
            throw new IllegalArgumentException("Root must support file watching", new RecursiveDirectoryWatcher(root, watchService));
        }
        return new RecursiveDirectoryWatcher(root, watchService);
    }
```

worldedit-core/src/main/java/com/sk89q/worldedit/internal/schematic/SchematicsManager.java
```diff
@@ -0,0 +1,129 @@
...
+            var fileWatcherBackend = FileWatcherSchematicsBackend.create(schematicsDir);
+            if (fileWatcherBackend.isPresent()) {
+                backend = fileWatcherBackend.get();
+            } else {
+                createFallbackBackend();
+            }
+        } catch (IOException e) {
+            createFallbackBackend();
+        }
+    }
```
octylFractal, line 73: Assuming that my `Optional`-removing suggestion is accepted, let's clean this up too:
```suggestion
    private void setupBackend() {
        try {
            backend = FileWatcherSchematicsBackend.create(schematicsDir);
        } catch (IOException e) {
            LOGGER.warn("Failed to initialize file-monitoring based schematics backend. Falling back to polling.", e);
            backend = PollingSchematicsBackend.create(schematicsDir);
        }
    }
```

worldedit-core/src/main/java/com/sk89q/worldedit/internal/schematic/SchematicsManager.java
```diff
@@ -0,0 +1,129 @@
...
+     */
+    public void init() {
+        try {
+            schematicsDir = worldEdit.getWorkingDirectoryPath(worldEdit.getConfiguration().saveDir);
+            Files.createDirectories(schematicsDir);
+            schematicsDir = schematicsDir.toRealPath();
+            setupBackend();
+        } catch (IOException e) {
+            LOGGER.warn("Failed to create schematics directory", e);
+            backend = new DummySchematicsBackend(); //fallback to dummy backend
```
octylFractal, line 87: This comment isn't needed. It's obvious.
```suggestion
            backend = new DummySchematicsBackend();
```

worldedit-core/src/main/java/com/sk89q/worldedit/internal/schematic/SchematicsManager.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Initialize this SchematicsManager.
+     * This sets everything up, and initially scans the schematics folder.
+     */
+    public void init() {
+        try {
+            schematicsDir = worldEdit.getWorkingDirectoryPath(worldEdit.getConfiguration().saveDir);
+            Files.createDirectories(schematicsDir);
+            schematicsDir = schematicsDir.toRealPath();
```
octylFractal, line 83: Let's not set this twice.
```suggestion
            Path schematicsDirByConfig = worldEdit.getWorkingDirectoryPath(worldEdit.getConfiguration().saveDir);
            Files.createDirectories(schematicsDirByConfig);
            schematicsDir = schematicsDirByConfig.toRealPath();
```

worldedit-core/src/main/java/com/sk89q/worldedit/internal/schematic/backends/DummySchematicsBackend.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.sk89q.worldedit.internal.schematic.backends;
+
+import java.nio.file.Path;
+import java.util.Set;
+
+/**
+ * A backend that never lists any files.
+ */
+public class DummySchematicsBackend implements SchematicsBackend {
```
octylFractal, line 28: I feel like this backend should throw instead. We're using it as the driving force behind `//schem list` now, so if it fails, people shouldn't just get an empty list out, they should get an error. Or possibly, we shouldn't have this at all and should crash out if we have an error setting up.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/schematic/backends/FileWatcherSchematicsBackend.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * A backend that efficiently scans for file changes using {@link RecursiveDirectoryWatcher}.
+ */
+public class FileWatcherSchematicsBackend implements SchematicsBackend {
+    private static final Logger LOGGER = LogManagerCompat.getLogger();
+
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+    private final Set<Path> schematics = new HashSet<>();
```
octylFractal, line 40: Why are we not using a `Sets.newConcurrentHashSet()`? It would work much better since it doesn't need to lock the whole set.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,234 @@
...
+     * Make this RecursiveDirectoryWatcher instance start monitoring the root folder it was created on.
+     * When this is called, RecursiveDirectoryWatcher will send initial notifications for the entire
+     * file structure in the configured root.
+     * @param eventConsumer The lambda that's fired for every file event.
+     */
+    public void start(Consumer<DirEntryChangeEvent> eventConsumer) {
+        Path schematicRoot = WorldEdit.getInstance().getSchematicsManager().getRoot();
+
+        this.eventConsumer = eventConsumer;
+        watchThread = new Thread(() -> {
```
octylFractal, line 139: Just in case, we should check for an existing `watchThread` and kill it or throw. Killing requires us to have a means of communicating to it.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,234 @@
...
+                        if (watchRootMap.isEmpty()) {
+                            break; // nothing left to watch
+                        }
+                    }
+                }
+            } catch (ClosedWatchServiceException | FilenameException ignored) {
+            }
+            LOGGER.debug("RecursiveDirectoryWatcher::EventConsumer exited");
+        });
+        watchThread.setName("RecursiveDirectoryWatcher");
```
octylFractal, line 208: At the very least we should identify ourselves fully. Perhaps we should also use a global counter.
```suggestion
        watchThread.setName("WorldEdit-RecursiveDirectoryWatcher");
```

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/WorldEdit.java
me4502: So I do agree fundamentally here, although I feel like it'd make sense to do that separately to this change

worldedit-core/src/main/java/com/sk89q/worldedit/internal/schematic/backends/DummySchematicsBackend.java
me4502: So the current setup appears to be that it prints a warn in the console when it swaps out to this. IMO having it not completely explode is probably ideal, but still throwing an error makes sense. Would you be fine with me just swapping that log point to an error-level one?

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
me4502: So my memory of the Java thread model is a bit lacking; I feel like I've now set it up like that (it already appeared to be semi-setup for a Thread-interruption model), but not 100% sure

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    @Override
+    public void close() {
+        try {
+            watchService.close();
+        } catch (IOException e) {
+            LOGGER.error(e);
+        }
+        if (watchThread != null) {
+            try {
```
octylFractal, line 235: Can we `interrupt()` here as well?

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
```diff
@@ -0,0 +1,245 @@
...
+     * @param eventConsumer The lambda that's fired for every file event.
+     */
+    public void start(Consumer<DirEntryChangeEvent> eventConsumer) {
+        Path schematicRoot = WorldEdit.getInstance().getSchematicsManager().getRoot();
+
+        this.eventConsumer = eventConsumer;
+
+        if (watchThread != null) {
+            // We have an existing thread, interrupt it first to trigger a clean shutdown
+            watchThread.interrupt();
```
octylFractal, line 145: We should `join` here first.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
octylFractal: Yes, this should mostly be fine.

octylFractal (approved):

PR #2544, opened 2024-06-05
Forward `getBlock` and `getFullBlock` to extents chain

EpicPlayerA10:
Forward `getBlock` and `getFullBlock` to the extents chain. I don't see the reason why it must directly access world extent. For example the `getBiome` already passes its call to extents chain:
https://github.com/EngineHub/WorldEdit/blob/00148305cea5fe99187b5b30912349a5c4c8b8b7/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java#L690-L693

After this change, people can override `getBlock` and `getFullBlock` in `AbstractDelegateExtent` to make some modifications in getting blocks. For example in plugins that adds custom blocks and they need to modify block nbt a bit.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -699,12 +699,12 @@ public class EditSession implements Extent, AutoCloseable {
 
     @Override
     public BlockState getBlock(BlockVector3 position) {
-        return world.getBlock(position);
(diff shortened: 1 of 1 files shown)


octylFractal:
This would need a much further rework and extensive testing of pretty much every feature as it at least affects how masks apply when multiple blocks are set in the same location for one `EditSession`, and likely many other features that rely on this to bypass to the world. Feel free to test and fix these issues, but as it stands this PR cannot be merged.

PR #2566, opened 2024-07-03
Lazily compute BlockState#getAsString

me4502:
This PR adds lazy computation of the BlockState#getAsString (and therefore toString) method, allowing this to be computed once rather than calculated every time it's called.

This also partially benefits BaseBlock#toString, as it delegates to this method for the BlockState portion of the string, with NBT data appended to the end.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BaseBlock.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BaseBlock.java
@@ -193,7 +193,7 @@ public class BaseBlock implements BlockStateHolder<BaseBlock>, TileEntityBlock {
     }
 
     /**
-     * Gets a string representation of this BaseBlock, in the format expected WorldEdit's block parsers.
+     * Gets a string representation of this BaseBlock, in the format expected by WorldEdit's block parsers.
      *
      * <p>
      * If NBT data is present, it will be included in the string. If you only want the underlying block state, call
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BaseBlock.java
```diff
@@ -192,8 +192,18 @@ public int hashCode() {
         return ret;
     }
 
+    /**
+     * Gets a string representation of this BaseBlock.
```
octylFractal, line 196: Can we be more specific about the format? This method isn't useful without strict details on the output.

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockStateHolder.java
```diff
@@ -133,6 +133,11 @@ default BaseBlock applyBlock(BlockVector3 position) {
         return toBaseBlock();
     }
 
+    /**
+     * Gets a String representation of this BlockStateHolder.
```
octylFractal, line 137: Similar here.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BaseBlock.java
me4502: Can I be specific to the extent that we state it complies with MC's formatting / the parser's formatting? i'd rather not lock us into formatting by API contract

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BaseBlock.java
```diff
@@ -192,8 +192,18 @@ public int hashCode() {
         return ret;
     }
 
+    /**
+     * Gets a string representation of this BaseBlock, in the format expected by the block parser.
```
octylFractal, line 196: I assume this is our block parser? Can we link to the class for clarity?

octylFractal (approved):

PR #2571, opened 2024-07-07
Refactor PlatformManager to use locks

me4502:
This is mostly a test to see if we can replace the synchronized methods in this class with two locks, one for the platforms and one for the capability preferences.

Theoretically this should improve performance due to the overhead of synchronization, and the fact that platforms & capabilities are independent systems that can be locked separately.


octylFractal:
IMO we don't need to do this any more. `synchronized` should be around the same performance as a lock now, with improvements such as https://openjdk.org/jeps/450#Locking coming soon. There will also no longer be any pinning with `synchronized`, see https://openjdk.org/jeps/491.

In general I don't think we contest on these methods a lot and I would prefer to have hard data on the improvement before trying to push this further.

PR #2580, opened 2024-07-17
Add clipboard support to //brush deform

TomyLobo:
After #2276 adds clipboard support to //deform, the same feature can be added to //brush deform.

Due to some additional complexities and unresolved discussions, I'm splitting this off into its own PR.

This comment thread on the //deform PR about //brush deform was left unresolved:
comment thread from the other PR


TomyLobo:
Rebased

TomyLobo:
and rebased onto deform-clipboard again, after making the requested changes on that PR

PR #2581, opened 2024-07-18
Completely eliminate usage of other repos

octylFractal:
Probably, I'm not firewalling it to make sure


me4502 (approved):

octylFractal:
Attention: this PR might break building on your machine if you have a particular resolution result cached. Please try deleting files matching `~/gradle/caches/modules-2/*/resource-at-url.bin` first, and report here if that doesn't work.

PR #2583, opened 2024-07-21
Cleanup registry init

me4502:
Pretty minor PR that cleans up a few of the Bukkit Material API usages to be a bit cleaner / use more modern APIs, as well as reduces calls to `toString` of NamespacedKey/ResourceLocation. Also disables legacy ID conversions for the internal parser used for adapters in Bukkit, which should theoretically speed it up a little bit given it doesn't need to do legacy parsing

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
@@ -82,6 +82,7 @@ public class BukkitAdapter {
 
     static {
         TO_BLOCK_CONTEXT.setRestricted(false);
+        TO_BLOCK_CONTEXT.setTryLegacy(false);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
```diff
@@ -334,7 +335,8 @@ public static Entity adapt(org.bukkit.entity.Entity entity) {
      */
     public static Material adapt(ItemType itemType) {
         checkNotNull(itemType);
-        return Material.matchMaterial(itemType.id());
+        var key = checkNotNull(NamespacedKey.fromString(itemType.id()), "Item type key is invalid");
```
octylFractal, line 338:
```suggestion
        NamespacedKey key = checkNotNull(NamespacedKey.fromString(itemType.id()), "Item type key is invalid");
```

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
```diff
@@ -345,7 +347,8 @@ public static Material adapt(ItemType itemType) {
      */
     public static Material adapt(BlockType blockType) {
         checkNotNull(blockType);
-        return Material.matchMaterial(blockType.id());
+        var key = checkNotNull(NamespacedKey.fromString(blockType.id()), "Item type key is invalid");
```
octylFractal, line 350:
```suggestion
        NamespacedKey key = checkNotNull(NamespacedKey.fromString(blockType.id()), "Block type key is invalid");
```

octylFractal (approved):

Issue #2575, opened 2024-07-08
//rotate and //flip don't modify block data correctly


LoliColleen:
### WorldEdit Version

Bukkit-Official(7.3.4+6823-6263244)

### Platform Version

Paper version 1.21-37-master@dd49fba (2024-06-24T02:08:49Z) (Implementing API version 1.21-R0.1-SNAPSHOT)

### Confirmations

- [X] I am using the most recent Minecraft release.
- [X] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [X] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

//rotate and //flip command can make the block data of the clipboard broken. (When the neighbour block update is DISBALED)
For example, you can copy some rails, then do //rotate 90 and //paste. You can find that the rails are not rotated correctly.
This bug affects "rail", "cobblestone_wall" and so on. (Excluding fences, doors, trapdoors and ... fine i haven't tested all the blocks)
The walls blocks will even be invisible because their block data is broken.

### Expected Behavior

The blocks like rails and cobblestone walls should be correctly rotated or flipped

### Reproduction Steps

1. Do "//perf neighbors off"
2. Select an area with rails and cobblestone walls 
3. Do "//copy"
4. Do "//rotate 90" or "//flip"
5. Do "//paste"

LoliColleen:
Wall blocks will lose Directional data after rotating. But they still have their Wall data

brickmonster:
I've been meaning to report this exact issue.

Same repro steps as OP.
* 7.3.1 on Paper 1.20.4 succeeds
* 7.3.2 on Paper 1.20.4 fails
* 7.3.4 on Paper 1.20.4 fails
* 7.3.1 on Paper 1.20.6 succeeds
* 7.3.2 on Paper 1.20.6 fails
* 7.3.4 on Paper 1.20.6 fails
* 7.3.1 on Paper 1.21 is unsupported
* 7.3.2 on Paper 1.21 is unsupported
* 7.3.4 on Paper 1.21 fails
* FAWE on Paper 1.21 succeeds (but they have updates permanently off, so only a temporary solution)

FakkaJohan:
[...]

PR #2601, opened 2024-08-13
Fix incorrect paste orientation with `//perf -h update off`

brickmonster:
Fixes https://github.com/EngineHub/WorldEdit/issues/2575

Regression was introduced in https://github.com/EngineHub/WorldEdit/commit/6818272ace6f0ccf23b9ef448f0b04356d9ac784

The commit does not mention why the change was made. To me it looks like a cleanup that didn't intend for any changes that got squashed into the addition of support for crafters.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
@@ -283,7 +283,7 @@ public class BlockTransformExtent extends AbstractDelegateExtent {
                                 Property<Object> propAsObj = (Property<Object>) prop;
                                 result = result.with(propAsObj, "none");
                             }
-                            directionalProperties.put(closestProp, result.getState(prop));
+                            directionalProperties.put(closestProp, state);
                         }
                     }
                 }


worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
```diff
@@ -126,7 +126,7 @@ public static <B extends BlockStateHolder<B>> B transform(B block, Transform tra
 
         for (Property<?> property : properties) {
             if (property instanceof DirectionalProperty dirProp) {
-                Direction value = (Direction) result.getState(property);
+                Direction value = (Direction) block.getState(property);
```
me4502, line 129: are you sure these changes are the issue? these were done specifically because it caused only the last property to be modified to actually apply to the resulting block

me4502: also outside of the case where this is a block that modifies multiple properties with a transform, this code is functionally identical. `B result = block;` is line 124, so for the vast majority of blocks this should be identical

brickmonster: The referenced commit is the only one in the 7.3.2 release that seemed suspicious to me. Completely reverting it fixes the issue. This pull request is a smaller revert that still fixes it; not sure why it does, I didn't get that far.

brickmonster: `directionalProperties` was getting the wrong value inserted. The regression code was getting the result state rather than the original state.

me4502: makes sense, thanks

me4502 (approved):

brickmonster:
Thanks :)

PR #2613, opened 2024-08-27
Commands hpos1/2: Calculate .toBlockPoint() only once.

mk-pmb:
I'm working on a patch that will use the block position here, so I'd like to simplify the upstream to make my rebases easier.

Should I go further and also unify the common pattern in hpos1/hpos2?
I could make an `hposCommon` method that hpos1 and hpos2 then call and pass the diverging objects as argument.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
@@ -223,13 +223,14 @@ public class SelectionCommands {
         Location pos = player.getBlockTrace(300);
 
         if (pos != null) {
-            if (!session.getRegionSelector(player.getWorld()).selectPrimary(pos.toVector().toBlockPoint(), ActorSelectorLimits.forActor(player))) {
+            BlockVector3 block = pos.toVector().toBlockPoint();
+            if (!session.getRegionSelector(player.getWorld()).selectPrimary(block, ActorSelectorLimits.forActor(player))) {
                 player.printError(TranslatableComponent.of("worldedit.hpos.already-set"));
                 return;
             }
 
             session.getRegionSelector(player.getWorld())
-                    .explainPrimarySelection(player, session, pos.toVector().toBlockPoint());
+                    .explainPrimarySelection(player, session, block);
         } else {
             player.printError(TranslatableComponent.of("worldedit.hpos.no-block"));
         }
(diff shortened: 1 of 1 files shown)


octylFractal:
> I'm working on a patch that will use the block position here, so I'd like to simplify the upstream to make my rebases easier.

I'm not sure what you mean by this. Rebase will cleanly apply unless we touch this code, which we haven't in years and aren't planning to.

mk-pmb:
Sorry for not explaining in detail. In my usual work, deduplication of function calls is a default, so I had considered it an obvious desirable thing.
My code will calculate based on those coordinates. If I want to leave the upstream code as is, I'd have to invoke  .toBlockPoint() a 3rd time. I'd then save the result to a variable in order to not need even further invocations, but still it would confuse readers about why I call it three times.
Any other patch that wants to reuse it, would also have to introduce their own. Each could have a comment referencing this thread for justification, but IMO the right thing to do is to deduplicate upstream.
If there is a good reason to call .toBlockPoint() multiple times here, I'd be curious, and it should probably be documented.

octylFractal:
It's more that this seems like unnecessary minor code churn. If you're going to do a cleanup like this, (1) you should discuss it with us in more detail so we can come up with some more precise standards, and (2) it should apply everywhere. Doing little changes all over without them being part of a larger effort requires a lot of effort on our part (CI runs, review, credit, changelog). These type of random weird changes with no context can also lead to less attention to detail on each one, and end up in an [xz type of backdoor](https://www.openwall.com/lists/oss-security/2024/03/29/4).

If you're planning to make a PR, this should be part of it. We do not see frequent enough PRs to warrant doing something because some other theoretical PR might use it. This is likely going to be the only PR that touches this code for a while. If you're maintaining your own patches, there really is no reason to upstream this random change. Duplication in other forks is unlikely since again, this code is rarely touched.

> My code will calculate based on those coordinates. If I want to leave the upstream code as is, I'd have to invoke .toBlockPoint() a 3rd time

Your code _should_ change this. Not doing so seems rather silly to me.

>  If there is a good reason to call .toBlockPoint() multiple times here, I'd be curious, and it should probably be documented.

There isn't, but it's not a big deal at all: the objects are immutable so there's no thread concerns, and the conversion is essentially truncating 3 doubles. This code is in a command so the performance impact is non-existent. There's no reason to just add this change. If you're touching the code in another PR, it can be part of that. Updating code as it's touched is how we prefer handling code debt unless there is a whole-codebase change. Otherwise, who knows how many PRs like this you and others would send to us? There is tons of tech debt that could be cleaned up, but doing it one at a time like this is just too much overhead for the amount of code.

mk-pmb:
Thanks for the explanation! Do we have a list of our tech debt? Then I could add it there, and maybe even clean up a much larger chunk of that list.

PR #2618, opened 2024-09-05
Re-enable buffering

octylFractal:
This is required for masks to work properly.


me4502 (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -618,10 +621,12 @@ public void setBatchingChunks(boolean batchingChunks) {
             }
             return;
         }
+        assert batchingExtent != null : "same nullness as chunkBatchingExtent";
```
me4502, line 624: so i don't necessarily have issue with assert statements, but i don't believe we really use these anywhere else?

octylFractal: I don't know if I've needed these elsewhere, it was just to shut IntelliJ up about null checks.

octylFractal: That is -- this is purely informational for the reader and the IDE, not meant to be run.

PR #2621, opened 2024-09-07
Chunk section editing

octylFractal:
Currently Fabric-only.

Fixes #chunk

TODO:
- [ ] Clean up code
- [ ] NeoForge
- [ ] Paper (augh reflection)


octylFractal:
@avaruus1 if you're interested in keeping the Sponge part up-to-date with everything, there are changes here that need to be made. As we don't want to maintain it, I just removed the broken code. This won't land for at least a couple months so there's plenty of time to adjust.

avaruus1:
> @avaruus1 if you're interested in keeping the Sponge part up-to-date with everything, there are changes here that need to be made. As we don't want to maintain it, I just removed the broken code. This won't land for at least a couple months so there's plenty of time to adjust.

I think the side effect calculation is fine to remove as is. The block change flags are set already [here](https://github.com/EngineHub/WorldEdit/blob/485bb4116709be42af81219ac28e20f54dbe12e6/worldedit-sponge/src/main/java/com/sk89q/worldedit/sponge/SpongeWorld.java#L193). Is there some other scenarios where this is possibly used?

octylFractal:
No, it's only useful if the platform calls out to it. If you don't want to port the chunk section stuff then there should be nothing to do.

octylFractal:
I have tested this as much as I could, but it's still possible some edge case isn't being handled. Otherwise, this should be good to merge.

worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/collection/ChunkSectionMask.java
```diff
@@ -0,0 +1,107 @@
...
+import it.unimi.dsi.fastutil.shorts.AbstractShortCollection;
+import it.unimi.dsi.fastutil.shorts.ShortCollection;
+import it.unimi.dsi.fastutil.shorts.ShortIterator;
+
+import java.util.BitSet;
+
+/**
+ * A mask for a chunk section.
+ */
+public class ChunkSectionMask {
```
me4502, line 31: I wonder if just due to context, we should use a name other than `Mask` here.

octylFractal:
Not really, but I haven't had energy to work on it.

PR #2676, opened 2024-12-14
Add a Full Cube mask to mask to blocks that fill an entire cube

me4502:
This adds a mask (`#fullcube`) that masks to anything that fits the "isFullCube" material property. This is distinct from the solid check, which includes blocks that don't quite fill the cube. Oddly, the solid check includes blocks that fill more than `0.7291666666666666` of the bounds of a block, which means stairs, etc are all included. fullblock would only include blocks that are exactly the bounds of the block.

This PR somewhat ideally needs https://github.com/EngineHub/WorldEdit/pull/2447, which implements this check everywhere.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/FullCubeMaskParser.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/FullCubeMaskParser.java
@@ -19,7 +19,6 @@
 
 package com.sk89q.worldedit.extension.factory.parser.mask;
 
-import com.google.common.collect.ImmutableList;
 import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.extension.input.InputParseException;
 import com.sk89q.worldedit.extension.input.ParserContext;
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/FullCubeMaskParser.java
```diff
@@ -0,0 +1,49 @@
...
+import com.sk89q.worldedit.extension.input.ParserContext;
+import com.sk89q.worldedit.function.mask.FullCubeMask;
+import com.sk89q.worldedit.function.mask.Mask;
+import com.sk89q.worldedit.internal.registry.SimpleInputParser;
+
+import java.util.List;
+
+public class FullCubeMaskParser extends SimpleInputParser<Mask> {
+
+    private final List<String> aliases = ImmutableList.of("#fullcube");
```
octylFractal, line 34: `static`?

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/FullCubeMask.java
```diff
@@ -0,0 +1,39 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.function.mask;
+
+import com.sk89q.worldedit.extent.Extent;
+import com.sk89q.worldedit.math.BlockVector3;
+import com.sk89q.worldedit.world.block.BlockState;
+
+public class FullCubeMask extends AbstractExtentMask {
```
octylFractal, line 26: `final`?

wizjany (approved):

PR #2683, opened 2024-12-25
Update worldedit-neoforge.mixins.json to use java 21

Thunderrock424242:
since we are in minecraft 1.21 we should be using java 21 to use mixins

### DIFF
--- a/worldedit-neoforge/src/main/resources/worldedit-neoforge.mixins.json
+++ b/worldedit-neoforge/src/main/resources/worldedit-neoforge.mixins.json
@@ -1,7 +1,7 @@
 {
     "required": true,
     "package": "com.sk89q.worldedit.neoforge.mixin",
-    "compatibilityLevel": "JAVA_17",
+    "compatibilityLevel": "JAVA_21",
     "mixins": [
         "AccessorServerPlayerGameMode",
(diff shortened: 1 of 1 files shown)


me4502:
Is this correct? when MC started requiring Java 21, the general consensus was that it had to remain set to JAVA_17 due to some issues that i don't entirely recall. have these issues been resolved now? if so, this probably also would need to change in the Fabric distribution of WorldEdit too?

wizjany:
[The docs](https://github.com/SpongePowered/Mixin/wiki/PSA-Forward-Compatibility-Features-in-Mixin#2-mixin-config-property-compatibilitylevel) indicate
> `compatibilityLevel` should always be set to the highest level required by your mixins.
So as long as there's no breaking ("non-upgrade-able") changes between 17 and 21 (there aren't, obviously, since our mixins run fine on 21), this shouldn't actually matter. It doesn't really have anything to do with MC version outside of the fact that Mojang is now bumping Java version more liberally as opposed to the past where we were stuck on Java8 forever.
If we wanted to pedantically have the mixin level to match the MC version/target build version/whatever, I'd rather do it via gradle resource replacement and have it apply equally across the project. At the very least, there's no reason to update the NF one here without the Fabric one even without automation.

PR #2701, opened 2025-01-22
Add pale oak tree variants

SirYwell:
1.21.4 brought a new tree type that isn't supported yet. I tested the changes on fabric.

### DIFF
--- a/worldedit-bukkit/build.gradle.kts
+++ b/worldedit-bukkit/build.gradle.kts
@@ -38,11 +38,6 @@ val adapters = configurations.create("adapters") {
 dependencies {
     "api"(project(":worldedit-core"))
     "api"(project(":worldedit-libs:bukkit"))
-    // Technically this is api, but everyone should already have some form of the bukkit API
-    // Avoid pulling in another one, especially one so outdated.
-    "localImplementation"(libs.spigot) {
-        exclude("junit", "junit")
-    }
 
     "localImplementation"(platform(libs.log4j.bom)) {
         because("Spigot provides Log4J (sort of, not in API, implicitly part of server)")
@@ -56,6 +51,9 @@ dependencies {
         because("Resolving Spigot annotations")
     }
     "compileOnly"(libs.paperApi) {
+        exclude("junit", "junit")
+    }
+    "testImplementation"(libs.paperApi) {
         exclude("junit", "junit")
     }
     "implementation"(libs.paperLib)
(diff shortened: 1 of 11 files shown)


me4502:
This will likely require bumping the spigot & paper APIs up to 1.21.6 so that the tests pass

SirYwell:
Not sure why both spigot and paper api are needed, but it seems like that doesn't work anymore?

me4502 (approved):

worldedit-bukkit/build.gradle.kts
```diff
@@ -58,6 +53,9 @@ dependencies {
     "compileOnly"(libs.paperApi) {
         exclude("junit", "junit")
     }
+    "testImplementation"(libs.paperApi) {
+        exclude("junit", "junit")
+    }
```
octylFractal, line 58: Can we move these up and change to:
```suggestion
    "localImplementation"(libs.paperApi) {
        exclude("junit", "junit")
    }
```

worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/CommandRegistration.java
```diff
@@ -97,7 +98,11 @@ public CommandMap getCommandMap() {
         if (commandMap == null) {
             Bukkit.getServer().getLogger().severe(plugin.getDescription().getName()
                 + ": Could not retrieve server CommandMap, using fallback instead!");
-            fallbackCommands = commandMap = new SimpleCommandMap(Bukkit.getServer());
+            try {
+                fallbackCommands = commandMap = SimpleCommandMap.class.getConstructor(Server.class).newInstance(Bukkit.getServer());
+            } catch (Throwable e) {
```
octylFractal, line 103: I don't like catching this without logging. Why are we doing this new call anyways? Is this Spigot <-> Paper compat? Can it be documented if so?

octylFractal (changes requested):

worldedit-bukkit/src/main/java/com/sk89q/bukkit/util/CommandRegistration.java
me4502: it's for Spigot/Paper compat yes, afaict this will never actually be called as the above conditions will be met- it's more to just retain this compatibility for whatever niche server does it.

in a Paper-only future this whole thing can just be the simple call above

octylFractal: I still think we should log, even in a fallback. Not logging in these situations has caused issues before when we needed to debug them.

me4502: i've dropped the Paper version here given Paper will never need a fallback anyway, and just opted to throw if this fails. It's a non-recoverable position anyway afaik

octylFractal (approved):

PR #2703, opened 2025-01-25
[Fabric] Workaround old Fabric Perms API issues by explicitly checking version

me4502:
This PR adds a check on the loader for the Fabric Perms API, to ensure that it's at least the current 1.21.4 release. If not, it prints a warning and continues loading.

I've also made any currently uncaught errors here print a warning in the console rather than crashing the game, as we have an available fallback.

### DIFF
--- a/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
+++ b/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
@@ -56,6 +56,9 @@ import net.fabricmc.fabric.api.event.player.UseItemCallback;
 import net.fabricmc.fabric.api.networking.v1.ServerPlayConnectionEvents;
 import net.fabricmc.loader.api.FabricLoader;
 import net.fabricmc.loader.api.ModContainer;
+import net.fabricmc.loader.api.Version;
+import net.fabricmc.loader.api.metadata.ModMetadata;
+import net.fabricmc.loader.api.metadata.version.VersionPredicate;
 import net.minecraft.commands.CommandBuildContext;
 import net.minecraft.commands.CommandSourceStack;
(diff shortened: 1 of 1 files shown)


worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
```diff
@@ -205,10 +208,22 @@ private void registerCommands(CommandDispatcher<CommandSourceStack> dispatcher,
...
+                    .map(ModMetadata::getVersion);
+
+            if (version.isPresent() && !VersionPredicate.parse(">=0.3.3").test(version.get())) {
+                throw new RuntimeException("Fabric permissions version " + version.get() + " is not supported. Please update Fabric Permissions API");
+            }
+
             return new FabricPermissionsProvider.LuckoFabricPermissionsProvider(platform);
         } catch (ClassNotFoundException ignored) {
             // fallback to vanilla
+        } catch (Throwable e) {
```
octylFractal, line 222: We should keep this as `Exception` or we risk suppressing real errors.
```suggestion
        } catch (Exception e) {
```

me4502: My logic here was any error in that mod is something we just want to provide a warn for, rather than actually crash, given we have a fallback available

I'm happy to change it to Exception, but it does partially reduce the intention

octylFractal: An `Error` is likely to be OOM or similar fatal exception, so I wouldn't worry about it. Or it could be an `AssertionError` from us passing bad values somewhere, which should be louder I think.

octylFractal (approved):

Issue #2480, opened 2024-03-07
Deprecate `HttpRequest`, use JDK's `HttpClient` instead


octylFractal:

me4502:
Closed in https://github.com/EngineHub/WorldEdit/commit/afa2f3faa4c4e2653d7d2b60b2d655b1e39ee3f2

PR #2713, opened 2025-02-16
Deprecate HttpRequest in favor of native JDK HttpClient

me4502:
This PR deprecates the HttpRequest class, and swaps its one usage over to the JDK HttpClient class.

I've also taken the opportunity to move the paste callback to use the V2 signed paste API, which has a larger file size cap and doesn't require multipart form encoding.

Closes https://github.com/EngineHub/WorldEdit/issues/2480

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/paste/EngineHubPaste.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/paste/EngineHubPaste.java
@@ -21,11 +21,13 @@ package com.sk89q.worldedit.util.paste;
 
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
-import com.sk89q.worldedit.util.net.HttpRequest;
 
 import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
 import java.util.Map;
 import java.util.concurrent.Callable;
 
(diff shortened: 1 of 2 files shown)


wizjany (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/util/paste/EngineHubPaste.java
```diff
@@ -49,47 +51,49 @@ private PasteTask(String content, PasteMetadata metadata) {
...
-            }
-            if (metadata.author != null) {
-                requestBuilder.header("x-paste-meta-author", metadata.author);
+            try (HttpClient client = HttpClient.newHttpClient()) {
+                HttpRequest.Builder signRequestBuilder = HttpRequest.newBuilder()
+                        .uri(URI.create("https://paste.enginehub.org/signed_paste_v2"))
+                        .header("x-paste-meta-from", "EngineHub");
+
+                if (metadata.name != null) {
+                    signRequestBuilder = signRequestBuilder.header("x-paste-meta-name", metadata.name);
```
wizjany, line 60: Unnecessary reassignment. Builder can be mutated in place.

worldedit-core/src/main/java/com/sk89q/worldedit/util/paste/EngineHubPaste.java
```diff
@@ -49,47 +51,49 @@ private PasteTask(String content, PasteMetadata metadata) {
...
+                }
+                if (metadata.author != null) {
+                    signRequestBuilder = signRequestBuilder.header("x-paste-meta-author", metadata.author);
+                }
+                if (metadata.extension != null) {
+                    signRequestBuilder = signRequestBuilder.header("x-paste-meta-extension", metadata.extension);
+                }
+
+                SignedPasteResponse response = GSON.fromJson(
+                        client.send(signRequestBuilder.build(), HttpResponse.BodyHandlers.ofString()).body(),
```
wizjany, line 70: Previous code required a status code here. If this request fails, will the new code just continue with some (bad?) data? Is this relying on the fact that the json deserialization will fail if the response doesn't have a specific structure? If so, won't it throw a different exception, which doesn't extend IOException and thus will give an ugly result back instead of a converted message?

wizjany (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/util/paste/EngineHubPaste.java
me4502: Fixed.

wizjany (approved):

PR #2715, opened 2025-02-17
Make task executor lifecycled to platform readiness.

### LINKED ISSUE #2459: Fabric with worldedit hangs when stopping during shutdown

wizjany:
Closes #2459.

Note that the platform list (now map) was/is not threadsafe, which in theory might be problematic if platforms throw ready/unready in multiple threads. Ready/unready should be fixed by #2571 if this is an issue.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
@@ -19,6 +19,9 @@
 
 package com.sk89q.worldedit.extension.platform;
 
+import com.google.common.collect.Maps;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.sk89q.worldedit.LocalConfiguration;
 import com.sk89q.worldedit.LocalSession;
 import com.sk89q.worldedit.WorldEdit;
@@ -41,7 +44,9 @@ import com.sk89q.worldedit.session.request.Request;
 import com.sk89q.worldedit.util.HandSide;
 import com.sk89q.worldedit.util.Location;
 import com.sk89q.worldedit.util.SideEffect;
+import com.sk89q.worldedit.util.concurrency.EvenMoreExecutors;
 import com.sk89q.worldedit.util.eventbus.Subscribe;
+import com.sk89q.worldedit.util.lifecycle.SimpleLifecycled;
 import com.sk89q.worldedit.world.World;
 import org.apache.logging.log4j.Logger;
 
@@ -69,7 +74,8 @@ public class PlatformManager {
 
     private final WorldEdit worldEdit;
     private final PlatformCommandManager platformCommandManager;
-    private final List<Platform> platforms = new ArrayList<>();
+    private final SimpleLifecycled<ListeningExecutorService> executorService;
+    private final Map<Platform, Boolean> platforms = Maps.newHashMap();
     private final Map<Capability, Platform> preferences = new EnumMap<>(Capability.class);
     private @Nullable String firstSeenVersion;
     private final AtomicBoolean initialized = new AtomicBoolean();
@@ -84,6 +90,7 @@ public class PlatformManager {
         checkNotNull(worldEdit);
         this.worldEdit = worldEdit;
         this.platformCommandManager = new PlatformCommandManager(worldEdit, this);
+        this.executorService = SimpleLifecycled.invalid();
 
         // Register this instance for events
         worldEdit.getEventBus().register(this);
@@ -101,7 +108,7 @@ public class PlatformManager {
 
         // Just add the platform to the list of platforms: we'll pick favorites
         // once all the platforms have been loaded
-        platforms.add(platform);
+        platforms.put(platform, false);
 
         // Make sure that versions are in sync
         if (firstSeenVersion != null) {
@@ -126,7 +133,7 @@ public class PlatformManager {
     public synchronized boolean unregister(Platform platform) {
         checkNotNull(platform);
 
-        boolean removed = platforms.remove(platform);
+        boolean removed = platforms.remove(platform) != null;
 
         if (removed) {
             LOGGER.info("Unregistering " + platform.getClass().getCanonicalName() + " from WorldEdit");
@@ -212,7 +219,7 @@ public class PlatformManager {
         Platform preferred = null;
         Preference highest = null;
 
-        for (Platform platform : platforms) {
+        for (Platform platform : platforms.keySet()) {
             Preference preference = platform.getCapabilities().get(capability);
             if (preference != null && (highest == null || preference.isPreferredOver(highest))) {
                 preferred = platform;
@@ -231,7 +238,7 @@ public class PlatformManager {
      * @return a list of platforms
      */
     public synchronized List<Platform> getPlatforms() {
-        return new ArrayList<>(platforms);
+        return new ArrayList<>(platforms.keySet());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
```diff
@@ -284,6 +291,21 @@ public PlatformCommandManager getPlatformCommandManager() {
         return platformCommandManager;
     }
 
+    /**
+     * Get the executor service. Internal, not for API use.
+     *
+     * @return the executor service
+     */
+    public ListeningExecutorService getExecutorService() {
```
wizjany, line 299: Potentially the platform manager should expose the `Lifecycled` and the `valueOrThrow` can be moved up to the `WorldEdit` class. That would allow things to add listeners to the `Lifecycled` events if it becomes needed at some point. But also, this isn't API so we can change it at any time if that's needed.

wizjany:
This is now tested on Fabric, Neoforge, and Bukkit; client and server.

Notably, the schematic tasks don't actually listen to interrupts, so a long-running schematic load in single player will actually run to completion when closing a world and loading back into it (or another). However, the running schematic task doesn't affect the new world as the actor/session/etc. is no longer in use. It also doesn't prevent server shutdowns as the executor is no longer waiting for task completion.

me4502 (approved):

Fallen-Breath:
Since #2459 is locked, I'm posting more information on #2459 here

There are 2 bugs in worldedit that made #2459 happen

1. `WorldEdit.getInstance().getExecutorService()` does not shutdown on server stop. This might block the server for up to 60s. This is what this PR fixes
2. `FutureProgressListener` contains a non-daemon `Timer` thread. This might block the server forever

All attempts (#2460 / #2570 / #2592 / #2715) in this repository focus on the 1st bug, not the 2nd one

Believe it or not, the 2nd bug still exists in the latest release 7.3.11, and is still reproduce-able with the steps metioned in https://github.com/EngineHub/WorldEdit/issues/2459#issuecomment-2412020684. Here's the [server log](https://mclo.gs/LNd3MLO) and the [jstack output](https://mclo.gs/chVGfz9)

So, this PR does not close #2459

Please always remember that arrogantly locking a conversation does not truly help solve the problem

me4502:
> Since #2459 is locked, I'm posting more information on #2459 here
> 
> There are 2 bugs in worldedit that made #2459 happen
> 
> 1. `WorldEdit.getInstance().getExecutorService()` does not shutdown on server stop. This might block the server for up to 60s. This is what this PR fixes
> 2. `FutureProgressListener` contains a non-daemon `Timer` thread. This might block the server forever
> 
> All attempts (#2460 / #2570 / #2592 / #2715) in this repository focus on the 1st bug, not the 2nd one
> 
> Believe it or not, the 2nd bug still exists in the latest release 7.3.11, and is still reproduce-able with the steps metioned in [#2459 (comment)](https://github.com/EngineHub/WorldEdit/issues/2459#issuecomment-2412020684). Here's the [server log](https://mclo.gs/LNd3MLO) and the [jstack output](https://mclo.gs/chVGfz9)
> 
> So, this PR does not close #2459
> 
> Please always remember that arrogantly locking a conversation does not truly help solve the problem

The first merged fix explicitly did touch the second issue, and as you were told last time you brought this up, was part of the reason it had to be reverted.

Fallen-Breath:
> The first merged fix originally did touch the second issue

True. But eventaully all of these PRs got ignored or choose to not focus on the 2nd one. Also it's definitely not meaning that "but it doesn't seem to be causing any issues". Please double check and verify the provided reproducle steps before closing #2459

me4502:
> > The first merged fix originally did touch the second issue
> 
> True. But eventaully all of these PRs got ignored or choose to not focus on the 2nd one. Also it's definitely not meaning that "but it doesn't seem to be causing any issues". Please double check and verify the provided reproducle steps before closing #2459

None of us have ever been able to reproduce the issue laid out in there, it seems to only affect a very small subset of users. We can’t verify that it’s fixed, we can only verify that things aren’t more broken than they were prior. And making that Timer a daemon _does_ break things substantially.

Fallen-Breath:
>  but it caused so many major issues

Any example issue caused by daemonizing the timer thread? I haven't encountered any issue after making the Timer thread daemon

Fallen-Breath:
> None of us have ever been able to reproduce the issue laid out in there

If you don't know how to reproduce the issue with steps metioned in https://github.com/EngineHub/WorldEdit/issues/2459#issuecomment-2412020684 and https://github.com/EngineHub/WorldEdit/pull/2715#issuecomment-2759216725, here's a universal reproduce scripts that works in all x86 linux system with docker installed:

```bash
docker run --rm -w /worldedit --name worldedit2459 eclipse-temurin:21.0.6_7-jdk bash -c "$(cat <<EOF

mkdir mods
wget -q https://cdn.modrinth.com/data/1u6JkXh5/versions/bxlboAan/worldedit-mod-7.3.11.jar -O mods/worldedit-mod-7.3.11.jar
wget -q https://maven.fabricmc.net/net/fabricmc/fabric-installer/1.0.1/fabric-installer-1.0.1.jar
wget -q https://piston-data.mojang.com/v1/objects/4707d00eb834b446575d89a61a11b5d548d8c001/server.jar
java -jar fabric-installer-1.0.1.jar server -mcversion 1.21.4

echo "eula=true" > eula.txt
echo -e "searchitem stone\nstop\n" | java -Xmx1G -jar fabric-server-launch.jar

EOF
)"
```

This script will run MC 1.21.4 with latest fabricloader and WorldEdit 7.3.11 in the temurin 21 docker container. [example output](https://mclo.gs/tlGM8It)

After executing this script, you'll get a server that stucks on server-stop forever. You can inspect the java process stack with:

```bash
docker exec worldedit2459 bash -c 'jstack $(pidof java)'
```

There'll be a `Timer-1` thread inside the jstack output. [example output](https://mclo.gs/2Zocp5B)

wizjany:
To the contrary, the fact that saving/loading schematics is a rather common use case but only a few people have ever reported running into this issue is what implies it only affects a small subset of users. The fact that that code has been in use since 2018 ([technically since 2015](https://github.com/EngineHub/WorldGuard/commit/471b4496c947156058da50268e9108604a86984e#diff-35b9ada79408551f82b8eaa6d06279e3fcf29632b7425e2c46456f18f2333223R33)) but this was only reported a year ago also implies that something else (JVM changes? GC specific? server software specific?) might be causing the issue.

I did specifically mention [in my comment above](https://github.com/EngineHub/WorldEdit/pull/2715#issuecomment-2692825958) that I was aware of the discussion but wasn't able to reproduce the issue. While I'm not sure if daemonizing the timer thread would cause issues or not, it didn't seem necessary both historically and in testing, so I didn't do it.

Fallen-Breath:
Okay, I give up on trying to get you to completely fix this issue. I don't have the time or energy to set up a series of environments from 2015 to 2025 and reproduce this issue one by one just to disprove the claim that "if it took this long for an issue to appear, then it must not be a bug in the code"

For anyone who comes across this later and is simply looking for a solution, you can check out [this mod](https://github.com/Fallen-Breath/worldedit-hang-fix). It consists of just two simple mixins and works on all Fabric/Forge/NeoForge platforms since the 1.14 release in 2019

wizjany:
You're either not reading what I said or purposefully misrepresenting it. I never said I'm against making the timer thread a daemon (I don't actually know that it causes issues, but the fact that [making the timer non-static](https://github.com/EngineHub/WorldEdit/issues/2459#issuecomment-1896609710), like the one in sessionmanager, didn't fully fix the issue does seem strange), or that it's not "a bug in the code", just that it's been like that forever and isn't as easily reproducible as you're making it out to be, and that I didn't make that change here because I couldn't reproduce the issue.

Fallen-Breath:
> I never said I'm against making the timer thread a daemon

I also never said you have said that

I'm just tired of this. Whatever you guys think, just let it be. Thanks for maintaining WorldEdit for such a long time

Issue #2719, opened 2025-02-22
Non-persistent entities are pasted incorrectly


brickmonster:
### WorldEdit Version

7.3.10

### Platform Version

1.21.4-177

### Confirmations

- [x] I am using the most recent Minecraft release.
- [x] I am using a version of WorldEdit compatible with my Minecraft version.
- [x] I am using the latest or recommended version of my platform software.
- [x] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [x] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

Entities set to not persist in bukkit are pasted in a default state.

### Expected Behavior

They should either not be pasted, or their state should be copied correctly.

### Reproduction Steps

```
final var entity = (LivingEntity) player.getWorld().spawnEntity(player.getLocation(), org.bukkit.entity.EntityType.ZOMBIE);
entity.setPersistent(false);
entity.setAI(false);
```

Select a region with the spawned entity.
`//copy -e`
`//paste -e`

The zombie will have their AI back, attributes set to defaults, BukkitValues empty, etc.

### Anything Else?

Example with interaction entity. Note the incorrect width/height and missing BukkitValues.

Before:
```
{
	Air: 300s,
	Bukkit.persist: 0b,
	Bukkit.updateLevel: 2,
	BukkitValues: {
		"minecraft:test": 4b
	},
	FallDistance: 0f,
	Fire: -1s,
	Invulnerable: 0b,
	Motion: [0d, 0d, 0d],
	OnGround: 0b,
	Paper.Origin: [1.5d, 4d, 1.5d],
	Paper.OriginWorld: [I; -1736324837, 981615463, -1862036380, -371568799],
	Paper.SpawnReason: "CUSTOM",
	PortalCooldown: 0,
	Pos: [1.5d, 4d, 1.5d],
	Rotation: [0f, 0f],
	Spigot.ticksLived: 110,
	UUID: [I; 2133294147, 930369498, -1087649753, -1186562898],
	WorldUUIDLeast: -7997385352138829983L,
	WorldUUIDMost: -7457458389165915289L,
	height: 0.125f,
	id: "minecraft:interaction",
	response: 0b,
	width: 0.125f
}
```

After:
```
{
	Air: 300s,
	Bukkit.updateLevel: 2,
[...]

PR #2721, opened 2025-02-24
Fix entities with persist=false copied in an invalid state

brickmonster:
Remove a redundant passenger check, as entity.save() returns false in that case.

This also causes leash knots to not be copied. I don't think this is a problem because:
- They would not be saved to disk, it's misleading for users that they appear.
- Pasted leashed mobs still think they're leashed to the original position and get unleashed* - no change in behaviour.
    \* Unless they're pasted close enough to the original position, in which case this has better behaviour because they create their own leash_knot entity.

Fixes https://github.com/EngineHub/WorldEdit/issues/2719

### DIFF
--- a/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/internal/FabricEntity.java
+++ b/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/internal/FabricEntity.java
@@ -52,12 +52,14 @@ public class FabricEntity implements Entity {
     @Override
     public BaseEntity getState() {
         net.minecraft.world.entity.Entity entity = entityRef.get();
-        if (entity == null || entity.isPassenger()) {
+        if (entity == null) {
+            return null;
+        }
+        CompoundTag tag = new CompoundTag();
+        if (!entity.save(tag)) {
             return null;
         }
         ResourceLocation id = FabricWorldEdit.getRegistry(Registries.ENTITY_TYPE).getKey(entity.getType());
-        CompoundTag tag = new CompoundTag();
-        entity.saveWithoutId(tag);
         return new BaseEntity(
             EntityTypes.get(id.toString()),
             LazyReference.from(() -> NBTConverter.fromNative(tag))
--- a/worldedit-bukkit/adapters/adapter-1.21.3/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_21_3/PaperweightAdapter.java
+++ b/worldedit-bukkit/adapters/adapter-1.21.3/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_21_3/PaperweightAdapter.java
@@ -290,8 +290,8 @@ public final class PaperweightAdapter implements BukkitImplAdapter {
      * @param entity the entity
      * @param tag the tag
      */
-    private static void readEntityIntoTag(Entity entity, net.minecraft.nbt.CompoundTag tag) {
-        entity.save(tag);
+    private static boolean readEntityIntoTag(Entity entity, net.minecraft.nbt.CompoundTag tag) {
+        return entity.save(tag);
     }
 
     private static Block getBlockFromType(BlockType blockType) {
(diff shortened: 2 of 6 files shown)


me4502:
Thanks for this, does this issue also affect the other platforms? (NeoForge, Fabric, Sponge)

me4502:
my understanding of this is that it's fixing a wider bug, where the persist tag was basically one of the symptoms on Bukkit. given worldedit is a cross platform mod, generally if something is fixed/changed in one platform it's best to keep it consistently applied across all

wizjany:
I'm a bit perplexed by the minecraft behavior here. the save method returns a bool, but still fills the tag in? doesn't this mean that if, for example, there's extra data on a leash entity, it will get lost after a save/load? does the leashed animal recreate a new one each time?

brickmonster:
Using fabric as a reference here just because I have it open:

Minecraft calls saveSelfNbt and returns false if it's a passenger or if it's not supposed to save that entity type.

WorldEdit instead calls a deeper function, writeNbt, fetching the entity type itself and checking if it's a passenger, but does not check if it's supposed to save the entity type.

And yes the leashed animal creates the leash on spawn.

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/internal/FabricEntity.java
```diff
@@ -52,12 +52,14 @@ public FabricEntity(net.minecraft.world.entity.Entity entity) {
...
     public BaseEntity getState() {
         net.minecraft.world.entity.Entity entity = entityRef.get();
-        if (entity == null || entity.isPassenger()) {
+        if (entity == null) {
             return null;
         }
-        ResourceLocation id = FabricWorldEdit.getRegistry(Registries.ENTITY_TYPE).getKey(entity.getType());
         CompoundTag tag = new CompoundTag();
-        entity.saveWithoutId(tag);
+        if (!entity.save(tag)) {
```
me4502, line 59: is changing this away from the `saveWithoutId` function definitely correct here (and in NF)?

me4502: sorry i was meaning going from `saveWithoutId` -> `save`. i don't have the decompiled source in front of me right now, but from memory there was a reason we were using the `WithoutId` variant here

brickmonster: Oh yeah I do remember it being different in the past but it's really simple now.

The `getState` function in sponge is now a 1:1 copy of what nms does except nms also checks if the entity is removed. I didn't add that because I assumed this codepath wouldn't be hit if it was removed.

brickmonster: Oh and they add the `id` tag

me4502: sorry for the delay; i've been fairly unavailable for a bit

my point was more that we didn't include the ID for a reason, from memory doing that caused a _lot_ of problems. so making this change is likely not safe

brickmonster: No prob.

I've had a look at the 4 implementations. Each of them, which the exception of sponge, just overrides the id tag with the result of WorldEdit's `EntityType.id()` which is to be the same value as the one vanilla puts there (they use the registry).

Sponge instead passes it along, which the implementation then fills in if it's missing.

Removing the id tag before returning would be the least amount of code change and logic change, but I'd also like to propose removing the overrides instead since it's the same value.

brickmonster: Rebased, tested fabric + neoforge + paper 1.21.5.

I've not removed the id tag because said it's fine now, lmk if I should.

me4502 (approved):

PR #2747, opened 2025-04-01
Clarified message on code editing

Mandlo75:


me4502:
I'm really not sure what this is actually clarifying? It seems to make things less clear, because you do not make edits to the code in the readme file..?

Issue #2653, opened 2024-11-04
An option for //distr to only show certain blocks.


mk-pmb:
### The Problem

It's cumbersome to extract the interesting parts of the //distr report when the list spans several pages and most of it is irrelevant to you. (Example use case: Verifying oak stairs in #2652.)

### A Solution

A mask option like `-m` from `//copy`. Only matching blocks would be reported.

### Alternatives

Copy the world, replace irrelevant blocks with air, //distr.

me4502:
Closed by https://github.com/EngineHub/WorldEdit/commit/cc43cc3a964510c626959d11cfeb170b6de9d40e

mk-pmb:
Thanks a lot!

PR #2754, opened 2025-04-10
Add a -m flag to the //distr command to filter included blocks

me4502:
Adds https://github.com/EngineHub/WorldEdit/issues/2653

This PR adds a source mask for the `//distr` command. I was initially semi concerned about it not making as much sense here as with `//count`, but after putting some thought into it I can think of a fair few fairly useful usecases.

Eg, getting the distribution of surface blocks with `//distr -m #surface`

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/block/BlockDistributionCounter.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/block/BlockDistributionCounter.java
@@ -22,9 +22,12 @@ package com.sk89q.worldedit.function.block;
 import com.sk89q.worldedit.WorldEditException;
 import com.sk89q.worldedit.extent.Extent;
 import com.sk89q.worldedit.function.RegionFunction;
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/block/BlockDistributionCounter.java
```diff
@@ -35,18 +38,28 @@
...
+    public BlockDistributionCounter(Extent extent, @Nullable Mask mask, boolean separateStates) {
         this.extent = extent;
+        this.mask = mask == null ? Masks.alwaysTrue() : mask;
         this.separateStates = separateStates;
     }
 
     @Override
     public boolean apply(BlockVector3 position) throws WorldEditException {
+        if (!mask.test(position)) {
+            return true;
```
wizjany, line 60: This should likely return false. I doubt we use the value within WE, but generally RegionFunctions should only return true if "something was changed" (though in this case it's a counter rather than the world).

wizjany (approved):

PR #2757, opened 2025-04-12
Update click and hover text component serialization for 1.21.5.

### LINKED ISSUE #2756: 1.21.5, clickEvent not updated to click_event (same for hoverEvent if any)
### WorldEdit Version

Bukkit-Official(7.3.12-SNAPSHOT+7097-4ac9103)

### Platform Version

This server is running Paper version 1.21.5-DEV-update/1.21.5@e8eba88 (1970-01-01T00:00:00Z) (Implementing API version 1.21.5-R0.1-SNAPSHOT) You are 3 version(s) behind Download the new version at: https://papermc.io/downloads/paper Previous version: 1.21.5-DEV-ce91a8c (MC: 1.21.5)

### Confirmations

### Bug Description

Since 1.21.5, the clickEvent key in text components has been updated to click_event, same for hoverEvent to hover_event.
It seems like you didn't updated yet the `//schem list` command as there are no click or hover events working.

Image

From the [Minecraft 1.21.5 changelog](https://www.minecraft.net/en-us/article/minecraft-java-edition-1-21-5):
```
Hover Events

    The hoverEvent field has been renamed to hover_event
    The legacy value field (which was parsed from a rendered text component) is no longer supported
    For the show_text action:
        The contents field has been renamed to value
        e.g. {action:'show_text',contents:'Hello world'} becomes {action:'show_text',value:'Hello world'}
    For the show_item action:
        The contents field has been inlined
        e.g. {action:'show_item',contents:{id:'minecraft:stick',count:2}} becomes {action:'show_item',id:'minecraft:stick',count:2}
        If contents was specified only as an item id, it is replaced with the full format and inlined
        e.g. {action:'show_item',contents:'minecraft:stick'} becomes {action:'show_item',id:'minecraft:stick'}
    For the show_entity action:
        The contents field has been inlined
        The id field has been renamed to uuid
        The type field has been renamed to id
        e.g. {action:'show_entity',contents:{id:[I;0,0,0,0],type:'minecraft:pig'}} becomes {action:'show_entity',uuid:[I;0,0,0,0],id:'minecraft:pig'}

Click Events

    The clickEvent field has been renamed to click_event
    For the open_url action:
        The value field has been renamed to url
        The click event will no longer parse if not a valid URI with either https:// or http:// schemes, instead of simply not working
        e.g. {action:'open_url',value:'https://minecraft.net'} becomes {action:'open_url',url:'https://minecraft.net'}
    For the run_command action:
        The value field has been renamed to command
        The click event will no longer parse if the command contains disallowed characters, instead of simply not working
        It is no longer required that the specified command field has a / prefix
        e.g. {action:'run_command',value:'/tp @e @s'} becomes {action:'run_command',command:'/tp @e @s'}
    For the suggest_command action:
        The value field has been renamed to command
        The click event will no longer parse if the command contains disallowed characters, instead of simply not working
        e.g. {action:'suggest_command',value:'/help'} becomes {action:'suggest_command',command:'/help'}
    For the change_page action:
        The value field has been renamed to page
        The page value now requires a positive integer instead of a string
        e.g. {action:'change_page',value:'1'} becomes {action:'change_page',page:1}
    The copy_to_clipboard format is unchanged
```

### Expected Behavior

We should be able to change pages and load schematic by clicking

### Reproduction Steps

1. //schem list
2. Move your mouse

### Anything Else?

_No response_

wizjany:
Hacky workaround by overwriting the text3 StyleSerializer class with our own modified copy. Note that the hover events "show_entity" and "show_item" aren't supported as they no longer take a rendered component, but just the NBT structure.

Fixes #2756.

This is really ugly but it might be the easiest solution without updating text3 directly (and figuring out things like show item/entity). ~~Note that this also breaks MC <= 1.21.4 support. If we go with this we can also remove adapters again (lol).~~

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/formatting/text/serializer/gson/StyleSerializer.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/formatting/text/serializer/gson/StyleSerializer.java
@@ -1,0 +1,207 @@
+/*
+ * This file is part of text, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2020 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package com.sk89q.worldedit.util.formatting.text.serializer.gson;
+
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.JsonSerializationContext;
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/util/formatting/text/serializer/gson/StyleSerializer.java
```diff
@@ -0,0 +1,188 @@
...
+
+// This class was copied and manually "relocated" from text3 to update the click/hover events for 1.21.5.
+public class StyleSerializer implements JsonDeserializer<Style>, JsonSerializer<Style> {
+    public static final StyleSerializer INSTANCE = new StyleSerializer();
+
+    private static final TextDecoration[] DECORATIONS = TextDecoration.values();
+
+    static final String COLOR = "color";
+    static final String INSERTION = "insertion";
+    static final String CLICK_EVENT = "click_event";
```
me4502, line 52: How many differences are there between this & the 1.21.4 one? If it's just these constants or a little beyond that- we should be able to fairly trivially just add some version guards here and support <1.21.5

wizjany: there's very few changes in this class. the main issue is that this class has no way to know what MC version is running. usually that's configured by the platform classes, but overriding that would require including a modified version of the platform classes. while we're doing hacky i suppose we could have a public static field in here and have our worldedit platform impls set it to modify the behavior but /shrug.

me4502 (approved):

Issue #1778, opened 2021-06-12
Fix POMs for Sponge


octylFractal:
I did a lot of manual mapping to get the POMs/GMM for Forge and Fabric in line, however I could not immediately do this for Sponge. After Sponge 8 is merged, the POM/GMM for it needs to be revisted.

wizjany:
Seems to have been since fixed, poms/gmm for sponge were checked while reviewing the linked issue.

PR #2761, opened 2025-04-26
Configure PublishingExtension for WorldEdit Sponge

tyhdefu:
Currently, sponge api aren't be published to the maven repository (only the `pom`s are): https://maven.enginehub.org/repo/com/sk89q/worldedit/worldedit-sponge/7.4.0-SNAPSHOT/

Looks like the `worldedit-sponge` is just not configuring publishing:
- Sponge (No `configure<PublishingExtension> `) - https://github.com/EngineHub/WorldEdit/blob/version/7.3.x/worldedit-sponge/build.gradle.kts
- Bukkit (configures it) - https://github.com/EngineHub/WorldEdit/blob/version/7.3.x/worldedit-bukkit/build.gradle.kts#L110
- Fabric (configures it) - https://github.com/EngineHub/WorldEdit/blob/version/7.3.x/worldedit-fabric/build.gradle.kts#L71

Presumably since `worldedit-sponge` is missing:
```kt
configure<PublishingExtension> {
    publications.named<MavenPublication>("maven") {
        from(components["java"])
    }
}
```
It doesn't get the information from the `java-library` plugin to actually include the jar, so it just includes the pom.

Unfortunately I obviously don't have permission to publish the enginehub repository, but perhaps someone could take a look to see if this fix would work?

### DIFF
--- a/worldedit-sponge/build.gradle.kts
+++ b/worldedit-sponge/build.gradle.kts
@@ -90,3 +90,9 @@ tasks.named<ShadowJar>("shadowJar") {
 tasks.named("assemble").configure {
     dependsOn("shadowJar")
 }
+
+configure<PublishingExtension> {
+    publications.named<MavenPublication>("maven") {
+        from(components["java"])
+    }
+}


wizjany:
FWIW publishing behavior is testable locally via the `publishToMavenLocal` task.

octylFractal:
The POMs themselves may have issues (https://github.com/EngineHub/WorldEdit/issues/1778) so a thorough verification of those should be done as well.

tyhdefu:
Thanks for letting me know about `publishToMavenLocal`.
After running this, the JARs were correctly put in a local maven repo, and I could compile my plugin with the worldedit api jar - so I think thats good confirmation.

wizjany (approved):

PR #2763, opened 2025-05-03
Split out CUI protocol

me4502:
This PR starts the process of splitting out the CUI protocol to the separate jar-in-jar library.

I've added an extra method to handle the CUI initialisation message. I'd ideally like to use the CUIPacket here, but the protocol common library still depends on specific MC versions and MC code, which makes it less feasible for use in WorldEdit-Core. Not sure if we want to change something about this in the future, so for now I just pass in the basic data.

~~This cannot be tested without WECUI for 1.21.5~~


worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/FabricWorldEdit.java
```diff
@@ -414,6 +414,17 @@ private void onPlayerDisconnect(ServerGamePacketListenerImpl handler, MinecraftS
...
     }
 
+    private void onCuiPacket(CUIPacket payload, CUIPacketHandler.PacketContext context) {
+        if (!(context.player() instanceof ServerPlayer player)) {
+            // Ignore - this is not a server-bound packet
+            return;
+        }
+
+        LocalSession session = FabricWorldEdit.inst.getSession(player);
+        FabricPlayer actor = FabricAdapter.adaptPlayer(player);
```
wizjany, line 424:
```suggestion
        FabricPlayer actor = FabricAdapter.adaptPlayer(player);
        LocalSession session = WorldEdit.getInstance().getSessionManager().get(actor);
```

worldedit-neoforge/src/main/java/com/sk89q/worldedit/neoforge/NeoForgeWorldEdit.java
```diff
@@ -397,6 +398,16 @@ public void onPlayerLogOut(PlayerEvent.PlayerLoggedOutEvent event) {
         }
     }
 
+    private void onCuiPacket(CUIPacket payload, CUIPacketHandler.PacketContext context) {
+        if (!(context.player() instanceof ServerPlayer player)) {
+            // Ignore - this is not a server-bound packet
+            return;
+        }
+        LocalSession session = NeoForgeWorldEdit.inst.getSession(player);
+        NeoForgePlayer actor = NeoForgeAdapter.adaptPlayer(player);
```
wizjany, line 407:
```suggestion
        NeoForgePlayer actor = NeoForgeAdapter.adaptPlayer(player);
        LocalSession session = WorldEdit.getInstance().getSessionManager().get(actor);
```

wizjany (changes requested):
tbh not sure why those session methods even exist on the mod class, they're not used anywhere (else).

wizjany (approved):

PR #2782, opened 2025-06-03
feat: default flora additions

DereC4:
I was thinking of adding the new 1.21.5 flora for temperate generator (bush) and for desert generation, add the dry grass variants since those are confirmed to fit the generator 

Also added a check for badlands


DereC4:
whoops let me change that

DereC4:
> Thanks for the PR- as this is a feature addition it should be targeting the master branch.
> 
> Re: my comment on the issue- in general PRs to add it are fine, it's more that any major effort doesn't make sense to do. This is a fairly simple PR to add it without any major reworks, so I personally have nothing against it

unrelated but are you the original creator by any chance ?

wizjany:
I don't really think it makes sense to change flora in this way at this point. It was written when there were only two flowers, and updating it properly would be a lot more work than it's worth given feature gen exists now. A partial attempt like this only brings us further from its original state while not really bringing it in line with Minecraft's gen either.

DereC4:
aw ok

PR #2802, opened 2025-07-06
Use NIO in the YAMLProcessor class

me4502:
Minor tech debt PR, cleaning up usage of old IO in the YAMLProcessor class.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/wepif/PermissionsResolverManager.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/wepif/PermissionsResolverManager.java
@@ -30,7 +30,6 @@ import org.bukkit.event.server.PluginDisableEvent;
 import org.bukkit.event.server.PluginEnableEvent;
 import org.bukkit.plugin.Plugin;
 
-import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Method;
 import java.nio.file.Files;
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/util/yaml/YAMLProcessor.java
```diff
@@ -177,10 +197,12 @@ public String getHeader() {
...
     public boolean save() {
-
-        File parent = file.getParentFile();
+        Path parent = path.getParent();
         if (parent != null) {
-            parent.mkdirs();
+            try {
+                Files.createDirectories(parent);
+            } catch (IOException ignored) {
+            }
```
octylFractal, line 205: Add `LOGGER` and:
```suggestion
            } catch (IOException e) {
                LOGGER.warn("Exception while saving yaml file to " + path, e);
            }
```

worldedit-bukkit/src/main/java/com/sk89q/wepif/PermissionsResolverManager.java
```diff
@@ -100,7 +102,7 @@ protected PermissionsResolverManager(Plugin plugin) {
         this.server = plugin.getServer();
         (new ServerListener()).register(plugin); // Register the events
 
-        loadConfig(new File("wepif.yml"));
+        loadConfig(new File("wepif.yml").toPath());
```
octylFractal, line 105: why
```suggestion
        loadConfig(Path.of("wepif.yml"));
```

octylFractal (approved):

PR #2811, opened 2025-07-30
Fix for 2D polygon area algorithm

### LINKED ISSUE #888: getArea() and Polygonal2DRegion - don't return real value.
WORLDEDIT-3096 - Reported by BukkitSmerf

Every Polygonal2DRegion returns other sizes than should, even rectangles:
Example code:

			Polygonal2DRegion polReg = new Polygonal2DRegion();
			polReg.addPoint(new BlockVector2D(1, 1));
			polReg.addPoint(new BlockVector2D(1, 10));
			polReg.addPoint(new BlockVector2D(10, 10));
			polReg.addPoint(new BlockVector2D(10, 1));
			// 10x10 = 100
			sender.sendMessage(Integer.toString(polReg.getArea()));
			// send 89 (9x9) instead of 100

I need that to block creating too big regions/selections.

TiagoFar78:
This PR introduces a new algorithm to calculate the area of 2D polygons using discrete points. The previous implementation was designed for concrete points and did not handle certain cases correctly.

Since the new algorithm is relatively complex, I have also prepared a [detailed explanation](https://drive.google.com/file/d/1CnZRujyKNNEuoJOXsl5oLnGDMPnhrf5U/view) for reference.

Closes #888

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/Polygonal2DRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/Polygonal2DRegion.java
@@ -19,6 +19,7 @@
 
 package com.sk89q.worldedit.regions;
 
+import com.google.common.math.IntMath;
 import com.sk89q.worldedit.math.BlockVector2;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.regions.iterator.FlatRegion3DIterator;
@@ -26,8 +27,6 @@ import com.sk89q.worldedit.regions.iterator.FlatRegionIterator;
 import com.sk89q.worldedit.util.formatting.text.TranslatableComponent;
 import com.sk89q.worldedit.world.World;
 
-import java.math.BigDecimal;
-import java.math.RoundingMode;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


me4502:
I've done some testing of this PR, and it definitely seems vastly better than the existing behaviour, thank you for this! I'm approving it based on code changes & my testing- but still would prefer someone more maths-focused to take a look before I merge it :)

me4502 (approved):

TiagoFar78:
> 2. For each of the now-simple polygons, calculate the number of boundary integer points, _b_.
> 3. Calculate the area of the simple polygon using the [Shoelace formula](https://en.wikipedia.org/wiki/Shoelace_formula), _A_.
> 4. Calculate the number of integer points inside by inversing [Pick's theorem](https://en.wikipedia.org/wiki/Pick%27s_theorem): _i_ = _A_ - _b_ / 2 + 1
> 5. The area is then _i_ + _b_, and the volume is that times the height.

These steps produce the same result as my current algorithm for non-self-crossing polygons, but with significantly less complexity and without the second limitation of my approach. 

I think it’s a great idea, and with your permission, I’d love to implement it. I’d be happy to add you as a co-author to the commit as well.

---

Regarding the self crossing polygons...

> I saw in the limitations that this doesn't handle self-intersecting polygons. We do allow that, so I'm not sure this code is better or worse than the previous one for that case.

Both approaches behave the same here: the previous algorithm also did not support self-crossing polygons. I assumed this was not required since the original version did not address it either.

> 1. Triangulate (or at least simplify) the polygon using some method that can handle self-intersections.

I don’t think triangulation is the right approach. For example, in the figure below, the orange points would be counted twice, once per triangle, leading to double counting:

image

> 1. I think we can get away with just taking the outer boundary as a whole. Perhaps we can use https://github.com/AngusJohnson/Clipper2's union operation or implement something based off the [Bentley–Ottmann algorithm](https://en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm)?

This seems like the ideal solution to me as well. I’ll try moving forward with that approach.

Thanks for the great idea!

octylFractal:
Please feel free to implement it. Add me as co-author if it's easy for you.

octylFractal:
I will look more into it in the next few weeks, when I have time.

octylFractal (changes requested):

octylFractal (approved):

octylFractal:
Thank you!

PR #2813, opened 2025-08-01
Fix to ConvexPolyhedralRegion started with a line

### LINKED ISSUE #1950: Can't add more vertices to convex region after creating a line of vertices.
### WorldEdit Version

Bukkit-Official(7.2.7+9f3e795)

### Platform Version

git-Paper-788

### Confirmations

### Bug Description

When adding vertices to a convex region, if you add them in a line no other vertices will be added.

### Expected Behavior

To add extra vertices.

### Reproduction Steps

Start a convex region selection.
Add a vertex to the left.
Add a vertex to the right.
Try to add more vertices somewhere else (Part that doesn't work).

### Anything Else?

Issue is still present using 7.2.8-SNAPSHOT (7.2.8-SNAPSHOT+5982-1aad921) and 7.3.0-SNAPSHOT (7.3.0-SNAPSHOT+5979-de6fa17).

Video of the issue: (Shown using FAWE, but confirmed the issue still happens using default WE)
https://user-images.githubusercontent.com/32806171/143670759-d7e2ff4d-fa56-4708-b5ea-f3651e390588.mp4

TiagoFar78:
## Summary
Fixes an issue where additional vertices could not be added to a `ConvexPolyhedralRegion` if the three initial vertices formed a straight line.

## Problem
Previously, the region attempted to create a triangle from the first three vertices. If these vertices were collinear, the convex hull logic treated subsequent vertices as part of the backlog, without ever integrating them into the mesh.

## Solution
- Added a collinearity check when processing the third vertex.
- If the third vertex is collinear with the first two, it is now added to the backlog for later processing.
- New points that extend beyond the initial collinear set can now form valid triangles and update the convex hull.

## Related Issue
Closes #1950


worldedit-core/src/main/java/com/sk89q/worldedit/regions/ConvexPolyhedralRegion.java
```diff
@@ -217,6 +219,26 @@ public boolean addVertex(BlockVector3 vertex) {
         return true;
     }
 
+    private boolean isCollinear(Set<BlockVector3> ab, BlockVector3 c) {
+        Iterator<BlockVector3> it = ab.iterator();
+        BlockVector3 a = it.next();
```
me4502, line 224: It might make sense to give this a name more explicit to only working for vertex lists that only have two elements- otherwise it might lead to someone using this in the future expecting it to detect n-length collinearity. And then if that's done, it'd likely make sense to add a guard to the method to throw if a list of non-2 length is given

TiagoFar78:
> In testing, I'm not sure this entirely fixes the problem. It seems that it now ignores the third point, rather than creating a single line for the convex.

Thanks for testing this. I understand what you’re seeing.

This behavior is actually consistent with how the region currently handles degenerate configurations: when the fourth added point is coplanar with the previous three points, that point isn't immediately integrated into the mesh. Instead, they remain in the backlog until a point is added that allows the region to form a valid 3d region.

image

Even if this fourth point extends beyond the area covered by the initial three, the convex hull logic waits for enough points to define a proper 3D region before incorporating it.

image

That’s why in my change, the third collinear point follows the same rule. It’s held back until the hull has volume. Once a non-coplanar point is added, the previously skipped points are incorporated correctly.

image

If you prefer, I can change it so that the point in the middle is the one ignored, keeping the two outermost points active.

me4502:
That makes sense, thanks for explaining. Given the new behaviour in this PR is most similar to the existing behaviour, I feel it's good for now.

me4502 (approved):

PR #2815, opened 2025-08-03
Fix crafter block rotation

### LINKED ISSUE #2686: Crafter block isn´t rotated with //rotate (Fabric, 1.21.4)
### WorldEdit Version

worldedit-mod-7.3.10-beta-01.jar

### Platform Version

fabric-api-0.113.0+1.21.4

### Confirmations

### Bug Description

Copy a structure with a crafter block, rotate it by 180° and it will not get rotated

### Expected Behavior

Crafter output side also gets rotated

### Reproduction Steps

1. Build Structure with Crafter
2. Select the Structure
3. //copy
4. //rotate 180
5. //paste
6. Structre does not longe work because the Crafter spits out on the wrong side

### Anything Else?

[AutoCrafter.txt](https://github.com/user-attachments/files/18264638/AutoCrafter.txt)
My example Structure (Rename .txt to .schem) and place in `/config/worldedit/schematics`

### LINKED ISSUE #2814: //flip incorrectly flips crafters when orientation should remain unchanged
### WorldEdit Version

7.3.17-SNAPSHOT

### Platform Version

fabric-api-base 0.4.64 + 1.21.7

### Confirmations

### Bug Description

When using //flip on a region containing a crafter, the crafter’s facing direction is flipped even when the flip axis should not alter its orientation.

In the screenshot below, the original crafters are labeled 1 and 2, with the red glass acting as the flip axis.
- Crafter 3 is flipped correctly.
- Crafter 4 should still be facing the player, but it is flipped to face away.
- Crafter 5 is correct.
- Crafter 6 should also be facing the player, but it is flipped incorrectly.

Image

### Expected Behavior

Crafters should not have their facing direction changed when a flip along the chosen axis would not affect their orientation.

### Reproduction Steps

1. Select a region containing two crafters, one facing north/south, the other facing east/west.
2. //copy
3. //flip north
4. //paste
5. //copy
6. //flip east
7. //paste

### Anything Else?

Discovered this bug while trying to find a solution to #2686

TiagoFar78:
## Summary
Fixes an issue where crafters did not rotate correctly due to their orientation property not being handled like other directional blocks.

## Problem
Previously, WorldEdit’s block transformation logic primarily handled blocks with a facing property. Crafters instead use an orientation property. When rotating a crafter, the property value was left unchanged, causing the block to keep its original facing regardless of the applied transformation.

## Solution
Added support for transforming the orientation property.

## Related Issue
Closes #2686
Closes #2814

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
@@ -213,36 +213,28 @@ public class BlockTransformExtent extends AbstractDelegateExtent {
                     }
                 } else if (property.getName().equals("orientation") && transform instanceof AffineTransform affineTransform) {
                     // crafters
-                    if (affineTransform.isHorizontalFlip()) {
-                        String value = (String) result.getState(property);
-                        String newValue = switch (value) {
-                            case "north_up" -> "south_up";
-                            case "south_up" -> "north_up";
-                            case "east_up" -> "west_up";
-                            case "west_up" -> "east_up";
-                            default -> null;
-                        };
-                        if (newValue != null && enumProp.getValues().contains(newValue)) {
-                            result = result.with(enumProp, newValue);
-                        }
+                    String current = (String) result.getState(property);
+
+                    String[] parts = current.split("_");
+                    Direction facing = Direction.valueOf(parts[0].toUpperCase(Locale.ROOT));
+                    Direction top = Direction.valueOf(parts[1].toUpperCase(Locale.ROOT));
+
+                    Vector3 newFacingVec = transform.apply(facing.toVector());
+                    Vector3 newTopVec = transform.apply(top.toVector());
+
+                    Direction newFacing = Direction.findClosest(newFacingVec, Direction.Flag.CARDINAL | Direction.Flag.UPRIGHT);
+                    Direction newTop = Direction.findClosest(newTopVec, Direction.Flag.CARDINAL | Direction.Flag.UPRIGHT);
+                    if (newTop.toString().equals(Direction.DOWN.toString())) {
+                        newTop = Direction.UP;
                     }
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
```diff
@@ -213,36 +213,28 @@ public static <B extends BlockStateHolder<B>> B transform(B block, Transform tra
...
+
+                    String[] parts = current.split("_");
+                    Direction facing = Direction.valueOf(parts[0].toUpperCase(Locale.ROOT));
+                    Direction top = Direction.valueOf(parts[1].toUpperCase(Locale.ROOT));
+
+                    Vector3 newFacingVec = transform.apply(facing.toVector());
+                    Vector3 newTopVec = transform.apply(top.toVector());
+
+                    Direction newFacing = Direction.findClosest(newFacingVec, Direction.Flag.CARDINAL | Direction.Flag.UPRIGHT);
+                    Direction newTop = Direction.findClosest(newTopVec, Direction.Flag.CARDINAL | Direction.Flag.UPRIGHT);
```
TiagoFar78, line 226: Currently, if the transformed orientation is not exactly valid for the crafter, it is snapped to the nearest valid orientation (because of `Direction.Flag.CARDINAL | Direction.Flag.UPRIGHT`) rather than being left unchanged (if it was `Direction.Flag.ALL`). I’m looking for feedback on whether this snapping behavior is preferred, or if it would be better to leave the orientation unchanged in such cases, since I’m not even sure if WorldEdit supports rotations by 45° or other non-cardinal angles.

me4502: WorldEdit does support non-cardinal angles (although not perfectly, so some oddities are fine), so I feel the snapping behaviour is ideal here. It's better to "try" coming up with something close, rather than just not making any change.

PR #2816, opened 2025-08-04
Fix rails rotation

### LINKED ISSUE #2657: Rails do not rotate when rotating a schematic
### WorldEdit Version

7.3.8

### Platform Version

Paper 1.21.1-131-ver/1.21.1@84281ce (2024-10-31T17:43:44Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)

### Confirmations

### Bug Description

When using Clipboard#transform in the api to rotate a schematic, rails in that schematic don't rotate accordingly. 
(Same happens when using `//rotate` command to rotate it in the clipboard)

### Expected Behavior

Rails should rotate with the rest of the schematic instead of keeping their rotation.

### Reproduction Steps

1. Save a schematic with rails in it
2. Load it into the clipboard
3. Rotate it by 90 degrees (`//rotate 90`)
4. Paste it
5. See how the rail rotation hasn't changed according to the rotation

### Anything Else?

_No response_

TiagoFar78:
## Summary
Fixes an issue where rails did not rotate correctly because their shape property was not being taken into consideration when doing transformations.

## Problem
Previously, WorldEdit’s block transformation logic did not account for the fact that rails use a shape property instead of a rotation or facing property. As a result, many rail shapes remained unchanged after rotation or flip transformations.

## Solution
- Added vertical flip handling for ascending rail shapes.
- Split the shape property into two direction components and applied the transformation to each separately.

## Related Issue
Closes #2657

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java
@@ -211,6 +211,45 @@ public class BlockTransformExtent extends AbstractDelegateExtent {
                             result = result.with(enumProp, newValue);
                         }
                     }
+
+                    // rails
+                    if (affineTransform.isVerticalFlip()) {
+                        String value = (String) result.getState(property);
+                        String newValue = switch (value) {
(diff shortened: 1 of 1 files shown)


me4502 (approved):
Thank you for this! :)

PR #2820, opened 2025-08-09
Fix item frame rotation

### LINKED ISSUE #1727: Down/Up Facing Item Frames have incorrect rotations
### WorldEdit Version

7.2.3

### Platform Version

Spigot

### Bug Description

When copying, rotating, and pasting a clipboard (or schematic) that contains an upward or downward facing item frame, the item frame has an unexpected rotation behavior (at least to me?)

The behavior can be observed in [this image](https://cdn.discordapp.com/attachments/139247849350103040/830936056295063582/unknown.png)

The block and item frame on the left is the original clipboard, and the right is a rotated version of it. I will point out the item in the item frame does get rotated too.

### Expected Behavior

I would expect an item frame pointing up or down, when being rotated horitontally, would remain facing the way it is, but the item in the frame would rotate. 

### Reproduction Steps

1. Place an item frame facing up or down
2. Copy it
3. Rotate it
4. Paste it

TiagoFar78:
## Summary
Fixes an issue where item frames’ item would not rotate after applying transformations.

## Problem
Previously, WorldEdit’s entity transformation logic for item frames did not take into consideration the item in an item frame.

## Solution
Implemented the item rotation calculation to operate entirely in the frame’s local plane. Used a consistent local axis basis to determine the correct clockwise rotation relative to the frame’s facing.

## Related Issue
Closes #1727


me4502:
Thank you for this! Item frames seem to behave correctly with this PR. However, it seems this breaks painting rotation. I've done `//copy -e`, `//rotate 180`, `//paste -e`, with a selection of a 2x2 painting on a 2x2 wall of stone- and the painting didn't rotate, and therefore broke after a few ticks.

Issue #2550, opened 2024-06-12
//update command does not respect block limit


DNAmaster10:
### WorldEdit Version

7.3.3

### Platform Version

Paper-1.20.6-138

### Confirmations

- [X] I am using the most recent Minecraft release.
- [x] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [X] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

The //update command does not respect max block restrictions, meaning it's possible to run the //update command over a larger region than is configured.

### Expected Behavior

If the block limit is set to 1000, world edit shouldn't update more than 1000 blocks.

### Reproduction Steps

1. Set a block limit
2. Select a region larger than the block limit
3. Run the //update command on that region

### Anything Else?

This can cause issues for server performance when the command is executed over a very large area.

PR #2836, opened 2025-09-10
Limit //update by block change limit

me4502:
This PR adds a new `ChangeLimitingFilter` region function, which is used to add a limit to the `//update` command. Fixes https://github.com/EngineHub/WorldEdit/issues/2550

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/util/ChangeLimitingFilter.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/util/ChangeLimitingFilter.java
@@ -44,19 +44,17 @@ public final class ChangeLimitingFilter implements RegionFunction {
      */
     public ChangeLimitingFilter(int limit, RegionFunction function) {
         checkNotNull(function);
-        checkArgument(limit >= -1, "limit >= -1 required");
+        checkArgument(limit >= 0, "limit >= 0 required");
         this.limit = limit;
         this.function = function;
     }
 
     @Override
     public boolean apply(BlockVector3 position) throws WorldEditException {
-        if (limit >= 0) {
-            if (count >= limit) {
-                throw new MaxChangedBlocksException(limit);
-            }
-            count++;
+        if (count >= limit) {
+            throw new MaxChangedBlocksException(limit);
         }
+        count++;
         return function.apply(position);
     }
 


worldedit-core/src/main/java/com/sk89q/worldedit/function/util/ChangeLimitingFilter.java
```diff
@@ -0,0 +1,63 @@
...
+import com.sk89q.worldedit.function.RegionFunction;
+import com.sk89q.worldedit.math.BlockVector3;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A filter that limits the number of changes that can be made.
+ */
+public class ChangeLimitingFilter implements RegionFunction {
```
octylFractal: Mark `final`, as we prefer composition over inheritance for these types I think.

worldedit-core/src/main/java/com/sk89q/worldedit/function/util/ChangeLimitingFilter.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Create a new change limiting filter.
+     *
+     * @param limit the limit
+     * @param function the function
+     */
+    public ChangeLimitingFilter(int limit, RegionFunction function) {
+        checkNotNull(function);
+        checkArgument(limit >= -1, "limit >= -1 required");
```
octylFractal: Would it make more sense to enforce `>= 0` instead, and you should just not add this if your limit is less than that? It would reduce potential branches in a hot area.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/function/util/ChangeLimitingFilter.java
me4502: So I've not changed this, but I've made our usage of it be conditionally applied. IMO it doesn't make sense to restrict this beyond valid values

octylFractal: I don't really get why a negative value is valid though. If it rejected it, the branch in apply could be removed.

PR #2849, opened 2025-11-04
Implement convex selection enhancements and error handling for duplicate vertices and limits. (Resolve #2817)

### LINKED ISSUE #2817: Specify how the convex select mode is meant to work, to fix edge cases
### The Problem

Currently the convex selection mode is complicated, and not documented in _exactly_ how it's meant to work. There are inconsistencies/oddities in the behaviour, but that's just "how it's always worked."

### A Solution

If we come up with a strict specification, we'll be able to ensure that behaviour keeps to that spec.

### Alternatives

_No response_

### Anything Else?

_No response_

ssquadteam:


octylFractal:
No response was given, so closing.

PR #2851, opened 2025-11-13
Add a revolve command

me4502:
I'm not 100% clear on the actual requirements of this command, but I've added a `//revolve` command that will revolve the selection around the placement position.

It ensures that each revolution is evenly spaced around the edge, so if it's 4 revolutions it'll be in the four corners- if it's 16 it'll be 16 evenly spaced entries, etc.

Due to WorldEdit not supporting interpolation of clipboard rotations, this does not behave _super_ nicely with higher revolution counts at small sizes, however this does not appear to be the main usecase of the command.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
@@ -287,9 +287,11 @@ public class ClipboardCommands {
         checkRegionBounds(region, session);
         checkCommandArgument(pasteCount >= 2, TranslatableComponent.of("worldedit.revolve.too-few-pastes"));
 
+        BlockVector3 pasteOrigin = session.getPlacementPosition(actor);
+
         // Copy the selection into a clipboard
         BlockArrayClipboard clipboard = new BlockArrayClipboard(region);
-        clipboard.setOrigin(session.getPlacementPosition(actor));
+        clipboard.setOrigin(pasteOrigin);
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
```diff
@@ -265,4 +266,58 @@ public void clearClipboard(Actor actor, LocalSession session) {
...
+        }
+        Operations.complete(copy);
+
+        double perPasteDegrees = 360.0 / pasteCount;
+        ClipboardHolder holder = new ClipboardHolder(clipboard);
+
+        // Now paste it multiple times, rotating each time
+        for (int i = 1; i < pasteCount; i++) {
+            AffineTransform transform = new AffineTransform();
+            transform = transform.rotateY((reverse ? 1 : -1) * perPasteDegrees * i);
```
octylFractal: Can we do this as:
```suggestion
            AffineTransform transform = new AffineTransform().rotateY((reverse ? 1 : -1) * (360 * i) / (double) pasteCount);
```
I think we're much less likely to lose precision as much scaling up first, then scaling down; versus the other way around which at large numbers might start to get very small `perPasteDegrees`. This way, even at hundreds of thousands of pastes, it will still be quite accurate, as we keep integers up to the division.

Also, I don't think reassigning to this variable is very useful, so just in-lining the initialization.

octylFractal (approved):

PR #2852, opened 2025-11-14
Fix pos commands not clamping to world bounds

### LINKED ISSUE #1270: pos1/2 commands do not clamp to world bounds
**Describe the bug**

Using the commands `//pos1` and `//pos2` while outside world bounds (<0, >255) will set the pos to that y value.

**To Reproduce**

1. a. Fly up to y=256 or go below y=0 and use `//pos1` and `//pos2`
1. b. Alternative: Run `//pos1 0,-10,0` and `//pos2 0,260,0`. Run `//expand 1 u` and observe the negative expand due to clamped region size.

**Expected behavior**

Expect the selection to get clamped to world bounds, such as what happens when using other selection commands like //expand, //shift. Alternative would be erroring instead.

**Screenshots**

https://i.imgur.com/nqAL56g.png

**Additional context**

DataValidatorExtent ensures that gets/sets to the selection return air/noop. This does allow some interesting things like WG regions that are below bedrock etc. though. Unsure if that's problematic for anything. This is mostly just to be consistent with selection modification commands. Possibly a toggle/option of some sort could be added to allow out-of-bounds selections more generally?

TiagoFar78:
# Summary

Fixes an issue where `//pos`, `//pos1` and `//pos2` allowed Y-coordinates outside world bounds. Adds a session toggle and command to enable or disable y clamping.

# Problem

`//pos`, `//pos1` and `//pos2` accepted Y values below min and above max world boundaries. When this happened, the selection stored those coordinates instead of clamping or erroring. This led to inconsistent behavior with other selection commands and could produce negative expand operations.

# Solution

Positions set through commands are now clamped to the valid world height range. A session toggle and corresponding command were added to control whether this clamping behavior is active.

# Related Issue

Closes #1270

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
@@ -449,6 +449,31 @@ public class GeneralCommands {
     }
 
     @Command(
+        name = "/ignoreybounds",
+        aliases = {"/iyb"},
+        desc = "Toggles or sets whether selections can extend outside the world's Y boundaries"
+    )
+    @CommandPermissions("worldedit.ignoreybounds")
+    public void ignoreYBounds(Actor actor, LocalSession session,
+                                     @Arg(desc = "If specified, sets the new state. If omitted, the state will be toggled.", def = "")
+                                     Boolean state) {
+        boolean isUnchanged = false;
+
+        if (state != null) {
+            isUnchanged = session.isIgnoringYBounds() == state;
+        } else {
+            state = !session.isIgnoringYBounds();
+        }
+
+        session.setIgnoreYBounds(state);
+
+        String baseKey = "worldedit.ignoreybounds";
+        String stateKey = state ? ".enabled" : ".disabled";
+        String alreadyKey = isUnchanged ? ".already" : "";
+        actor.printInfo(TranslatableComponent.of(baseKey + stateKey + alreadyKey));
+    }
+
+    @Command(
         name = "searchitem",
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -448,6 +448,31 @@ public void placement(Actor actor, LocalSession session,
         placementImpl(actor, session, new Placement(placementType, offset));
     }
 
+    @Command(
+        name = "/ignoreybounds",
+        aliases = {"/iyb"},
+        desc = "Toggles or sets whether selections can extend outside the world's Y boundaries"
```
octylFractal: This description is kind of misleading. It implies that e.g. `//outset` would also be allowed to go beyond the Y bounds, but this is not true since we will clamp it in the region impl with `recalculate`.

Now, this could just be reworded to state it only affects `pos` commands, but I think if there's going to be support for out-of-bounds regions, it should be comprehensive, not a one-off thing we check only at the command level, relying on us incidentally not recalculating regions in certain places. The regions should be constructed with a flag to control if they clamp, so that flipping this makes _all_ selection interactions clamp or not.

Potentially this also implies that certain commands may need re-implementation to support working outside the worlds Y bounds.

TiagoFar78:
> Also, I'm not sure this change makes sense being as small as it is, see the other comment I left in the review.

I understand. Should I drop the PR?

octylFractal:
It's up to you if you want to try to rework it or not.

TiagoFar78:
I may come back to this in the future.

PR #2857, opened 2025-11-29
Fix: rail detection

MCRcortex:
Fix broken rail detection and lower rail up down flipping into the if statement


me4502 (approved):
Thank you :)

me4502:
This currently appears to be failing checkstyle sorry

PR #2869, opened 2025-12-30
added "hand", "offhand" and "pos1" to the pattern suggestion

### LINKED ISSUE #2868: Hand, offhand and pos1 aren't suggested in block patterns
### The Problem

only the registered blocks are suggested, but `hand`, `offhand `and `pos1` are valid patterns too

### A Solution

add `hand`, `offhand `and `pos1` to the getSuggestions method of the DefaultBlockParser

### Alternatives

_No response_

### Anything Else?

_No response_

Adrien5902:
resolves #2868

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
@@ -62,7 +62,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.stream.Stream;
 
-import static org.enginehub.piston.converter.SuggestionHelper.limitByPrefix;
+import static org.enginehub.piston.converter.SuggestionHelper.byPrefix;
 
 /**
  * Parses block input strings.
@@ -213,7 +213,7 @@ public class DefaultBlockParser extends InputParser<BaseBlock> {
                     : Stream.of("pos1");
 
             return Stream.concat(
-                    limitByPrefix(additionalSuggestions, input).stream(),
+                    additionalSuggestions.filter(byPrefix(input)),
                     SuggestionHelper.getNamespacedRegistrySuggestions(BlockType.REGISTRY, input)
             );
         }


worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -206,7 +207,36 @@ private static Map<Property<?>, Object> parseProperties(BlockType type, String[]
     public Stream<String> getSuggestions(String input, ParserContext context) {
         final int idx = input.lastIndexOf('[');
         if (idx < 0) {
-            return SuggestionHelper.getNamespacedRegistrySuggestions(BlockType.REGISTRY, input);
+            ArrayList<String> playerSpecificBlocks = new ArrayList<>();
+
+            var actor = context.getActor();
+            if (actor instanceof Player player) {
+                try {
```
me4502: I'm a little confused by these try statements wrapping it, and why you're calling the getBlockInHand methods here? if it's a player, you should be able to _always_ add `hand` and `offhand` to this list.

For `pos1`, it's realistically better to just always provide it as well. It's not only players who can have selections, so pos1 is _always_ a valid selection. It's better to suggest it and have it give the "need to select a region" message upon running it, as that informs the user that it's actually possible.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -206,7 +207,36 @@ private static Map<Property<?>, Object> parseProperties(BlockType type, String[]
     public Stream<String> getSuggestions(String input, ParserContext context) {
         final int idx = input.lastIndexOf('[');
         if (idx < 0) {
-            return SuggestionHelper.getNamespacedRegistrySuggestions(BlockType.REGISTRY, input);
+            ArrayList<String> playerSpecificBlocks = new ArrayList<>();
```
me4502: This should be `List<String> playerSpecificBlocks = new ArrayList<>();` to conform with general styling

me4502 (changes requested):
Thanks for this, I'm just a little confused about a few parts of this so I've left a few comments :)

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -206,7 +206,10 @@ private static Map<Property<?>, Object> parseProperties(BlockType type, String[]
     public Stream<String> getSuggestions(String input, ParserContext context) {
         final int idx = input.lastIndexOf('[');
         if (idx < 0) {
-            return SuggestionHelper.getNamespacedRegistrySuggestions(BlockType.REGISTRY, input);
+            return Stream.concat(
+                Stream.of("hand", "offhand", "pos1"),
```
octylFractal: Shouldn't hand/offhand only be suggested if the actor is a player? This can be called in a console context so I don't think you can assume it's a player.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -57,9 +57,7 @@
 import org.enginehub.linbus.stream.exception.NbtParseException;
 import org.enginehub.linbus.tree.LinCompoundTag;
 
-import java.util.HashMap;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
```
me4502: This is currently breaking the build, can you please revert the change to the imports?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -57,10 +57,12 @@
 import org.enginehub.linbus.stream.exception.NbtParseException;
 import org.enginehub.linbus.tree.LinCompoundTag;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.HashMap;
-import java.util.Locale;
 import java.util.Map;
 import java.util.stream.Stream;
+import java.util.Locale;
```
me4502: It looks like these imports are still out of order- Locale shouldn't have moved; it's probably worth ensuring that it builds locally and making the fixes until it actually builds first

octylFractal (approved):

PR #2870, opened 2025-12-31
Reduce allocations during hot property code

me4502:
This PR swaps a few different maps over to the Object2ObjectArrayMap during property code, and cleans up the "immutability state"- reducing the number of times we actually need to create wrapper objects for immutability. As some of these calls happen relatively often in hot code, this should reduce allocation count during edits during specific situations such as type applying pattern.

Sorting of properties is also moved to an earlier step of the startup process, which might speed things up a bit during startup.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockType.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockType.java
@@ -55,16 +55,7 @@ public class BlockType implements Keyed {
         = LazyReference.from(() -> new FuzzyBlockState(this));
     @SuppressWarnings("this-escape")
     private final LazyReference<Map<String, ? extends Property<?>>> properties
-        = LazyReference.from(() -> {
-            var propertiesMap = WorldEdit.getInstance().getPlatformManager()
-                    .queryCapability(Capability.GAME_HOOKS).getRegistries().getBlockRegistry().getProperties(this);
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockType.java
```diff
@@ -54,8 +55,16 @@ public class BlockType implements Keyed {
...
-        .queryCapability(Capability.GAME_HOOKS).getRegistries().getBlockRegistry().getProperties(this)));
+        = LazyReference.from(() -> {
+            var propertiesMap = WorldEdit.getInstance().getPlatformManager()
+                    .queryCapability(Capability.GAME_HOOKS).getRegistries().getBlockRegistry().getProperties(this);
+            List<String> sortedPropertyNames = propertiesMap.keySet().stream().sorted().toList();
+            Map<String, Property<?>> sortedPropertiesMap = new Reference2ObjectArrayMap<>(propertiesMap.size());
+            for (String propertyName : sortedPropertyNames) {
+                sortedPropertiesMap.put(propertyName, propertiesMap.get(propertyName));
+            }
+            return Collections.unmodifiableMap(sortedPropertiesMap);
```
octylFractal: Can we pull this out to a static method?

me4502: I can pull it out to a method, not static though as it requires a "this" reference

octylFractal (approved):

PR #2871, opened 2025-12-31
Make properties into records

me4502:
This PR turns properties into records, for current & future performance improvements with JVM optimisations (and eventual value class conversion).

This is theoretically a breaking change, but not within general API surface area. It's a break to anything that was extending AbstractProperty creating their own Property type, which in general shouldn't really be done- so this change is viewed as acceptable.


octylFractal (approved):

SirYwell:
Note that this makes subsequent changes to any of the fields impossible (e.g., using an `(Enum)Set` in `DirectionalProperty`) while not providing any actual performance benefit.

me4502:
These classes need to be as absolutely thin/direct as possible, to avoid memory overhead issues on modded platforms, so this doesn't actually change anything about feasibility of altering underlying fields. We've intended to move these towards some form of set (likely sequenced) in WorldEdit 8. 

This is also step one of other changes, such as making the base Property interface sealed. Plus, in even rough simple testing there was found to be a measurable performance improvement, one that likely becomes more measurable in more complex (especially heavily modded) situations.

PR #2877, opened 2026-01-02
Update Checkstyle and add ErrorProne

octylFractal:


worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolCommands.java
```diff
@@ -300,11 +300,11 @@ public void farwand(Player player, LocalSession session) throws WorldEditExcepti
     )
     @CommandPermissions("worldedit.tool.lrbuild")
     public void longrangebuildtool(Player player, LocalSession session,
-                                   @Arg(desc = "Pattern to set on left-click")
+                                   @Arg(desc = "Pattern to set on main hand interaction")
                                        Pattern primary,
-                                   @Arg(desc = "Pattern to set on right-click")
+                                   @Arg(desc = "Pattern to set on off-hand interaction")
```
me4502: I'm not sure main/off hand interactions are what happens here though right? that's referring to the hand an item is held in- left/right click _better_ correspond to these. it's tied to place/hit basically

Issue #2658, opened 2024-11-13
[Potential Bug] World Edit using a severe amount of RAM per blockstate


Barerock:
### WorldEdit Version

7.2.15

### Platform Version

Forge 47.3.0

### Confirmations

- [ ] I am using the most recent Minecraft release.
- [x] I am using a version of WorldEdit compatible with my Minecraft version.
- [ ] I am using the latest or recommended version of my platform software.
- [x] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [x] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

Additional Placements dev, FirEmerald, found through my memory leak bug report that in fact, World Edit is using an incredible amount of RAM per blockstate ingame. This causes extensively modded instances to require upwards of 16GB to load worlds, which is an unplayable amount because of GC lag. My own Heapdump reflected similar data to what FirEmerald found in their tests. They know a lot more technical info than I do about this issue, so perhaps they can provide more relevant and enlightening data.

AP adds more blockstates; however, World Edit seems to be at fault for the RAM bloat described in this report:
https://github.com/FirEmerald/AdditionalPlacements/issues/96

### Expected Behavior

I'm not sure. This may be an unfortunate side effect of World Edit's functions, in which case this is an unavoidable incompatibility with mods like Additional Placements that add extensive additional blockstates. This is only in the case of large modpacks exceeding 500+ extra blocks.

### Reproduction Steps

1. install a lot of mods with stairs and slabs, or a few mods with a lot of either. A compressed mod and config package is in linked report
2. install World Edit
3. install Additional Placements 2.0.5 -> 2.0.8 (unknown if 2.1.0 causes this yet, probably)
4. attempt to create or load a pre-existing world with all of these mods
5. fail
6. allocate 10-16GB RAM to Minecraft instance
7. try again until the world loads

### Anything Else?

There are some interesting details about how this issue began, but I don't think they're entirely relevant to World Edit's RAM usage, only relevant to Additional Placements' functionality.

wizjany:
this is not a bug nor a leak. it's also a duplicate. even if we were to optimize memory usage for mods that use 10x as many blockstates for a single block as minecraft uses for all blocks combined, we don't support or backport to old versions either.

Issue #2488, opened 2024-03-15
World Edit causes MASSIVE memory leak in server. 


dooblykid:
### WorldEdit Version

7.2.15

### Platform Version

Forge 47.2.20

### Confirmations

- [X] I am using the most recent Minecraft release.
- [X] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [ ] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

World edit seems to inflate the heap for servers to sizes that pretty much cause the server to lag out. It took me forever to find out what mod was causing the tps to drop because I thought it was a mod that adds content and features. But somehow it's being caused by world edit. I extracted the Heap using spark and was able to get this information. 
World edit is currently causing a memory leak that occupies 1.1 GB worth of Memory. Which makes my server unplayable when world edit is installed.

I have provided the full heap file so someone with more knowledge can get to the bottom of this. 

[heap-2024-03-15_05.04.56_Leak_Suspects.zip](https://github.com/EngineHub/WorldEdit/files/14616686/heap-2024-03-15_05.04.56_Leak_Suspects.zip)

`469,374 instances of “com.sk89q.worldedit.world.block.BlockState”, loaded by “cpw.mods.modlauncher.TransformingClassLoader @ 0x686293dc8” occupy 1,204,680,040 (40.96%) bytes. 

Most of these instances are referenced from one instance of “java.lang.Object[]”, loaded by “<system class loader>”, which occupies 4,194,328 (0.14%) bytes. The instance is referenced by “it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap @ 0x6d100fa38”, loaded by “cpw.mods.cl.ModuleClassLoader @ 0x68319eaa8”. 

Thread “java.lang.Thread @ 0x6c7664080 FileSystemWatchService” has a local variable or reference to “cpw.mods.modlauncher.TransformingClassLoader @ 0x686293dc8” which is on the shortest path to “java.lang.Object[1048577] @ 0x713000000”. The thread java.lang.Thread @ 0x6c7664080 FileSystemWatchService keeps local variables with total size 528 (0.00%) bytes.

[...]

PR #2883, opened 2026-01-04
Re-implement backing state maps

octylFractal:
This uses a single array combined with efficiently computed indexes to avoid needing to do entire map equality comparisons or hold large hash-based tables, improving memory and CPU usage.

```
# master with obj2obj maps to fix a bug
Benchmark 1: java -jar fabric-server-mc.1.21.11-loader.0.18.4-launcher.1.1.1.jar --nogui
  Time (mean ± σ):      6.781 s ±  0.128 s    [User: 33.699 s, System: 1.062 s]
  Range (min … max):    6.567 s …  7.017 s    10 runs
# my changes
Benchmark 1: java -jar fabric-server-mc.1.21.11-loader.0.18.4-launcher.1.1.1.jar --nogui
  Time (mean ± σ):      6.464 s ±  0.045 s    [User: 32.450 s, System: 1.104 s]
  Range (min … max):    6.399 s …  6.529 s    10 runs
```
master with obj2obj maps:
heap dump with BlockType retaining 2,097,040 bytes

my changes:
heap dump with BlockType retaining 1,217,728 bytes

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/DefaultBlockTypeStateList.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/DefaultBlockTypeStateList.java
@@ -28,6 +28,8 @@ import it.unimi.dsi.fastutil.objects.Object2ObjectMaps;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 // Suppress Immutable: Properly annotating BlockState with Immutable is for the future
 @SuppressWarnings("Immutable")
@@ -151,11 +153,17 @@ final class DefaultBlockTypeStateList extends BlockTypeStateList {
 
     @Override
     public int calculateIndex(Map<Property<?>, ?> state) {
+        if (state.size() != propertyEntries.size()) {
+            throw new IllegalArgumentException(getDetailedPropertyMismatchException(state));
+        }
         int index = 0;
         for (PropertyEntry entry : propertyEntries) {
             Object value = state.get(entry.property);
             if (value == null) {
-                throw new IllegalArgumentException("Missing or null value for property " + entry.property.name());
+                if (!state.containsKey(entry.property)) {
+                    throw new IllegalArgumentException(getDetailedPropertyMismatchException(state));
+                }
+                throw new IllegalArgumentException("Null value for property " + entry.property.name());
             }
             int offset = entry.getOffsetForValueOrInvalid(value);
             if (offset == -1) {
@@ -164,6 +172,25 @@ final class DefaultBlockTypeStateList extends BlockTypeStateList {
             index += offset;
         }
         return index;
+    }
+
+    private String getDetailedPropertyMismatchException(Map<Property<?>, ?> state) {
+        // Spend some time computing the incorrect properties for better error messages
+        Set<Property<?>> missingProperties = propertyEntries.stream()
+            .map(e -> e.property)
+            .filter(p -> !state.containsKey(p))
+            .collect(Collectors.toSet());
+        Set<Property<?>> extraProperties = state.keySet().stream()
+            .filter(p -> propertyEntries.stream().noneMatch(e -> e.property == p))
+            .collect(Collectors.toSet());
+        StringBuilder errorMessage = new StringBuilder("State has incorrect number of properties.");
+        if (!missingProperties.isEmpty()) {
+            errorMessage.append(" Missing properties: ").append(missingProperties).append(".");
+        }
+        if (!extraProperties.isEmpty()) {
+            errorMessage.append(" Extra properties: ").append(extraProperties).append(".");
+        }
+        return errorMessage.toString();
     }
 
     @Override
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/DefaultBlockTypeStateList.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/DefaultBlockTypeStateList.java
@@ -1,0 +1,186 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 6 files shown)


me4502 (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/DefaultBlockTypeStateList.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Override
+    public BlockState get(int index) {
+        return states.get(index);
+    }
+
+    @Override
+    public int calculateIndex(Map<Property<?>, ?> state) {
+        int index = 0;
+        for (PropertyEntry entry : propertyEntries) {
```
SirYwell: This now silently ignores properties that aren't part of block type, right? `SingletonBlockTypeStateList` throws an exception instead.

octylFractal: Thanks, fixed.

me4502 (approved):

me4502 (approved):

atyrode:
Real-world validation of this change at a scale well beyond the benchmark above, plus a question about reachability on older Minecraft lines.

We hit the pre-#2883 behaviour in production on a 292-mod NeoForge 1.21.1 server (WorldEdit 7.3.8, NeoForge 21.1.235, Java 21.0.12, generational ZGC, `-Xmx6G`). The pack registers **820,533 blockstates**. The server ran at 100% heap, reclaimed nothing, and eventually died on NeoForge's `ServerHangWatchdog`.

### Measured

`jcmd <pid> GC.class_histogram`, top entries by bytes:

| bytes | instances | class |
| --- | --- | --- |
| 835,499,584 | 26,109,362 | `com.google.common.collect.ImmutableMapEntry` |
| 787,676,400 | 9,845,955 | `com.google.common.collect.SingletonImmutableBiMap` |
| 452,607,656 | 6,819,348 | `[Lcom.google.common.collect.ImmutableMapEntry;` |
| 436,105,472 | 6,814,148 | `com.google.common.collect.RegularImmutableMap` |
| 394,643,280 | 9,866,082 | `ImmutableMapEntry$NonTerminalImmutableMapEntry` |
| 393,023,456 | 6,809,285 | `[Ljava.util.Map$Entry;` |
| 50,881,856 | 795,029 | `com.google.common.collect.SparseImmutableTable` |
| 76,825,320 | 873,015 | `java.util.LinkedHashMap` |

That is roughly **3.4 GB** of guava collections. Supporting counts line up with `BlockState` in 7.3.8 exactly: `com.sk89q.worldedit.world.block.BlockState` 820,533 instances (52,514,112 bytes), `BaseBlock` 820,533, `LazyReference` 1,031,890 of which 994,515 were already initialised, and the 873,015 `LinkedHashMap` matches one `values` map per state.

Before and after removing the mod, same server, same heap window, nothing else changed:

| | with WorldEdit 7.3.8 | without |
| --- | --- | --- |
| GC result | `6144M(100%)->6144M(100%)` | `3496M(57%)->2074M(34%)` |
| live-set floor | 6144M, unreclaimable | ~2.07 GB |
| ZGC allocation stalls | 10,027 in 393 s | 4 in 310 s idle |
| TPS (overall) | 1.192 (838.68 ms/tick) | 20.000 (9.82 ms/tick) |

A second server on the same host, running the **same pack on the same `-Xmx6G`** but without WorldEdit, was healthy throughout (`5800M(94%)->2450M(40%)`, 359 stalls in 4.3 h). That was the control that identified WorldEdit rather than the pack or the heap size.

### Why this is the derived index, not the block registry

Worth separating, because earlier reports of this facet (#2488, #2658) were answered with the observation that WorldEdit must inherently store a list of every block, which is correct but accounts for a small fraction of the cost. In 7.3.8 the `BlockState` objects themselves are only ~52 MB across 820,533 states. The ~65x multiplier on top is the per-state neighbouring-state table:

```java
// Neighbouring state table.
private Table<Property<?>, Object, BlockState> states;
```

built eagerly in `populate()`, which for each state walks every property and every value of that property and stores the resulting state in an `ImmutableTable`. Its size is therefore `sum over states of sum over properties of (values(property) - 1)`, which is what produces 26.1 M map entries at 820 k states. That is a *derived lookup index* over the registry, not the registry — which is exactly what this PR replaces with a single array plus computed indexes, and why the approach here is the right one rather than a micro-optimisation.

For anyone finding this via a heap dump: FerriteCore does not help. It deduplicates vanilla's blockstate property maps, not WorldEdit's own copies, and it was active on our server throughout.

Your own numbers show `BlockType` retention dropping 2,097,040 → 1,217,728 bytes (~42%) on a vanilla-scale server. On a pack at 820 k states the same structural change removes gigabytes rather than kilobytes, since the cost scales with the product of states and property values. So this change is worth considerably more on heavily modded servers than the benchmark suggests.

### The question

Is a backport of this to a Minecraft 1.21.1-compatible build considered feasible?

The reason for asking is that the fix is currently unreachable for anyone on that version. Following the tags: 7.3.8 is the last release targeting `neoforge-minecraft = "1.21"`, 7.3.9 moved to 1.21.3, 7.3.19 targets 1.21.11, and 7.4.5 targets 26.2. This PR merged 2026-01-05, well after the 1.21.1 line ended, and I confirmed by reading `BlockState.java` at the tags that 7.3.19 still carries the old `Table<Property<?>, Object, BlockState> states` while 7.4.5 has the new `stateListIndex` / `getInternalStateList()` implementation. So on MC 1.21.1 the only available WorldEdit is 7.3.8, and the structure above is unavoidable there.

I appreciate that maintaining old Minecraft versions is a cost you have explicitly declined before, and I am not asking you to reopen that policy — a straight "no, use a current version" is a perfectly reasonable answer and I will take it as settled. We resolved our own incident by removing WorldEdit from that server (the work it had been installed for turned out to be impossible over RCON anyway, since region commands need a player selection context that a console does not have). I am posting the numbers mainly so the scale of what #2883 fixes is on the record, and so the next person who finds 3 GB of `ImmutableMapEntry` in a heap dump can find the explanation and the fix version rather than concluding it is a leak.

octylFractal:
Not sure why you asked when you know the answer.

Issue #2895, opened 2026-01-21
schematics listing or loading always uses the default path instead of the configured path in WorldEdit 7.4


Brokkonaut:
### WorldEdit Version

Bukkit-Official(7.4.1-SNAPSHOT+7382-2ea1222)

### Platform Version

Paper version 1.21.11-DEV-main@81b9122

### Confirmations

- [x] I am using the most recent Minecraft release.
- [x] I am using a version of WorldEdit compatible with my Minecraft version.
- [x] I am using the latest or recommended version of my platform software.
- [x] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [x] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

- I have configured the schematics folder so it uses not the default folder
- but schematics are still only listed or loaded from the default folder
- When saving it uses the configured folder
- It was working in WorldEdit 7.3.x

### Expected Behavior

- Schematics should be listed and loadable from the configured folder

### Reproduction Steps

1. Configure some schematic path in the worldedit config
2. save schematics and try to list/load them
3. They are saved to the correct folder but loaded/listed from the default folder

### Anything Else?

- Maybe the SchematicsManager is just initialized before the config is loaded?

PR #2896, opened 2026-01-21
Ensure schematics manager is initialised on reloads

me4502:
Move the init point of the schematics manager to the reload configuration event, running `uninit` before the actual init to make sure that it's adequately cleaned up. `uninit` is safe to run when not yet initialised, as it checks before cleaning anything up.

This also ensures that fallback schematic systems are initialised late enough to obtain necessary data.

This should fix issues such as https://github.com/EngineHub/WorldEdit/issues/2895 with filesystems that need to fallback, as well the fact that reloading does not apply schematic folder changes (albeit without an issue number as it's not been reported)

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/Capability.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/Capability.java
@@ -53,7 +53,6 @@ public enum Capability {
         @Override
         void initialize(PlatformManager platformManager, Platform platform) {
             WorldEdit.getInstance().getAssetLoaders().init();
-            WorldEdit.getInstance().getSchematicsManager().init();
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/Capability.java
octylFractal, line 60: Should we move this to match where the `init` is happening now? Maybe just using `PlatformUnreadyEvent`?

me4502: I feel this spot does technically fit better- the issue is our capability system isn't tied very well to the concept of reloading.

imo keeping this part here, and maybe improving the entire lifecycle system as part of WE8 makes the most sense

octylFractal: Alright I guess

octylFractal (approved):

PR #2897, opened 2026-01-22
call ConfigurationLoadEvent after the configuration is loaded on start

### LINKED ISSUE #2895: schematics listing or loading always uses the default path instead of the configured path in WorldEdit 7.4

Brokkonaut:
fixes #2895

the cause of the problem is that ConfigurationLoadEvent is called before the config is actually loaded, i have added debug logging to visualize this: https://cpaste.de/figewenuhe.sql - the paths from the config were never set in the SchematicsManager (and also in the SessionManager).

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
@@ -33,6 +33,7 @@ import com.sk89q.worldedit.bukkit.adapter.BukkitImplAdapter;
 import com.sk89q.worldedit.bukkit.adapter.BukkitImplLoader;
 import com.sk89q.worldedit.event.platform.CommandEvent;
 import com.sk89q.worldedit.event.platform.CommandSuggestionEvent;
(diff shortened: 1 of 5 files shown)


me4502 (approved):
I'm happy with this as-is. Realistically we need to redo the startup lifecycle in WE8 as it doesn't currently match MC _super well_, but doing it this way at least is tying configuration load directly to the config.load call, which is better than what we had before.

PR #2898, opened 2026-01-22
add a flag to filter the schematics list

Brokkonaut:
just an optional filter when listing schematics


me4502:
thanks for the PR. as this is a feature addition, could you please target this to the `master` branch?

me4502:
also just for consistency, it might make sense to use the same filtering that the //registry command uses- it supports globbing/etc

octylFractal (changes requested):
Flagging for changes per above comments.

PR #2900, opened 2026-01-26
Update entity leash position and home position in new minecraft versions when pasting

Brokkonaut:
leash position updating was broken for some versions now and the home postion is used by some mobs like happy ghasts. I changed the code to not return after the first change so multiple changes can be applied to an entity if needed

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
@@ -177,14 +177,39 @@ public class ExtentEntityCopy implements EntityFunction {
                         tagX.value().intValue(), tagY.value().intValue(), tagZ.value().intValue()
                     );
                     BlockVector3 newLeash = transform.apply(tilePosition.subtract(from)).add(to).toBlockPoint();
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/entity/ExtentEntityCopy.java
```diff
@@ -177,15 +177,40 @@ private BaseEntity transformNbtData(BaseEntity state) {
...
                         .put("Leash", leashCompound.toBuilder()
                             .putInt("X", newLeash.x())
                             .putInt("Y", newLeash.y())
                             .putInt("Z", newLeash.z())
                             .build()
                         ).build()));
+                    tag = state.getNbt();
                 }
             }
+            LinIntArrayTag leashArray = tag.findTag("leash", LinTagType.intArrayTag());
```
me4502, line 190: I forget if LinBus is case sensitive, but am I correct in that MC changes from capital `Leash` compound tag, to lowercase `leash` int array

Brokkonaut: Yes its case sensitive and its an int array in newer versions if its bound to a fence and a compound when bound to an entity and always lowercase now

me4502 (approved):

PR #2905, opened 2026-02-16
Add very basic Folia support

### LINKED ISSUE #2348: Folia support
### The Problem

Now WorldEdit doesn't support Folia and adding `folia-supported: true` doesn't fix it because of using BukkitScheduler.

### A Solution

Add Folia support the way you like

### Alternatives

_No response_

### Anything Else?

_No response_

me4502:
This PR adds some _very_ rudimentary Folia support to WorldEdit.

The goal here is not to make the plugin perfectly usable on Folia, but moreso to get it running with anything unsafe disabled.

There are a very large number of plugins that rely on our API, and the fact that we cannot even run on Folia prevents them from adding support of their own.  This PR allows us to at least run on Folia, with some basic functionality.

If something is disabled/doesn't work in this PR, that's acceptable. If something _throws errors_ or otherwise causes issues, then it's something we want to fix.

This is just a base, and would allow us to look at better-integrating individual systems in the future.

Current caveats:

* Edits outside of the primary region are disabled. In the future we might be able to move this to "region of pos1" or similar.
* Entity-related actions are disabled
* Regen is disabled

Closes https://github.com/EngineHub/WorldEdit/issues/2348

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
@@ -133,15 +133,19 @@ public class BukkitBlockCommandSender extends AbstractCommandBlockActor {
             private volatile boolean active = true;
 
             private void updateActive() {
-                Block block = sender.getBlock();
-                if (!block.getWorld().isChunkLoaded(block.getX() >> 4, block.getZ() >> 4)) {
-                    active = false;
-                    return;
+                try {
+                    Block block = sender.getBlock();
+                    if (!block.getWorld().isChunkLoaded(block.getX() >> 4, block.getZ() >> 4)) {
+                        active = false;
+                        return;
+                    }
+                    Material type = block.getType();
+                    active = type == Material.COMMAND_BLOCK
+                        || type == Material.CHAIN_COMMAND_BLOCK
+                        || type == Material.REPEATING_COMMAND_BLOCK;
+                } catch (Throwable t) {
+                    WorldEdit.logger.warn("Exception while updating command block sender active state", t);
                 }
-                Material type = block.getType();
-                active = type == Material.COMMAND_BLOCK
-                    || type == Material.CHAIN_COMMAND_BLOCK
-                    || type == Material.REPEATING_COMMAND_BLOCK;
             }
 
             @Override
@@ -158,13 +162,7 @@ public class BukkitBlockCommandSender extends AbstractCommandBlockActor {
                         updateActive();
                     } else {
                         // We need to delegate to the right thread.
-                        Bukkit.getRegionScheduler().execute(plugin, sender.getBlock().getLocation(), () -> {
-                            try {
-                                updateActive();
-                            } catch (Throwable t) {
-                                WorldEdit.logger.warn("Exception while updating command block sender active state", t);
-                            }
-                        });
+                        Bukkit.getRegionScheduler().execute(plugin, sender.getBlock().getLocation(), this::updateActive);
                     }
 
                     return active;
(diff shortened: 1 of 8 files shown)


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
```diff
@@ -151,6 +151,25 @@ public String getName() {
...
+                        // This thread owns the block, so we can immediately update.
+                        updateActive();
+                    } else {
+                        // We need to delegate to the right thread.
+                        Bukkit.getRegionScheduler().execute(plugin, sender.getBlock().getLocation(), () -> {
+                            try {
+                                updateActive();
+                            } catch (Throwable t) {
+                                WorldEdit.logger.warn("Exception while updating command block sender active state", t);
+                            }
```
octylFractal: Can we pull this whole lambda out to a method so we can share it with non-folia?

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitServerInterface.java
```diff
@@ -121,7 +121,14 @@ public void reload() {
 
     @Override
     public int schedule(long delay, long period, Runnable task) {
-        return Bukkit.getScheduler().scheduleSyncRepeatingTask(plugin, task, delay, period);
+        if (plugin.isFolia()) {
+            Bukkit.getGlobalRegionScheduler().runAtFixedRate(plugin, scheduledTask -> task.run(), delay, period);
+            // TODO Paper doesn't appear to have a concept of task IDs, so return -1 here for now.
+            //  We may want to store these tasks and map them to our own IDs to cancel them later.
+            return -1;
```
octylFractal: This is bad, `-1` is used as a sentinel value for `queueBlockBreakEffect` and it'll just spawn a billion tasks if we do this.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
```diff
@@ -507,6 +523,26 @@ BukkitImplAdapter getBukkitImplAdapter() {
...
+        try {
+            // Folia is Paper-based, so this is a good first check.
+            if (PaperLib.isPaper()) {
+                // Then we can check against the `papermc:folia` key, as per the `isBrandCompatible` javadoc.
+                // This API is experimental so might randomly break on us (hence the try/catch)
+                // TODO if we drop Spigot support in the future, remove this check and purely use Folia-compatible APIs.
+                return ServerBuildInfo.buildInfo().isBrandCompatible(Key.key("papermc", "folia"));
+            }
+        } catch (Throwable t) {
+            // Ignore, this likely means an outdated version.
```
octylFractal, line 536: Can we at least log here?

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
me4502: one of these takes a callable that must return null, the other is just a runnable - so either way there'll be some difference here. the only "shared" part would be the try/catch. i can extract it anyway though

or i guess i can move the try/catch to updateActive itself

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitServerInterface.java
me4502: I'll just return `1` as that's at least a "valid" ID, we just can't cancel it.

octylFractal (approved):

PR #2912, opened 2026-02-28
Remove deprecated tool aliases

me4502:
This PR removes the deprecated tool aliases (`/tool info` -> `/info`)

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolCommands.java
@@ -19,7 +19,6 @@
 
 package com.sk89q.worldedit.command;
 
-import com.google.common.collect.Collections2;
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolCommands.java
```diff
@@ -183,7 +141,6 @@ public void none(Player player, LocalSession session) throws WorldEditException
 
     @Command(
         name = "selwand",
-        aliases = "/selwand",
```
me4502, line 186: Open question of whether we want to keep these three global aliases. I assume not to try making navwand/selwand less "special"

octylFractal: I don't see a reason to.

octylFractal (approved):

PR #2919, opened 2026-03-03
Fixed certain commands that teleport the player or an entity to Folia.

Euphillya:
The Entity#teleport method is not supported on Folia, unlike Entity#teleportAsync. Furthermore, since PaperLib is integrated with WorldEdit, it can be used if a Folia server is detected.


me4502 (approved):
Thanks for this. While I'm not super happy about just returning true on Folia for teleports, I don't feel there's a way around this with the current API. As this only impacts Folia I'm happy with it as-is (with the [intention of replacing it with something better in the future](https://github.com/EngineHub/WorldEdit/issues/2921), which is honestly the status of most Folia support anyway tbh)

PR #2923, opened 2026-03-08
Update Gradle and some plugins

### LINKED ISSUE #2920: Update Gradle Shadow plugin to v9
The Gradle shadow plugin needs to be updated to v9, but it appears to be a semi complex migration.

octylFractal:
Fixes #2920


octylFractal:
Verified with https://github.com/JakeWharton/diffuse and pulling out some of the changed `.class` and using `vbindiff` that the only changes are correctness changes in signatures. No more or less classes/methods are present in all distribution JARs.

me4502 (approved):

PR #2936, opened 2026-04-06
fix ExtendingCuboidRegionSelector constructor

SamuraiWorf:
Not sure if this caused any issues anywhere, but I found this after making changes in a forked WorldEdit.


octylFractal (approved):

octylFractal:
Thank you, that does look like it would make the select calls not work properly.

PR #2938, opened 2026-04-06
convert authorization exceptions

SamuraiWorf:
Pretty niche issue, but caused some more serious exceptions.

-> 1. Only have `worldedit.undo.self` permission. 2. Run `//undo anything`

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
@@ -114,7 +114,7 @@ public class BukkitBlockCommandSender extends AbstractCommandBlockActor {
     @Override
     public void checkPermission(String permission) throws AuthorizationException {
         if (!hasPermission(permission)) {
-            throw new AuthorizationException(TranslatableComponent.of("worldedit.command.permissions"));
+            throw new AuthorizationException();
         }
     }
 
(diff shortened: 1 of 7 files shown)


octylFractal:
What happens currently? The existing `WorldEditException` handler should already be mapping this properly.

Maybe there's some `null` issues? I think we should just pass the proper `Component` in at call sites for now, and maybe deprecate having `null` for `Component` in `WorldEditException` in a separate PR.

SamuraiWorf:
Yeah exactly, its the null issues. Switched to fixing it at the call sites.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockCommandSender.java
```diff
@@ -113,7 +114,7 @@ public String[] getGroups() {
     @Override
     public void checkPermission(String permission) throws AuthorizationException {
         if (!hasPermission(permission)) {
-            throw new AuthorizationException();
+            throw new AuthorizationException(TranslatableComponent.of("worldedit.command.permissions"));
```
me4502: It's worth noting that we _do_ have exceptions that assign the component inside the constructor when none is given. As it's being used with the same one each time here, it probably makes sense to do that instead (to prevent further issues in the future)

Currently the no-arg constructor for this class just does nothing, but it could instead call super with this component

SamuraiWorf: I'm fine with either. Done as suggested ^^

me4502 (approved):

PR #2940, opened 2026-04-06
create parity between //snow and //thaw

SamuraiWorf:
makes //thaw and //snow cover the same area


octylFractal (changes requested):
Is it possible to also make `thaw` use block positions from a `CylinderRegion` instead? I'm unsure if it works properly, but I would rather use the same source of truth if we can.

SamuraiWorf:
Done as suggested

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -2075,49 +2076,24 @@ public final int thaw(BlockVector3 position, double radius)
...
-                            ++affected;
-                        }
-                    } else if (id == BlockTypes.SNOW) {
-                        if (setBlock(pt, air)) {
-                            ++affected;
-                        }
-                    } else if (id.getMaterial().isAir()) {
-                        continue;
-                    }
+        return thaw(new CylinderRegion(position, Vector2.at(radius, radius), position.y(), height));
```
octylFractal: Can you change this and `simulateSnow` to use `position.y() - height, position.y() + height`, so they match the commands? Currently this doesn't do as the Javadoc describes.

octylFractal (changes requested):
BTW, if you're able to, re-requesting a review helps me sort PRs better.

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
SamuraiWorf: Done. I wasn't sure if a re-request would be perceived more as an annoyance, so I didn't do it. Will do, once the actions ran through ^^

octylFractal (approved):

octylFractal:
Since this is feature work, I'm going to put it in `master`.

PR #2941, opened 2026-04-07
Improve schematic filename exception handling

me4502:
Currently if the watcher encounters an illegal file it kills the watcher. Instead it should warn for this specific file, but continue along so that one single bad file doesn't prevent others being seen.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
@@ -122,8 +122,13 @@ public class RecursiveDirectoryWatcher implements Closeable {
                 if (Files.isDirectory(path)) {
                     triggerInitialEvents(path);
                 } else {
-                    path = WorldEdit.getInstance().getSafeOpenFile(null, schematicRoot.toFile(), schematicRoot.relativize(path).toString(), null).toPath();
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/internal/util/RecursiveDirectoryWatcher.java
```diff
@@ -182,17 +187,22 @@ public void start(Consumer<DirEntryChangeEvent> eventConsumer) {
...
+                                        triggerInitialEvents(path);
+                                    } catch (IOException | FilenameException e) {
+                                        LOGGER.error(e);
+                                    }
+                                } else { // new file created
+                                    eventConsumer.accept(new FileCreatedEvent(path));
                                 }
-                            } else { // new file created
-                                eventConsumer.accept(new FileCreatedEvent(path));
+                            } catch (FilenameException e) {
```
octylFractal, line 192: Shouldn't this just wrap the `path = ...` part?

me4502: do we want it to allow loading "unsafe paths"? i did it this way so not allowed paths are not added to the manager

octylFractal: Can you not use `continue;` in the catch block so it skips it?

octylFractal (approved):

PR #2942, opened 2026-04-07
expand disallowed blocks to categories and copying

SamuraiWorf:
Hiya! For a bit of context: I'm from zero.minr.org and we've been working on a WE fork the past few months to make it safer in order to open it up to a wide user base.
We felt it right to open up PRs on the non server-specific changes we made, in case you find any of them worth to add into WorldEdit.

This first one is meant to make the disallowed-blocks list work with block categories and //copy, //move, etc.
I think there are a couple things that are up for debate, f.e. where to put the config options, or whether calling WorldEdit.getInstance() is a terrible idea, but let me know what you think ^^

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
@@ -119,6 +119,7 @@ import com.sk89q.worldedit.util.collection.DoubleArrayList;
 import com.sk89q.worldedit.util.eventbus.EventBus;
 import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import com.sk89q.worldedit.util.formatting.text.TranslatableComponent;
+import com.sk89q.worldedit.util.mask.BlockMaskUtil;
 import com.sk89q.worldedit.world.NullWorld;
 import com.sk89q.worldedit.world.World;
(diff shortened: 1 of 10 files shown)


SamuraiWorf:
No idea what to do about the api compatability. The reports seem to suggest that this has nothing to do with my PR.

wizjany:
I don't think this makes sense to apply to source masks given the intent of the feature. It's a safety feature, not a security feature. That is, it doesnt make sense to prevent someone from creating a schematic (if they have permission to) with those blocks - they might just want to share the schematic and download it for use elsewhere. It's really just meant to prevent accidents like `//set grass` instead of `//set grass_block` (though the fact that people generally use op instead of assigning proper perms often makes this useless anyway...). I could see this logic applied to tags as well, but it would be in the form of a list of tags to disallow, not checking tags by the blocks they contain. (Also yea the implementation is really wasteful, grabbing config for every block is probably a lot of overhead)

SamuraiWorf:
Good call on the excessive worldEdit.getConfiguration() calls - fixed.

For us the feature certainly has been co-opted to also function as a security feature. But I get it if that's not the direction WorldEdit wants to take it.

As it stands, this is just a suggestion, I can remove the mask part, or we close the PR altogether, that's up to yall.

octylFractal:
I think we don't see much reason to expand this feature for the use cases we target.

PR #2943, opened 2026-04-08
Re-add random tree generator

me4502:
This PR re-adds the old "random" tree generator alias, via a custom `worldedit:random` tree type.

I've opted to keep it under the WorldEdit namespace to prevent any collisions with MC

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/generation/WorldEditTreeGeneration.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/generation/WorldEditTreeGeneration.java
@@ -1,0 +1,54 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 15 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/generation/WorldEditTreeGeneration.java
```diff
@@ -0,0 +1,54 @@
...
+    private WorldEditTreeGeneration() {
+    }
+
+    @SuppressWarnings("deprecation")
+    @Nullable
+    public static Boolean handleWorldEditTrees(World world, TreeType type, EditSession editSession, BlockVector3 position)
+            throws MaxChangedBlocksException {
+        if (type == WorldEditTreeTypes.RANDOM) {
+            var treeTypes = List.copyOf(TreeType.REGISTRY.values());
+            TreeType randomType = treeTypes.get(ThreadLocalRandom.current().nextInt(treeTypes.size()));
```
octylFractal: Instead of doing a full copy here, what about:
```suggestion
            Collection<TreeType> treeTypes = TreeType.REGISTRY.values();
            TreeType randomType = Iterables.get(treeTypes, ThreadLocalRandom.current().nextInt(treeTypes.size()));
```
This will only iterate as far as it needs to and won't need to allocate.

octylFractal (approved):
Minor optimization but otherwise LGTM

PR #2948, opened 2026-04-10
Cleanup non-platform registry init

me4502:
This PR cleans up registry init of non-platform-specific registries.

Also fixed that TREE_TYPE and REGISTRY were missing from our registry registry

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/registry/CommonRegistries.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/registry/CommonRegistries.java
@@ -1,0 +1,42 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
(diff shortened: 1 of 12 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/registry/CommonRegistries.java
```diff
@@ -0,0 +1,42 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.registry;
+
+import com.sk89q.worldedit.world.gamemode.GameModes;
+import com.sk89q.worldedit.world.generation.WorldEditTreeTypes;
+import com.sk89q.worldedit.world.weather.WeatherTypes;
+
+public class CommonRegistries {
```
octylFractal: `final`?

octylFractal (approved):

PR #2951, opened 2026-04-13
One Native Code to Rule Them All

octylFractal:
One native code to rule them all,
   one native code to find them,
One native code to bring them all
   and in the darkness bind them.

---

Merges most of Fabric and NeoForge into core-mc. This new project is just Fabric without the Fabric Loader or API, giving us a clean Minecraft dependency to build off of.

In the future, parts of this may be applicable to Paper as well. However, due to the nature of their patches, it's not as easy to assume that it builds cleanly. It may be necessary to instead copy sources to Paper to be compiled against it directly, to ensure there's no missing signatures or other incompatibilities.

See `worldedit-core-mc/README.md` for a better way to inspect the diff here.


worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/internal/CoreMcConfiguration.java
me4502, line 40: If we're making this shared, is there a point in not just having these in the base properties class?

Realistically the `cheat-mode` option here is more tied to "Integrated vs Dedicated server" rather than platform too, so I wonder if it makes sense to have some separate shared config for those two

worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/internal/CoreMcMod.java
```diff
@@ -0,0 +1,354 @@
...
+        this.config = new CoreMcConfiguration(workingDir);
+
+        platform.setPermissionsProvider(createPermissionsProvider(platform));
+    }
+
+    protected CoreMcConfiguration getConfig() {
+        return this.config;
+    }
+
+    protected void setupRegistries(MinecraftServer server) {
```
me4502, line 120: I am semi concerned about just how tightly we're tying it to how NF/Fabric _currently_ work. Eg, when it comes to registries we've in the past expressed interest in moving these to platform registry events etc.

I'm just a little concerned we're going to end up in spots where it's significant work to override something on a per-platform basis because of how closely we're tying it all to current behaviour

worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/internal/CoreMcWorld.java
```diff
@@ -207,20 +199,19 @@ public <B extends BlockStateHolder<B>> boolean setBlock(BlockVector3 position, B
     @Override
     public Set<SideEffect> applySideEffects(BlockVector3 position, BlockState previousType, SideEffectSet sideEffectSet) {
         worldNativeAccess.applySideEffects(position, previousType, sideEffectSet);
-        return Sets.intersection(FabricWorldEdit.inst.getPlatform().getSupportedSideEffects(), sideEffectSet.getSideEffectsToApply());
+        return Sets.intersection(CoreMcPlatform.SUPPORTED_SIDE_EFFECTS, sideEffectSet.getSideEffectsToApply());
```
me4502, line 202: Is this necessarily safe in multi-platform scenarios?

worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/CoreMcAdapter.java
```diff
@@ -41,69 +47,57 @@
...
-import java.util.TreeMap;
 import javax.annotation.Nullable;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-public final class NeoForgeAdapter {
+/**
+ * Common adapter methods for platforms sharing native Minecraft code.
+ */
+public final class CoreMcAdapter {
```
me4502, line 73: Can we please have this extended on both the Fabric & NF end to reduce confusion by API users? IMO CoreMC is an implementation detail, and users should still expect a "FabricAdapter" or "NeoForgeAdapter" class to exist, even if they delegate to CoreMcAdapter

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
```diff
@@ -78,7 +80,11 @@ public class PlatformManager {
     private final PlatformCommandManager platformCommandManager;
     private final SimpleLifecycled<ListeningExecutorService> executorService;
     private final Map<Platform, Boolean> platforms = Maps.newHashMap();
-    private final Map<Capability, Platform> preferences = new EnumMap<>(Capability.class);
+    private final ImmutableMap<Capability, SimpleLifecycled<Platform>> preferences = Stream.of(Capability.values())
```
me4502, line 83: Have these lifecycling changes been tested on other platforms? (Bukkit, etc)

worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/internal/CoreMcWorld.java
octylFractal: If this is running, we must be part of a `WORLD_EDITING` platform, so I think it makes sense to use these effects.

worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/CoreMcAdapter.java
octylFractal: Sure.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformManager.java
octylFractal: No, but I can't imagine they would affect anything, it's literally the equivalent of the old map just a layer deeper.

worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/internal/CoreMcConfiguration.java
octylFractal: We can't, that is shared by `CLIConfiguration`.

worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/internal/CoreMcMod.java
octylFractal: This is all internal so we can refactor it however we want later. There is nothing preventing us from changing all of this code.

worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/internal/CoreMcWorld.java
octylFractal: I changed it with the new commit anyways.

me4502 (approved):

PR #2952, opened 2026-04-19
FullCubeMask improvements

TomyLobo:
- Make #fullcube match double slabs
- Add #visualfullcube

**Note:** This is based on #2997, please don't squash.

### DIFF
--- a/worldedit-bukkit/adapters/adapter-1.21.11/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_21_11/PaperweightAdapter.java
+++ b/worldedit-bukkit/adapters/adapter-1.21.11/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_21_11/PaperweightAdapter.java
@@ -586,6 +586,11 @@ public final class PaperweightAdapter implements BukkitImplAdapter {
         return new PaperweightBlockMaterial(mcBlockState);
     }
 
+    @Override
+    public BlockMaterial getBlockMaterial(BlockState blockState) {
+        return new PaperweightBlockMaterial(adapt(blockState));
+    }
+
     @SuppressWarnings({ "unchecked", "rawtypes" })
     private static final LoadingCache<net.minecraft.world.level.block.state.properties.Property, Property<?>> PROPERTY_CACHE = CacheBuilder.newBuilder().build(new CacheLoader<>() {
         @Override
--- a/worldedit-bukkit/adapters/adapter-1.21.11/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_21_11/PaperweightAdapter.java
+++ b/worldedit-bukkit/adapters/adapter-1.21.11/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_21_11/PaperweightAdapter.java
@@ -606,6 +606,11 @@ public final class PaperweightAdapter implements BukkitImplAdapter {
         return new PaperweightBlockMaterial(mcBlockState);
     }
 
+    @Override
+    public BlockMaterial getBlockMaterial(BlockState blockState) {
+        return new PaperweightBlockMaterial(adapt(blockState));
+    }
+
     @SuppressWarnings({ "unchecked", "rawtypes" })
     private static final LoadingCache<net.minecraft.world.level.block.state.properties.Property, Property<?>> PROPERTY_CACHE = CacheBuilder.newBuilder().build(new CacheLoader<>() {
         @Override
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockRegistry.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockRegistry.java
@@ -35,7 +35,7 @@ import java.util.OptionalInt;
 import javax.annotation.Nullable;
 
 public class BukkitBlockRegistry implements BlockRegistry {
-    private final Map<Material, BukkitBlockMaterial> materialMap = new HashMap<>();
+    private final Map<BlockState, BukkitBlockMaterial> materialMap = new HashMap<>();
 
     @Override
     public Component getRichName(BlockType blockType) {
@@ -47,15 +47,16 @@ public class BukkitBlockRegistry implements BlockRegistry {
 
     @Nullable
     @Override
-    public BlockMaterial getMaterial(BlockType blockType) {
-        Material mat = BukkitAdapter.adapt(blockType);
-        if (mat == null) {
-            return null;
-        }
-        return materialMap.computeIfAbsent(mat, material -> {
+    public BlockMaterial getMaterial(BlockState blockState) {
+        return materialMap.computeIfAbsent(blockState, _ -> {
+            Material material = BukkitAdapter.adapt(blockState.getBlockType());
+            if (material == null) {
+                // return null means create no mapping
+                return null;
+            }
             BlockMaterial platformMaterial = null;
             if (WorldEditPlugin.getInstance().getBukkitImplAdapter() != null) {
-                platformMaterial = WorldEditPlugin.getInstance().getBukkitImplAdapter().getBlockMaterial(blockType);
+                platformMaterial = WorldEditPlugin.getInstance().getBukkitImplAdapter().getBlockMaterial(blockState);
             }
             return new BukkitBlockMaterial(platformMaterial, material);
         });
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockRegistry.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockRegistry.java
@@ -37,7 +37,7 @@ import java.util.OptionalInt;
 import javax.annotation.Nullable;
 
 public class BukkitBlockRegistry implements BlockRegistry {
-    private final Map<Material, BukkitBlockMaterial> materialMap = new HashMap<>();
+    private final Map<BlockState, BukkitBlockMaterial> materialMap = new HashMap<>();
 
     @Override
     public Component getRichName(BlockType blockType) {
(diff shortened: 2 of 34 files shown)


worldedit-bukkit/adapters/adapter-1.21.11/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_21_11/PaperweightAdapter.java
```diff
@@ -174,6 +175,7 @@
 import org.enginehub.linbus.tree.LinStringTag;
 import org.enginehub.linbus.tree.LinTag;
 import org.enginehub.linbus.tree.LinTagType;
+import org.jspecify.annotations.NonNull;
```
octylFractal, line 178: We don't use JSpecify currently, and we assume everything is non-null by default. Can you remove all the `@NonNull` and import/use `javax.annotation.Nullable` where you need that?

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockRegistry.java
```diff
@@ -61,6 +61,21 @@ public BlockMaterial getMaterial(BlockType blockType) {
         });
     }
 
+    @Override
+    public @org.jspecify.annotations.Nullable BlockMaterial getMaterial(BlockState blockState) {
+        Material mat = BukkitAdapter.adapt(blockState.getBlockType());
+        if (mat == null) {
+            return null;
+        }
+        return materialMap.computeIfAbsent(mat, material -> {
```
octylFractal, line 70: IMO we should just remove the `materialMap` entirely, and cache this in the `BlockState` using a `LazyReference` like `BlockType` already does.

Plus, the one in `BlockType` can be removed and replaced with a call to `getDefaultState().getMaterial()`.

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockStateHolder.java
```diff
@@ -41,14 +42,23 @@ public interface BlockStateHolder<B extends BlockStateHolder<B>> extends Pattern
      */
     BlockType getBlockType();
 
+    /**
+     * Get the block material.
+     *
+     * @return The material
+     */
+    default BlockMaterial getMaterial() {
```
octylFractal, line 50: New API should be targeted to `master`, not `version/7.4.x`. Also, IMO we shouldn't expose this here, it should just be on `BlockState` alone.

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
```diff
@@ -63,6 +63,17 @@ default String getName(BlockType blockType) {
...
     BlockMaterial getMaterial(BlockType blockType);
 
+    /**
+     * Get the material for the given block state.
+     *
+     * @param blockState the block state
+     * @return the material, or null if the material information is not known
+     */
+    @Nullable
+    default BlockMaterial getMaterial(BlockState blockState) {
```
octylFractal, line 73: It's a bit complicated, but I think this would be best as `@NonAbstractForCompatibility`, and `getMaterial(BlockType)` should be deprecated for removal.

octylFractal (changes requested):

worldedit-bukkit/adapters/adapter-1.21.11/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_21_11/PaperweightAdapter.java
TomyLobo: done

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockStateHolder.java
TomyLobo: retargeted to master and removed from BlockStateHolder/BaseBlock

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
TomyLobo: Added @NonAbstractForCompatibility and the apiNote like I saw in other places.
But I don't quite understand why `getMaterial(BlockType)` should be deprecated for removal.
As far as I can tell, this gets the actual nms-level default block state, whereas `getMaterial(blockType.getDefaultState())` merely gets the first in the list of possible states.

I don't know if that's a distinction without a difference, but I like to understand changes before I put my name on them :)

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockRegistry.java
TomyLobo: I'm not so sure about removing `materialMap`.
A `LazyReference` in `BlockState` would only survive as long as the `BlockState` does, and I'm not sure `BlockState`s are cached for as long as that `materialMap` would be.
From what I can tell, `BlockState`s can be created by a `LazyReference` in `BlockType` and `BlockType`'s constructors are public, so anyone could create a new instance at any time, and, along with it, a `BlockState`.

Basically, every `new BlockType(0).getDefaultState().getMaterial()` would create a new `BlockMaterial` instance representing air.
With the current implementation, the `materialMap` deduplicates that and we get the same `BlockMaterial` instance each time.

On the other hand, what do we gain?
- Do we reduce complexity?
  - Not really. We replace one field by another. I'd even argue that `LazyReference` is a more obscure feature than `Map.computeIfAbsent`.
- Do we save memory by discarding unneeded `BlockMaterial` instances?
  - Most of the time, no:
    - If we use a `LazyReference`, `BlockMaterial`s will be referenced by their `BlockState`, so they are tied to that `BlockState`'s lifetime.
    - All `BlockState`s are referenced by their `BlockType`'s internal `BlockTypeStateList`, so they are tied to that `BlockType`'s lifetime.
    - `BlockType`s, in turn, are mostly created on server start and held indefinitely, so nothing is ever discarded, just like with the `materialMap`.
    - And for user-created `BlockType` instances, we lose `BlockMaterial` deduplication, as mentioned before.

As best as I can tell, it's down to taste, so changing it is beyond the scope of this PR.

octylFractal: We don't expect people to be making new `BlockType`s at all, so this is acceptable. We save memory because a `HashMap` uses extra memory than just that required for the items in it, due to the load factor. It's also a lot faster to read two fields (the reference and its stored value) from an object we're already in, rather than do a hash lookup. Also, as soon as we hit a Java version with a stable [Lazy Constants](https://openjdk.org/jeps/526), we're likely going to replace all `LazyReference`s with that, so it won't be so obscure anymore.

I think we can do it outside of this PR though.

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
```diff
@@ -63,6 +64,23 @@ default String getName(BlockType blockType) {
     @Nullable
     BlockMaterial getMaterial(BlockType blockType);
```
octylFractal: Let's deprecate this since as we discussed the default state is the real default state.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
TomyLobo: after you explained on Discord how it actually is the same, I made the adjustments

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
TomyLobo: already done in the wake of the other change :)

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockRegistry.java
```diff
@@ -47,15 +47,15 @@ public Component getRichName(BlockType blockType) {
 
     @Nullable
     @Override
-    public BlockMaterial getMaterial(BlockType blockType) {
-        Material mat = BukkitAdapter.adapt(blockType);
+    public BlockMaterial getMaterial(BlockState blockState) {
+        Material mat = BukkitAdapter.adapt(blockState.getBlockType());
```
me4502: I'm a little unsure we can safely do this? It currently looks like this code just caches the BlockMaterial against a BlockType, using whatever BlockState is first given to it. Meaning we're not actually getting this data on a per-state basis.

I feel it should either be implemented properly to work on a per-state basis, or at the very least always use the default state rather than "whichever is passed first" for better consistency.

me4502 (changes requested):

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitBlockRegistry.java
TomyLobo: Ideally, materials should be cached based on WorldEdit BlockState, or its Bukkit equivalent.
What's the equivalent of WorldEdit's BlockState in Bukkit?
Alternatively, is there a reason not to use WorldEdit's BlockState as a key for materialMap?

TomyLobo: After talking with the #dev-team on discord, I decided to use the BlockState from the parameter as the key.
That seems to be the most efficient solution here.
It works fine on latest paper now.

me4502:
Would you be able to please extract the two shared commits into their own PR, so that they can be reviewed separately from the two PRs that make use of them? I feel like they are in a much more mergeable state, and it would make the other two PRs more reviewable / simpler to understand once those are in (alongside becoming squashable/etc)

TomyLobo:
> Would you be able to please extract the two shared commits into their own PR, so that they can be reviewed separately from the two PRs that make use of them? I feel like they are in a much more mergeable state, and it would make the other two PRs more reviewable / simpler to understand once those are in (alongside becoming squashable/etc)

done

PR #2983, opened 2026-07-04
refactor(regions)!: rename RegionIntersection to RegionUnion

### LINKED ISSUE #2982: Rename RegionIntersection to RegionUnion
And also update its documentation a bit. Because it is a union, not an intersection.

masonakdev:
## Summary

Renames `RegionIntersection` to `RegionUnion` to match its existing behavior.

Also updates related selection errors, API verification entries, and some minor cleanup items in `RegionUnion`.

## Changes

- Rename `RegionIntersection` to `RegionUnion`
- Update union expand/contract keys and messages
- Add the accepted core API break
- Fix minor warnings in `RegionUnion`

## Testing

- `./gradlew :worldedit-core:compileJava`
- `./gradlew :verification:checkCoreApiCompatibility`
  - *Note: `checkCoreApiCompatibility` still fails due to unrelated existing API changes. The `RegionIntersection` removals are accepted by the updated verification file.*

Fixes #2982


worldedit-core/src/main/java/com/sk89q/worldedit/regions/RegionIntersection.java
```diff
@@ -50,39 +51,17 @@ public class RegionIntersection extends AbstractRegion {
...
-     * @param world   the world
-     * @param regions a list of regions, which is copied
-     */
-    public RegionIntersection(World world, List<Region> regions) {
-        super(world);
-        checkNotNull(regions);
-        checkArgument(!regions.isEmpty(), "empty region list is not supported");
-        this.regions.addAll(regions);
-    }
-
```
octylFractal, line 78: Why did you delete these?

worldedit-core/src/main/java/com/sk89q/worldedit/regions/RegionUnion.java
```diff
@@ -133,7 +112,7 @@ public boolean contains(BlockVector3 position) {
     }
 
     @Override
-    public Iterator<BlockVector3> iterator() {
+    public @NonNull Iterator<BlockVector3> iterator() {
```
octylFractal: This shouldn't be annotated, our assumption for unannotated values is that it's non-null.

octylFractal (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/regions/RegionIntersection.java
masonakdev: They weren't used anywhere else in the over-arching WorldEdit project, but I suppose I failed to consider that worldedit-core may be used as a library in other projects that do use them. Want me to add them back?

octylFractal: Yes, we are a library.

masonakdev: Lol, good point. Apologies.

octylFractal (approved):

octylFractal:
Thank you for your contribution!

PR #2984, opened 2026-07-05
Remove redundant adapter methods for get/setBiome

Dreeam-qwq:
Cleanup redundant adapter methods for get/setBiome. These adapter methods were added many years ago, and for now, there is no point to keep use them; we can just use bukkit api instead.


wizjany:
the bukkit api does not (did not?) support setting custom biomes. on those versions, the output of BukkitAdapter.adapt can't be used. i haven't kept up with upstream changes (barely with EH stuff), but given the targeting matrix here (spigot and paper back multiple versions), i wouldn't be so quick to consider this pointless.

me4502:
When I last audited our registry hacks last year, I found that biomes still had issues on Spigot in versions that we still do currently support. I've not had a chance to take a look at versions since then as Spigot for 26.1+ is rather unstable on my machine, but we will need to keep these adapters either until all versions which don't work properly are unsupported, or if we ever chose to drop Spigot.

PR #2988, opened 2026-07-06
Adopt Crankcase for WorldEdit build logic

octylFractal:
yay


verification/src/changes/accepted-core-public-api-changes.json
```diff
@@ -1,27 +1,2 @@
 {
-    "Constructor was not useful": [
```
me4502, line 2: These are new changes since 7.4.x right? Should they actually be removed?

octylFractal: The reason this occurred is that I actually wired in the "stale changes" rejection logic, so these must not apply anymore. Obviously, if they did, it wouldn't build anymore :)

me4502: Only reason I asked was a concern that it wasn't working properly anymore. My understanding is the ones listed actually are valid changes since 7.4.x that should show up

Eg, the snapshot ones are for the bugfix to snapshots failing on reuse

me4502 (approved):

Issue #2989, opened 2026-07-14
Migrate to fabric-permission-api-v1


ThatCuteOne:
### The Problem

Fabric is moving towards fabric-permission-api-v1 (included in the normal fabric api ) i feel like for compatibility reasons this should be implemented

### A Solution

Migrate to fabric-permission-api-v1

### Alternatives

Add support for fabric-permission-api-v1 but also still support fabric-permission-api-v0 from lucko. This would be more maintain cost though since you are supporting 2 apis

### Anything Else?

not sure if i should have put this as a bug report or feat

me4502:
Thanks for making this issue! I'd been a little OOTL on Fabric permissions developments so was not aware this change had occurred.

> Add support for fabric-permission-api-v1 but also still support fabric-permission-api-v0 from lucko. This would be more maintain cost though since you are supporting 2 apis

Just in terms of this one, are you aware of the status of the various permissions mods, and how they've adopted -v1? If they're still fairly split, it might be that we need to keep -v0 around for a little while either way

me4502:
So from having a quick look over the new Fabric permissions API, it honestly seems it has a few _odd_ design decisions that will limit things a bit. Firstly it partially inherits some of the design oddities of NeoForge's permissions system, but thankfully not in a limited way that prevents a mod like WorldEdit from interfacing with the system.

The main two oddities I see (both workable realistically) are:
* Need to create an instance of Identifier per permissions check. This is fairly wasteful/churny when it comes to memory usage. These could theoretically be cached, but that is more a band-aid solution that doesn't scale well when dynamic permission nodes are at play.
* The usage of an identifier means it requires a _different permission node format to the rest of the MC ecosystem_. Theoretically we can adapt to this as all of our permission nodes begin with `worldedit.`, eg `worldedit.whatever.whatever`; so we can use `worldedit:whatever.whatever` fairly easily from a technical standpoint (although again, adds a layer of computation waste). The larger concern is from a user experience / documentation perspective. I guess that side will depend on how permission mods actually decide to show that to users.

me4502:
[...]

PR #2990, opened 2026-07-14
Integrate with the official Fabric Permissions API

me4502:
This is a PoC implementation of Fabric's official Permissions API.

Main limitations are:
* We're having to create a new object for each non-static permissions check (anything other than the command map, basically)
* It uses Identifier for permissions, which has a strict set of allowed characters. It's very very possible our dynamic permission checks will not map to this well. Currently we encode this using a system similar to URL encoding, but it's rather complex for the user. This is kind of a hard-limitation of the Fabric permissions API we cannot cleanly work around.
* It uses an Identifier, so that means `modid:a.b.c`, which of course conflicts with other platforms and our docs. LuckPerms in my testing appears to expose this as `modid.a.b.c` which matches other platforms, but there's no guarantee every permissions mod will follow convention here. The code works around this fine (albeit with some computational overhead), so this is almost entirely a UX issue.

Overall I'm just glad it's actually possible to implement, unlike the NeoForge permissions API, even though there are a few oddities in the design. Everything other than the limited character set of Identifier should be fine though IMO.

We fall back to v0, and then again to MC's permissions system. So this is not a breaking addition.

Adds https://github.com/EngineHub/WorldEdit/issues/2989

### DIFF
--- a/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/internal/FabricPermissionsProvider.java
+++ b/worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/internal/FabricPermissionsProvider.java
@@ -1,0 +1,80 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.fabric.internal;
+
+import com.sk89q.worldedit.coremc.CoreMcPermissionsProvider;
+import net.fabricmc.fabric.api.permission.v1.PermissionNode;
+import net.minecraft.resources.Identifier;
+import net.minecraft.server.level.ServerPlayer;
+import org.jspecify.annotations.Nullable;
+
+import java.util.Map;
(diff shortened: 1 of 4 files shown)


worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/internal/FabricPermissionsProvider.java
```diff
@@ -0,0 +1,108 @@
...
+        StringBuilder encoded = new StringBuilder(bytes.length);
+        for (byte rawByte : bytes) {
+            int valueByte = Byte.toUnsignedInt(rawByte);
+            if ((valueByte >= 'a' && valueByte <= 'z')
+                    || (valueByte >= '0' && valueByte <= '9')
+                    || valueByte == '.' || valueByte == '-'
+                    || (allowSlash && valueByte == '/')) {
+                encoded.append((char) valueByte);
+            } else if (valueByte == '_') {
+                encoded.append("__");
```
octylFractal: Remove to simplify rules?
```suggestion
```

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/internal/FabricPermissionsProvider.java
```diff
@@ -0,0 +1,108 @@
...
+            } else {
+                encoded.append('_');
+                encoded.append(Character.forDigit(valueByte >>> 4, 16));
+                encoded.append(Character.forDigit(valueByte & 0x0F, 16));
```
octylFractal: Easier to read (I did do a little microbench and this is the optimal form):
```suggestion
                encoded.append(HexFormat.of().toHexDigits(rawByte));
```

worldedit-fabric/src/main/java/com/sk89q/worldedit/fabric/internal/FabricPermissionsProvider.java
```diff
@@ -0,0 +1,108 @@
...
+            return fallback.hasPermission(player, permission);
+        }
+
+        Boolean result = player.checkPermission(node);
+        return result != null ? result : fallback.hasPermission(player, permission);
+    }
+
+    @Override
+    public void registerPermission(String permission) {
+        PermissionNode<Boolean> node = createNode(permission);
```
octylFractal, line 66: Should we do a quick `if (registeredPermissions.containsKey(permission)) return;` at the top, to skip making the node if we already know it?

octylFractal (approved):

PR #2992, opened 2026-07-18
Add further Folia checks to catch servers obfuscating that they are Folia

me4502:
If we detect Folia-specific code, print a warning that the server is not marking itself as Folia despite containing this code, and assume that it _is_ Folia.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
@@ -525,8 +525,28 @@ public class WorldEditPlugin extends JavaPlugin implements TabCompleter {
             if (PaperLib.isPaper()) {
                 // Then we can check against the `papermc:folia` key, as per the `isBrandCompatible` javadoc.
                 // This API is experimental so might randomly break on us (hence the try/catch)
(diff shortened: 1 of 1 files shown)


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
```diff
@@ -525,18 +525,45 @@ BukkitImplAdapter getBukkitImplAdapter() {
...
+
+        try {
+            Bukkit.class.getMethod("getRegionScheduler");
+            hiddenFoliaCheckPassed = true;
+        } catch (NoSuchMethodException ignored) {
+            // This is a method existence check, it's fine if not present.
+        }
+
+        if (hiddenFoliaCheckPassed) {
+            LOGGER.warn("Server platform not marked as Folia-based, but appears to have Folia-specific code. Assuming this server is running Folia.");
```
octylFractal: Can we maybe discourage this?
```suggestion
            LOGGER.warn("Server platform not marked as Folia-based, but appears to have Folia-specific code. Assuming this server is running Folia. This check is fragile, please tell the author of your server software to report that it is compatible with the `papermc:folia` brand.");
```

octylFractal (approved):

PR #2997, opened 2026-08-02
Refactor block material

TomyLobo:
- Add BlockState.getMaterial()
- Add ShapeType parameter to BlockMaterial.isFullCube

### DIFF
--- a/verification/src/changes/accepted-core-public-api-changes.json
+++ b/verification/src/changes/accepted-core-public-api-changes.json
@@ -141,5 +141,14 @@
                 "ANNOTATION_REMOVED"
             ]
         }
+    ],
+    "Why on earth is this even an issue?": [
+        {
+            "type": "com.sk89q.worldedit.world.registry.BlockMaterial",
+            "member": "Method com.sk89q.worldedit.world.registry.BlockMaterial.isFullCube()",
+            "changes": [
+                "METHOD_ABSTRACT_NOW_DEFAULT"
+            ]
+        }
     ]
 }
--- a/verification/src/changes/accepted-core-public-api-changes.json
+++ b/verification/src/changes/accepted-core-public-api-changes.json
@@ -102,6 +102,14 @@
         },
         {
             "type": "com.sk89q.worldedit.world.registry.BlockRegistry",
+            "member": "Method com.sk89q.worldedit.world.registry.BlockRegistry.getMaterial(com.sk89q.worldedit.world.block.BlockType)",
+            "changes": [
+                "METHOD_ABSTRACT_NOW_DEFAULT",
+                "ANNOTATION_DEPRECATED_ADDED"
+            ]
+        },
+        {
+            "type": "com.sk89q.worldedit.world.registry.BlockRegistry",
             "member": "Method com.sk89q.worldedit.world.registry.BlockRegistry.getName(com.sk89q.worldedit.world.block.BlockType)",
             "changes": [
                 "METHOD_REMOVED",
--- a/verification/src/changes/accepted-core-public-api-changes.json
+++ b/verification/src/changes/accepted-core-public-api-changes.json
@@ -101,6 +101,14 @@
             ]
         },
         {
+            "type": "com.sk89q.worldedit.world.registry.BlockMaterial",
+            "member": "Method com.sk89q.worldedit.world.registry.BlockMaterial.isFullCube()",
+            "changes": [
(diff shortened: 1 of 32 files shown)


verification/src/changes/accepted-core-public-api-changes.json
```diff
@@ -141,5 +141,14 @@
                 "ANNOTATION_REMOVED"
             ]
         }
+    ],
+    "Why on earth is this even an issue?": [
```
octylFractal: If someone was implementing `BlockMaterial`, this can sometimes cause weird issues that I don't recall exactly. But we should have the actual reasoning here, e.g.
```suggestion
    "No one should be implementing BlockMaterial": [
```

worldedit-core/src/main/java/com/sk89q/worldedit/internal/block/AbstractBlockMaterial.java
```diff
@@ -0,0 +1,49 @@
...
+
+package com.sk89q.worldedit.internal.block;
+
+import com.sk89q.worldedit.blocks.ShapeType;
+import com.sk89q.worldedit.util.concurrency.LazyReference;
+import com.sk89q.worldedit.world.registry.BlockMaterial;
+
+import java.util.EnumSet;
+
+public abstract class AbstractBlockMaterial<VS> implements BlockMaterial {
```
octylFractal, line 28: Instead of having a `VS` parameter, can we just push `getShape` into `isShapeFullBlock`? There doesn't really seem to be an advantage to having it here.

Also, we should probably just call `isShapeFullBlock` -> `isFullCubeUncached`?

worldedit-core/src/main/java/com/sk89q/worldedit/blocks/ShapeType.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldedit.blocks;
+
+public enum ShapeType {
+    SHAPE,
+    VISUAL_SHAPE,
```
octylFractal, line 31: This should explain what the difference is, and what a `ShapeType` is generally -- i.e., all members and the class should be documented.

octylFractal (changes requested):

me4502:
Also alongside octy's review, this currently does not compile, if you'd be able to please fix that too

worldedit-core/src/main/java/com/sk89q/worldedit/internal/block/AbstractBlockMaterial.java
TomyLobo: What about `isFaceFull` though?

octylFractal: It can have the same thing done to it as well? You literally just do `isFaceFull(getShape(shapeType), face)` and do not store the shape in any way.

verification/src/changes/accepted-core-public-api-changes.json
TomyLobo: I moved that under "Removal of deprecated API for WorldEdit 8" and deprecated the method.
I think that's what we settled on as a long-term goal anyway when we discussed this a few months back.

worldedit-core/src/main/java/com/sk89q/worldedit/blocks/ShapeType.java
TomyLobo: added docs

worldedit-core/src/main/java/com/sk89q/worldedit/internal/block/AbstractBlockMaterial.java
TomyLobo: I made getShape a private helper instead of an overridden method and made the other methods take `ShapeType` directly
thought it'd be more code duplication, for some reason, but that didn't pan out when I actually tried it :)

TomyLobo: I also renamed the method as suggested (and the other one to isFaceFullUncached in the other PR, too)

me4502: It seems like this still has the <VS> type parameter on the class?

Issue #3001, opened 2026-08-06
Make //drain command also remove bubble columns


madebyisaacr:
### The Problem

The //drain command removes all water within a specific radius. However, it leaves `bubble_column` blocks, so if you try to drain an area where there are some magma blocks or soul sand on the floor, it won't be fully drained.

### A Solution

Make the //drain command also remove `bubble_column` blocks.

This would be useful for draining water-filled caves, as magma blocks can generate on the floor of water caves causing pillars of bubble column blocks to be left behind after using the drain command.

me4502:
There are a few blocks that always provide water which we should probably support, including kelp. IMO this should be another flag on the command similar to the waterlogged removal one, as this is a much more destructive action than just the basic drain command.

Really depends on whether MC provides a clean way to tell if a block "provides water" despite not having a waterlogged property.

PR #3003, opened 2026-08-08
Properly support fluids

me4502:
This PR adds proper FluidState/FluidType/FluidCategory support, migrating our historic partial registries to be fully featured as well as adding FluidState as a derived BlockState property.

This allows us in the future to support features such as https://github.com/EngineHub/WorldEdit/issues/3001, as the way to differentiate those blocks is to use fluid states.

This PR purely exposes the fluid system within our API, it does not add any features _using_ those fluids at this point.


worldedit-core/src/main/java/com/sk89q/worldedit/world/fluid/FluidState.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * <p>A fluid state is derived from a block state by the active platform. This
+ * allows blocks such as kelp and waterlogged blocks to supply fluid without
+ * being fluid blocks themselves.</p>
+ *
+ * @param type the type of fluid supplied
+ */
+public record FluidState(FluidType type) {
+
+    public static final FluidState EMPTY = new FluidState(FluidTypes.EMPTY);
```
octylFractal: Can this be a `private` field and have a method `empty()` instead? It's harder to change a field in case we need to adjust initialization logic or something.

octylFractal (approved):
Approving as-is, we'll look at cleaning it up later alongside `BlockState`.

me4502:
Follow-up issue for cleanup: https://github.com/EngineHub/WorldEdit/issues/3006

PR #3009, opened 2026-08-15
feat: fully implement Player#sendFakeBlock

Timongcraft:
## Overview
This PR aims to improves the `Player#sendFakeBlock` implementation to support non structure-blocks (used by the CUI).

## Description
Replaces the structure block specific implementation with a generic implementation to allow API users to send fake blocks containing NBT such as player heads, banners etc.

## TODO
- [x] implement other platforms
- [x] fix breakage on spigot


Timongcraft:
I'm not sure if the old `BukkitImplAdapter#sendFakeNBT(Player, BlockVector3, LinCompoundTag)` should be removed directly or marked for removal since it shouldn't be used externally but is available in the bukkit module and kinda declared as public api [here](https://worldedit.enginehub.org/en/latest/api/internal-apis/), no?

octylFractal:
It is not API, you can freely delete it.

PR #3011, opened 2026-08-28
Fix pistons breaking when head is replaced 

### LINKED ISSUE #2956: Paste breaks and never replaces extended pistons
### WorldEdit Version

7.4.3-beta-01+80c6d6122

### Platform Version

Fabric 0.19.2

### Confirmations

### Bug Description

Basically when I use the paste command it tries to replace extended pistons with another extended or non-extended piston but instead I see break particles and the piston just never replaces.

### Expected Behavior

It should be able to replace it correctly instead of just dissapearing

### Reproduction Steps

1. Copy a piston
2. Paste a piston over an extended piston

### Anything Else?

_No response_

VanJackal:
Fixes #2956 by adding a check for if the block was changed before calling `affectNeighborsAfterRemoval`

### DIFF
--- a/worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/internal/CoreMcWorldNativeAccess.java
+++ b/worldedit-core-mc/src/main/java/com/sk89q/worldedit/coremc/internal/CoreMcWorldNativeAccess.java
@@ -166,7 +166,10 @@ public final class CoreMcWorldNativeAccess implements WorldNativeAccess<LevelChu
     @Override
     public void updateNeighbors(BlockPos pos, BlockState oldState, BlockState newState, int recursionLimit) {
         ServerLevel world = getWorld();
-        oldState.affectNeighborsAfterRemoval(world, pos, false);
+        boolean blockChanged = !oldState.is(newState.getBlock());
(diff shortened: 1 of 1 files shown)


me4502:
Thank you for looking into this! 😀 

I can't investigate or review too deeply right now as I'm just replying on my lunch break, but from memory the entire "dirty" / update process here should be gated by a "changed check" coming from MC itself further up, unless something has changed within MC.

Either way if that check itself no longer functions, it should be fixed at the top level rather than in here. There are niche situations where we want to force these to occur always (eg, //update command), and it's not the only change that should be skipped during a no-op change during typical edits.

VanJackal:
Thanks for the reply.

I haven't looked into the 'dirty' check yet but I did look into //update and currently it has the same issue, so any extended pistons that are affected by //update will be broken. 

I also looked at the other uses of `affectNeighborsAfterRemoval` and found a related issue with tripwires being put into a weird state when //update is used on them. It looks like `affectNeighborsAfterRemoval` is mostly used to push comparator updates from containers however comparators seem to still update even without the `affectNeighborsAfterRemoval` call, so I'm not totally sure.

PR #3019, opened 2026-09-22
Work around a MC DFU bug in MC 26.3

me4502:
This PR works around a few DFU bugs in MC 26.3, that causes some schematics to fail to load. 

One of the bugs can theoretically lead to data loss, so can only properly be fixed in MC itself. There's no way for us to work around that one with 100% safety, so I've done a best effort approach to it on our end (which should result in as minimal data loss as possible without MC fixing it)


worldedit-bukkit/adapters/adapter-26.3/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v26_3/PaperweightDataConverters.java
```diff
@@ -141,13 +141,19 @@ private String fixBlockState(String blockState, int srcVer) {
         CompoundTag stateNBT = stateToNBT(blockState);
         Dynamic<Tag> dynamic = new Dynamic<>(OPS_NBT, stateNBT);
         CompoundTag fixed = (CompoundTag) INSTANCE.fixer.update(References.BLOCK_STATE, dynamic, srcVer, DATA_VERSION).getValue();
-        return nbtToState(fixed);
+        var newState = nbtToState(fixed);
```
octylFractal: nit: let's use the type if not obvious
```suggestion
        String newState = nbtToState(fixed);
```

worldedit-bukkit/adapters/adapter-26.3/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v26_3/PaperweightDataConverters.java
```diff
@@ -141,13 +141,19 @@ private String fixBlockState(String blockState, int srcVer) {
...
+        // blocks like air, so it's theoretically safe to do this.
+        return newState.isBlank() ? blockState : newState;
     }
 
     private String nbtToState(net.minecraft.nbt.CompoundTag tagCompound) {
         StringBuilder sb = new StringBuilder();
-        sb.append(tagCompound.getString("Name").get());
-        tagCompound.getCompound("Properties").ifPresent(props -> {
+        // As of MC 26.3, this seems to be able to return _either_ "Name" or "id", so we check both. "id" is theoretically
+        // the correct one, but "Name" is a legacy field that can sometimes be returned.
```
octylFractal: thanks i hate it mojang

octylFractal (approved):

